PR #24, opened 2019-06-07
Boss erase edges dyn

karasikov:


karasikov:
Let's close this for now, as it has conflicts with the main branch.

We'll come back to this when we need dynamic edge removal.

So, the default strategy will still assume that the edges are removed in batches.

PR #188, opened 2020-07-27
Differential assembly on canonical and primary graphs

hmusta:
This includes
1) a revision of the differential assembly procedure to allow for differential sequences to be assembled from canonical and primary graphs
2) some optimization to only perform row queries when absolutely necessary
3) an option to perform multiple differential assemblies on the same annotated graph with a single load, reading the foreground and background group specifications from a text file
4) parallelization: perform multiple differential assemblies in parallel, using at most 4 cores for each assembly to avoid unnecessary overhead
5) a binary matrix query called `slice_columns` which allows for efficient column query in BRWT
6) a method for creating an `sdsl::int_vector<width>` with aligned memory
7) atomic increment and set operators for `sdsl::int_vector<>`


metagraph/src/annotation/binary_matrix/base/binary_matrix.hpp
```diff
@@ -21,6 +21,7 @@ class BinaryMatrix {
     typedef Vector<Column> SetBitPositions;
     typedef std::function<void(const SetBitPositions &)> RowCallback;
     typedef std::function<void(Row, Column)> ValueCallback;
+    typedef std::function<void(Column, std::vector<Row>&&)> ColumnCallback;
```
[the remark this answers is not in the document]
hmusta: I replaced this with `bitmap` and ended using `bit_vector_smart` in a lot of places.

hmusta:
I figured out that if I reallocate an `int_vector<>`'s underlying storage with `std::aligned_alloc`, then I can use atomic operations to implement `operator=` and `operator+=` on `int_vector<>`.

metagraph/src/common/vectors/vector_algorithm.hpp
```diff
@@ -109,6 +109,41 @@ inline void unset_bit(uint64_t *v,
...
+inline uint64_t atomic_increment(sdsl::int_vector<> &vector, uint64_t i,
+                                 uint64_t count = 1,
+                                 int memorder = __ATOMIC_SEQ_CST) {
+    size_t width = vector.width();
+    uint64_t bit_pos = i * width;
+    __uint128_t *limb = &reinterpret_cast<__uint128_t*>(vector.data())[bit_pos >> 7];
+    __uint128_t mask = ((__uint128_t(1) << width) - 1) << (bit_pos & 0x7F);
+    __uint128_t new_val;
+    __uint128_t inc = __uint128_t(count) << (bit_pos & 0x7F);
+    __uint128_t exp_val = *limb;
```
hmusta, line 121: TODO: check if mode TI is available

metagraph/src/common/vectors/vector_algorithm.hpp
```diff
@@ -109,6 +109,41 @@ inline void unset_bit(uint64_t *v,
...
+
+inline uint64_t atomic_set(sdsl::int_vector<> &vector, uint64_t i, uint64_t val = 1,
+                           int memorder = __ATOMIC_SEQ_CST) {
+    size_t width = vector.width();
+    uint64_t bit_pos = i * width;
+    __uint128_t *limb = &reinterpret_cast<__uint128_t*>(vector.data())[bit_pos >> 7];
+    __uint128_t mask = ((__uint128_t(1) << width) - 1) << (bit_pos & 0x7F);
+    __uint128_t new_val;
+    __uint128_t val_shift = __uint128_t(val) << (bit_pos & 0x7F);
+    __uint128_t exp_val = *limb;
```
hmusta, line 138: TODO: check if mode TI is available

metagraph/src/common/vectors/vector_algorithm.hpp
hmusta: if not, then this function has to be implemented with a mutex

metagraph/src/common/vectors/vector_algorithm.hpp
```diff
@@ -489,6 +524,48 @@ prev_bit(const t_int_vec &v,
...
+    assert(!width || width == var_width);
+
+    sdsl::int_vector<width> v;
+    auto &v_cast = reinterpret_cast<int_vector_access&>(v);
+    v_cast.m_size = size * var_width;
+    v_cast.m_width = var_width;
+    free(v_cast.m_data);
+
+    size_t capacity_bytes = (((((v_cast.m_size + 63) >> 6) << 3) + alignment - 1) / alignment) * alignment;
+    v_cast.m_data = (uint64_t*)std::aligned_alloc(alignment, capacity_bytes);
```
hmusta, line 551: TODO: `std::aligned_alloc` is not available in Xcode...

metagraph/src/common/vectors/vector_algorithm.hpp
```diff
@@ -109,6 +110,59 @@ inline void unset_bit(uint64_t *v,
...
+    __uint128_t exp_val = *limb;
+    do {
+        new_val = ((exp_val + inc) & mask) | (exp_val & (~mask));
+    } while (!__atomic_compare_exchange(limb, &exp_val, &new_val, true,
+                                        memorder /* order for exchange */,
+                                        __ATOMIC_RELAXED /* order for compare */));
+    return (exp_val & mask) >> (bit_pos & 0x7F);
+#else
+    std::ignore = memorder;
+    static std::mutex mu;
```
karasikov, line 132: ideally, this should be a local mutex for vector, in case you write to two or more vectors at the same time.

karasikov (changes requested):
checked .hpp files

metagraph/src/common/vectors/vector_algorithm.hpp
hmusta: Rather than passing a mutex, I'm thinking of using a builtin function for locking those memory locations.

metagraph/src/graph/representation/masked_graph.hpp
```diff
@@ -88,6 +84,7 @@ class MaskedDeBruijnGraph : public DeBruijnGraph {
     virtual size_t get_k() const override { return graph_->get_k(); }
 
     virtual bool is_canonical_mode() const override { return is_canonical_; }
+    virtual void set_canonical_mode(bool is_canonical_mode) { is_canonical_ = is_canonical_mode; }
```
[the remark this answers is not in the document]
hmusta: sorry, this was left behind from an earlier implementation

metagraph/src/graph/annotated_graph_algorithm.hpp
```diff
@@ -4,66 +4,57 @@
...
+    bool add_complement = false;
+};
+
+std::pair<sdsl::int_vector<>, std::unique_ptr<bitmap>>
+fill_count_vector(const AnnotatedDBG &anno_graph,
+                  const std::vector<AnnotatedDBG::Annotator::Label> &labels_in,
+                  const std::vector<AnnotatedDBG::Annotator::Label> &labels_out,
+                  size_t num_threads,
+                  bool update_in_place,
+                  const sdsl::int_vector<> *init_counts = nullptr);
```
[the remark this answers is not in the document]
hmusta: I removed `update_in_place` from the interface since it can be computed internally. Basically, it's set to true if the underlying graph is DBGSuccinct because the implementation of `call_sequences` for masked DBGSuccinct makes a copy of a bitmask, so `update_in_place` is used to prevent making another bitmask.

metagraph/CMakeLists.txt
```diff
@@ -129,7 +129,7 @@ try_compile(MODE_TI
   "${PROJECT_SOURCE_DIR}/external-libraries/sdsl-lite/CMakeModules/check_mode_ti.cpp"
 )
 if(MODE_TI)
-  string(APPEND CMAKE_CXX_FLAGS " -DMODE_TI")
+  string(APPEND CMAKE_CXX_FLAGS " -DMODE_TI -mcx16")
```
hmusta, line 132: This adds support for the `CMPXCHG16B` (128-bit compare and exchange) instruction

metagraph/src/graph/representation/succinct/dbg_succinct.hpp
```diff
@@ -109,7 +109,9 @@ class DBGSuccinct : public DeBruijnGraph {
     virtual uint64_t num_nodes() const override final;
 
     virtual void mask_dummy_kmers(size_t num_threads, bool with_pruning) final;
+    virtual const bit_vector* get_mask() const final;
```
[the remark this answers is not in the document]
hmusta: We don't always have a mask available, so making this return a const reference can be dangerous.

So I can either add a comment to this function saying that it returns nullptr if a mask hasn't been initialized, or I can add an extra function `bool is_masked() const` which can be checked before calling `get_mask`

metagraph/src/annotation/binary_matrix/multi_brwt/brwt.hpp
```diff
@@ -35,6 +35,9 @@ class BRWT : public BinaryMatrix {
     // get all selected rows appended with -1 and concatenated
     std::vector<Column> slice_rows(const std::vector<Row> &rows) const override;
 
+    void slice_columns(const std::vector<Column> &columns,
+                       const ColumnCallback &callback) const override;
+
```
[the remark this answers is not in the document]
hmusta: It was faster with the old implementation that returned `std::vector<Row>`, but I have to benchmark it again with the current implementation.

metagraph/tests/annotation/test_annotated_dbg_helpers.hpp
```diff
@@ -29,15 +31,8 @@ using namespace mtg::graph;
...
-
-MaskedDeBruijnGraph build_masked_graph(const AnnotatedDBG &anno_graph,
-                                       const std::vector<std::string> &ingroup,
-                                       const std::vector<std::string> &outgroup,
-                                       double mask_in_label_fraction = 1.0,
-                                       double mask_out_label_fraction = 0.0,
-                                       double other_label_fraction = 1.0,
-                                       double lazy_evaluation_density_cutoff = 0.05);
+                                               const std::vector<std::string> &labels,
+                                               BuildMode mode = BASE);
```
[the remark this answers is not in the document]
hmusta: ok, `DBGMode` works.

`BASE` refers to building a DBG directly from the sequences, `CANONICAL` builds a canonical DBG, while `WRAPPER` builds a primary DBG and then wraps it in a `CanonicalDBG`.

I'm thinking of also adding another mode `PRIMARY` which just returns a primary graph without wrapping it.

metagraph/src/graph/representation/succinct/dbg_succinct.hpp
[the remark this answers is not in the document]
karasikov: I see
Yes, a comment will help

metagraph/tests/annotation/test_annotated_dbg_helpers.hpp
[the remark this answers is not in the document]
karasikov: `WRAPPER` is not ideal, maybe `CANONICAL_WRAPPER`?

metagraph/src/common/vectors/vector_algorithm.hpp
hmusta: I ended up replacing this with a local mutex since that was the easiest approach.

metagraph/tests/annotation/test_annotated_dbg_helpers.hpp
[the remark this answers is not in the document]
hmusta: I added `PRIMARY` as a build mode, renamed `BASE` to `NORMAL`, and renamed `BuildMode` to `DBGMode`

metagraph/src/common/vectors/vector_algorithm.hpp
```diff
@@ -109,6 +110,90 @@ inline void unset_bit(uint64_t *v,
...
+
+#endif
+
+    std::lock_guard<std::mutex> lock(backup_mutex);
+    std::atomic_thread_fence(std::memory_order_acquire);
+    uint64_t old_val = vector[i];
+    vector[i] += count;
+    std::atomic_thread_fence(std::memory_order_release);
+    return old_val;
+}
```
hmusta, line 164: TODO: This creates a race condition....another thread could have done an atomic CAS operation in the middle of `vector[i] += count`

hmusta: I've disabled it for now so we can look into it later. Maybe we can use transactional memory intrinsics for this at some point.

hmusta: Unaligned atomic operations for <= 64-bit words are allowed, so this was rewritten accordingly.

metagraph/benchmarks/benchmark_matrix.cpp
```diff
@@ -115,4 +115,100 @@ BENCHMARK_TEMPLATE(BM_BRWTQueryRows, 3000000, 100, 30, 2, false, 0)
...
+        std::atomic<uint64_t> j;
+        #pragma omp parallel num_threads(3)
+        #pragma omp single
+        {
+            matrix->slice_columns(indexes, [&](auto i, auto&& bitmap) {
+                j.fetch_add(i, memory_order_seq_cst);
+                bitmap.call_ones([&](auto t) { j.fetch_add(t, memory_order_relaxed); });
+            });
+        }
+        ++j;
```
karasikov, line 206: why multiple therads?

metagraph/benchmarks/benchmark_matrix.cpp
```diff
@@ -115,4 +115,100 @@ BENCHMARK_TEMPLATE(BM_BRWTQueryRows, 3000000, 100, 30, 2, false, 0)
...
+    for (auto _ : state) {
+        std::atomic<uint64_t> j;
+        #pragma omp parallel num_threads(3)
+        #pragma omp single
+        {
+            matrix->slice_columns(indexes, [&](auto i, auto&& bitmap) {
+                j.fetch_add(i, memory_order_seq_cst);
+                bitmap.call_ones([&](auto t) { j.fetch_add(t, memory_order_relaxed); });
+            });
+        }
```
karasikov, line 205: This is called `SliceColumns`, so why do you test it with `fetch_add`?

karasikov (changes requested):

metagraph/benchmarks/benchmark_matrix.cpp
hmusta: I should write a comment about this in the header. `slice_columns` uses openmp tasks to run in multiple threads

metagraph/benchmarks/benchmark_matrix.cpp
hmusta: For this one, I don't really care about the value of j, so I don't need_ to avoid race conditions or anything by using `fetch_add`. I only really set those there so that the callback would not be optimized out.

karasikov: just remove it, I think it will work fine

metagraph/benchmarks/benchmark_matrix.cpp
karasikov: so what is the performance of this vs `get_column`?

metagraph/src/common/vectors/vector_algorithm.hpp
```diff
@@ -109,6 +110,184 @@ inline void unset_bit(uint64_t *v,
...
+    } else if ((bit_pos & 0x3F) + width <= 64) {
+        // element fits in an aligned word
+        uint64_t *word = &vector.data()[bit_pos >> 6];
+        uint8_t shift = bit_pos & 0x3F;
+        return (__atomic_fetch_add(word, count << shift, memorder) >> shift) & mask;
+    } else if ((bit_pos & 0x7F) + width <= 128) {
+        // element fits in an aligned double word
+        __uint128_t *word = &reinterpret_cast<__uint128_t*>(vector.data())[bit_pos >> 7];
+        uint8_t shift = bit_pos & 0x7F;
+        return (__sync_fetch_and_add(word, __uint128_t(count) << shift) >> shift) & mask;
```
hmusta, line 176: Add comment here that gcc only generates a `cmpxchg16b` instruction when older `__sync` instructions are used...

metagraph/src/common/vectors/vector_algorithm.hpp
```diff
@@ -109,6 +110,184 @@ inline void unset_bit(uint64_t *v,
...
+        __uint128_t *word = &reinterpret_cast<__uint128_t*>(vector.data())[bit_pos >> 7];
+        uint8_t shift = bit_pos & 0x7F;
+        __uint128_t desired;
+        __uint128_t exp;
+        __uint128_t inv_mask = ~(__uint128_t(mask) << shift);
+        __uint128_t big_val = __uint128_t(val) << shift;
+        do {
+            exp = *word;
+            desired = big_val | (inv_mask & exp);
+        } while (!__sync_bool_compare_and_swap(word, exp, desired));
```
hmusta, line 235: Add comment here that gcc only generates a `cmpxchg16b` instruction when older `__sync` instructions are used...

metagraph/benchmarks/benchmark_matrix.cpp
hmusta: It runs in about the same amount of time, but has 15% less RAM overhead

hmusta: Though that was testing with the BIGSI 720 subset. I think the size is too small in the benchmark, so there's more overhead and this isn't seen.

metagraph/benchmarks/benchmark_matrix.cpp
```diff
@@ -115,4 +115,98 @@ BENCHMARK_TEMPLATE(BM_BRWTQueryRows, 3000000, 100, 30, 2, false, 0)
...
     ->DenseRange(0, 10, 1);
 
+
+template <size_t rows_arg = 300000,
+          size_t cols_arg = 100,
+          size_t unique_arg = 10,
+          size_t arity_arg = 2,
+          bool greedy_arg = true,
+          size_t relax_arg = 2>
+static void BM_BRWTQueryColumns(benchmark::State& state) {
```
karasikov, line 125: Don't build it here, better pass a constructed one with an environment variable (have a look at boss benchmarks).

It will be much easier to benchmark things on different types of annotation. Also, you can write one benchmark for all annotation types, not only for BRWT.

metagraph/tests/test_vector_algorithm.cpp
```diff
@@ -123,7 +123,162 @@ TEST(IntVector, call_nonzeros_sparse) {
...
+                    EXPECT_EQ(vector[i], count);
+                    EXPECT_EQ(it++, i);
+                });
+                for (; it < end; ++it) {
+                    EXPECT_EQ(0u, vector[it]);
+                }
+            }
+        }
+    }
+}
```
karasikov, line 175: the test is called `call_nonzeros_sparse_atomic`, but call_nonzeros is called with a single thread and there is no atomics.

What do you actually want to test here?

metagraph/tests/test_vector_algorithm.cpp
```diff
@@ -123,7 +123,162 @@ TEST(IntVector, call_nonzeros_sparse) {
...
+                    EXPECT_EQ(0u, vector[it]) << vector.size()
+                                              << "\n" << begin
+                                              << " " << end
+                                              << "\n" << w
+                                              << "\n" << vector;
+                }
+            }
+        }
+    }
+}
```
karasikov, line 279: seems like a huge test. Hard to see what you want to check here.
Try to break it down into two or more tests and test simple things there.

Otherwise, describe in a comment why you need such a complex test.

metagraph/tests/test_vector_algorithm.cpp
```diff
@@ -123,7 +123,162 @@ TEST(IntVector, call_nonzeros_sparse) {
...
+                    EXPECT_EQ(vector[i], count);
+                    EXPECT_EQ(it++, i);
+                });
+                for (; it < end; ++it) {
+                    EXPECT_EQ(0u, vector[it]);
+                }
+            }
+        }
+    }
+}
```
karasikov, line 224: any reason to throw all vector_relaxed, vector_seq_cst, vector into one test?

metagraph/src/graph/representation/succinct/dbg_succinct.hpp
```diff
@@ -110,7 +110,12 @@ class DBGSuccinct : public DeBruijnGraph {
     virtual uint64_t num_nodes() const override final;
 
     virtual void mask_dummy_kmers(size_t num_threads, bool with_pruning) final;
+
+    // If a mask is defined, return a const pointer to it. Otherwise, return nullptr
```
karasikov, line 114: remove all words that are not necessary to keep it as short as possible
```suggestion
    // Return a pointer to the mask, or NULL if not initialized
```

metagraph/src/graph/representation/succinct/dbg_succinct.cpp
```diff
@@ -890,10 +890,18 @@ void DBGSuccinct::mask_dummy_kmers(size_t num_threads, bool with_pruning) {
     assert(!(*valid_edges_)[0]);
 }
 
+const bit_vector* DBGSuccinct::get_mask() const {
+    return valid_edges_.get();
+}
```
karasikov, line 895: to .hpp?

metagraph/src/graph/representation/succinct/dbg_succinct.cpp
```diff
@@ -890,10 +890,18 @@ void DBGSuccinct::mask_dummy_kmers(size_t num_threads, bool with_pruning) {
...
+
 void DBGSuccinct::reset_mask() {
     valid_edges_.reset();
 }
 
+bit_vector* DBGSuccinct::release_mask() {
+    return valid_edges_.release();
+}
```
karasikov, line 903: move to .hpp?

metagraph/src/graph/representation/masked_graph.hpp
```diff
@@ -113,14 +109,24 @@ class MaskedDeBruijnGraph : public DeBruijnGraph {
...
 
-    virtual void set_mask(bitmap *mask) { kmers_in_graph_.reset(mask); }
+    virtual void set_mask(std::unique_ptr<bitmap>&& mask) {
+        assert(mask.get());
+        assert(mask->size() == graph_->max_index() + 1);
+        kmers_in_graph_ = std::move(mask);
+    }
+
+    typedef std::function<void(const std::function<void(node_index, bool)>&)> GenerateNodes;
+    virtual void update_mask(const GenerateNodes &generate_nodes,
```
karasikov, line 119: not clear what that `bool` represents
```suggestion
    typedef std::function<void(const std::function<void(node_index, bool /* ??? */)>&)> GenerateNodes;
    virtual void update_mask(const GenerateNodes &generate_nodes,
```

metagraph/src/graph/annotated_graph_algorithm.hpp
```diff
@@ -4,66 +4,62 @@
...
+//                              present for a node to be kept
+// label_mask_in_unitig_fraction: minimum fraction of nodes within a unitig which
+//                                must satisfy the in-label criteria
+// label_mask_out_kmer_fraction: maximum fraction of out-labels which can be
+//                               present in a kept node
+// label_mask_out_unitig_fraction: maximum fraction of nodes within a unitig which
+//                                 can satisfy the out-label criteria
+// label_mask_other_unitig_fraction: maximum fraction of nodes within a unitig which
+//                                   can contain labels not in the in- and out-groups
+// add_complement: also consider the reverse complements of nodes
```
karasikov, line 28: /**
...
*/

metagraph/src/graph/annotated_graph_algorithm.cpp
```diff
@@ -1,238 +1,503 @@
...
 
-    return std::make_unique<bitmap_lazy>([=](DeBruijnGraph::node_index i) {
-        return i != DeBruijnGraph::npos
-            && is_node_in_mask(i, [&]() { return count_frequent_in_labels(i); },
-                                  [&]() { return count_frequent_out_labels(i); });
-    }, anno_graph.get_graph().max_index() + 1);
+    logger->trace("Kept {} out of {} nodes", kept_nodes, total_nodes);
 }
 
+
```
karasikov, line 501:
```suggestion
```

metagraph/src/cli/load/load_annotated_graph.hpp
```diff
@@ -17,8 +16,6 @@ initialize_annotated_dbg(std::shared_ptr<graph::DeBruijnGraph> graph,
 
 std::unique_ptr<graph::AnnotatedDBG> initialize_annotated_dbg(const Config &config);
 
-std::unique_ptr<graph::MaskedDeBruijnGraph>
-mask_graph(const graph::AnnotatedDBG &anno_graph, Config *config);
```
karasikov, line 19:
```suggestion
```

metagraph/src/cli/load/load_annotated_graph.cpp
```diff
@@ -19,7 +18,7 @@ using mtg::common::logger;
 std::unique_ptr<AnnotatedDBG> initialize_annotated_dbg(std::shared_ptr<DeBruijnGraph> graph,
                                                        const Config &config) {
     // TODO: check and wrap into canonical only if the graph is primary
-    if (config.canonical && !graph->is_canonical_mode())
+    if (config.canonical && !graph->is_canonical_mode() && config.identity != Config::ASSEMBLE)
```
karasikov, line 21: Too much digging for this loader.

Better pass a flag as a parameter:
`initialize_annotated_dbg(... graph, ... config, bool wrap_primary = true);`

metagraph/src/cli/config/config.hpp
```diff
@@ -85,6 +84,7 @@ class Config {
     unsigned int bloom_max_num_hash_functions = 10;
     unsigned int num_columns_cached = 10;
     unsigned int max_hull_forks = 4;
+    unsigned int parallel_assemblies = -1;
```
karasikov, line 87: add a comment what -1 is meant to represent

metagraph/src/cli/clean.cpp
```diff
@@ -72,7 +72,7 @@ int clean_graph(Config *config) {
             // skip zero k-mer counts for dummy k-mers in DBGSuccinct
             const auto _graph = dynamic_cast<graph::DBGSuccinct*>(graph.get())
                     ? std::make_shared<graph::MaskedDeBruijnGraph>(graph,
-                        [&](auto i) { return (*node_weights)[i] > 0; }, true)
+                        [&](auto i) { return (*node_weights)[i] > 0; })
```
karasikov, line 75: did you check if the cleaning results don't change?

karasikov: Maybe add integration tests for that?

metagraph/src/annotation/binary_matrix/base/binary_matrix.hpp
```diff
@@ -32,9 +35,13 @@ class BinaryMatrix {
     virtual SetBitPositions get_row(Row row) const = 0;
     virtual std::vector<SetBitPositions> get_rows(const std::vector<Row> &rows) const;
     virtual std::vector<Row> get_column(Column column) const = 0;
+
     // get all selected rows appended with -1 and concatenated
     virtual std::vector<Column> slice_rows(const std::vector<Row> &rows) const;
 
+    virtual void slice_columns(const std::vector<Column> &columns,
+                               const ColumnCallback &callback) const;
```
karasikov, line 43: why not just `std::function<void(bitmap&&)> ColumnCallback`?

metagraph/src/annotation/binary_matrix/base/binary_matrix.cpp
```diff
@@ -32,6 +34,25 @@ BinaryMatrix::slice_rows(const std::vector<Row> &row_ids) const {
...
+        auto column = get_column(j);
+        size_t nsetbits = column.size();
+        callback(j, bit_vector_smart([s=std::move(column)](const auto &index_callback) {
+            for (uint64_t i : s) {
+                index_callback(i);
+            }
+        }, nrows, nsetbits));
+    }
+
+    #pragma omp taskwait
```
karasikov, line 53: Using more threads will use more RAM as well, which is not cool. Maybe do it with a single thread?

metagraph/src/annotation/binary_matrix/base/binary_matrix.cpp
```diff
@@ -32,6 +34,25 @@ BinaryMatrix::slice_rows(const std::vector<Row> &row_ids) const {
...
+    #pragma omp taskloop
+    for (size_t k = 0; k < column_ids.size(); ++k) {
+        Column j = column_ids[k];
+        auto column = get_column(j);
+        size_t nsetbits = column.size();
+        callback(j, bit_vector_smart([s=std::move(column)](const auto &index_callback) {
+            for (uint64_t i : s) {
+                index_callback(i);
+            }
+        }, nrows, nsetbits));
```
karasikov, line 50: What happened to that other bitmap based on a sorted vector? Was it in this PR?

karasikov (changes requested):

metagraph/src/annotation/binary_matrix/base/binary_matrix.cpp
hmusta: I used that for the BRWT implementation, but forgot to change the general one to also use it.

metagraph/tests/test_vector_algorithm.cpp
hmusta: My original plan was to do the testing in `call_nonzeros`, but ended up testing the vector in the previous section. I can get rid of the second half of the test.

metagraph/tests/test_vector_algorithm.cpp
hmusta: I changed it so that now iterates through different memorders instead

metagraph/src/annotation/binary_matrix/base/binary_matrix.cpp
hmusta: good point, maybe we should just have a safe single-threaded implementation here and leave it up to the other methods to implement a multi-threaded one if there's a nice way to do it

metagraph/src/annotation/binary_matrix/base/binary_matrix.hpp
hmusta: The columns may not be called in the same order as `columns`, so I set up the callback to have a pair with the column index.

metagraph/tests/test_vector_algorithm.cpp
hmusta: I reorganized these tests to get rid of the redundant checks

metagraph/src/cli/load/load_annotated_graph.cpp
hmusta: doing it this way is kind of problematic because if `wrap_primary == false` and one passes a primary graph, we still need to wrap it and this function then ends up returning something inexpected

metagraph/src/annotation/binary_matrix/base/binary_matrix.cpp
karasikov: Yes, `get_column` is exposed, so you can always write this parallel loop in the caller.

metagraph/src/graph/representation/masked_graph.hpp
```diff
@@ -115,8 +115,8 @@ class MaskedDeBruijnGraph : public DeBruijnGraph {
         kmers_in_graph_ = std::move(mask);
     }
 
-    typedef std::function<void(const std::function<void(node_index, bool)>&)> GenerateNodes;
-    virtual void update_mask(const GenerateNodes &generate_nodes,
+    typedef std::function<void(const std::function<void(node_index, bool /* new value */)>&)> GenerateNodeUpdates;
```
karasikov, line 118: "new value" means added or removed node?

metagraph/src/cli/config/config.hpp
```diff
@@ -84,7 +84,7 @@ class Config {
     unsigned int bloom_max_num_hash_functions = 10;
     unsigned int num_columns_cached = 10;
     unsigned int max_hull_forks = 4;
-    unsigned int parallel_assemblies = -1;
+    unsigned int parallel_assemblies = -1; // if not set, run one single-threaded assembly per thread
```
karasikov, line 87: So why is it not 1 by default?

metagraph/src/graph/representation/masked_graph.hpp
hmusta: should I call it "update value" instead? It's the value that should be stored at that position

metagraph/src/cli/config/config.hpp
hmusta: Given N threads, the algorithm parallelizes better if you have N assemblies on one thread each, so I made that the default.

Though I agree that there is the problem that you don't know the number of experiments in advance (unless I change the specification of label-mask-file to start with a single number being the number of experiments), so this default option doesn't work if you just want to run one experiment

hmusta: Maybe the behaviour will be more expected if I just set it to run one assembly at a time (giving it all threads) by default

karasikov: Ah, got it, I misread it first. Maybe write ` // run `parallel` assemblies by default (one per thread).

metagraph/src/graph/representation/masked_graph.hpp
karasikov: Ah, I'm confused again. What does this function actually do?

karasikov: calls all nodes in the full graph and indicates those from the subgraph with 1/0?

hmusta: It generates a bunch of `(node_index, bool)` pairs. So for each `(i,j)` pair, it does `mask[i] = j`

karasikov: ah, I see. Why not pass an abstract `bitmap`?
You can define it from a callback, so the complexity will be equivalent, but much easier to read.

metagraph/src/graph/representation/masked_graph.cpp
```diff
@@ -278,5 +243,49 @@ bool MaskedDeBruijnGraph::operator==(const DeBruijnGraph &other) const {
...
+        generate_nodes([&](node_index i, bool val) {
+            if (val) {
+                set_bit(mask_updated.data(), i, async, memorder);
+            } else {
+                unset_bit(mask_updated.data(), i, async, memorder);
+            }
+        });
+        kmers_in_graph_ = std::make_unique<bit_vector_stat>(std::move(mask_updated));
+    }
+}
```
karasikov, line 288: This looks too complex and I bet it's suboptimal in some cases. So why do we restrict ourselves to this interface?

I think it's better to just return a pointer to the bitmap, so it can be updated externally by the user who may have extra information (for instance, how many bits are going to be updated. If too many, it faster to reinitialize the whole thing).

Alternatively, we can allow only the reinitialization of the whole thing.

metagraph/src/graph/representation/masked_graph.hpp
karasikov: I see now that bitmap wouldn't be sufficient. There is a design issue, this interface is not ideal (see the other comment in .cpp file)

metagraph/src/graph/representation/masked_graph.cpp
karasikov: So simply return a pointer to the bitmap `return &kmers_in_graph_;` and move this code for updating to the caller.

hmusta: The easiest thing will be to just return a non-const reference to the underlying mask and let an external function deal with modifying it.

hmusta: To give motivation for why I'm doing all of this, I want to be able to call `MaskedDeBruijnGraph::call_unitigs` and modify the mask in the callback while avoiding as much RAM overhead as possible.

If the underlying graph is DBGSuccinct, then a copy of the mask is made before traversal starts, so I can modify the mask in the callback without disturbing anything. For any other graph representation, modifying the mask in the callback will break the traversal, so I have to populate a new bit vector first, then swap the masks.

Since this distinction between the graph types is an implementation detail of MaskedDeBruijnGraph, I tried to put all of this logic in `update_mask`.

karasikov: Then return a non-const reference to the `unqiue_ptr` to allow the user to reinitialize the whole thing and change representations.

karasikov: Yeah, I see the need. Just the interface is too restricted and complex.

PR #190, opened 2020-08-10
Restrict RAM usage per alignment

hmusta:
This introduces a new parameter for limiting the amount of RAM that can be taken by an alignment. It also adds some members to the Alignment and QueryAlignment classes for storing alignment RAM statistics.

### DIFF
--- a/metagraph/src/cli/server.cpp
+++ b/metagraph/src/cli/server.cpp
@@ -130,6 +130,8 @@ std::string process_search_request(const std::string &received_message,
     config.alignment_max_nodes_per_seq_char = json.get(
         "max_num_nodes_per_seq_char",
         config.alignment_max_nodes_per_seq_char).asDouble();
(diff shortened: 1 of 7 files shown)


metagraph/src/cli/server.cpp
```diff
@@ -130,6 +130,8 @@ std::string process_search_request(const std::string &received_message,
     config.alignment_max_nodes_per_seq_char = json.get(
         "max_num_nodes_per_seq_char",
         config.alignment_max_nodes_per_seq_char).asDouble();
+    config.alignment_max_ram = json.get("max_ram_per_alignment",
+                                        config.alignment_max_ram).asDouble();
```
karasikov, line 134: this is not safe, don't pass it with api

hmusta: Good point. We could then write in the manual that the server has this limit and that the CLI version should be used if one needs more RAM for their alignment.

PR #191, opened 2020-08-11
Add hull to query graph. Convert the query graph to DBGSuccinct if min seed length < k

hmusta:


metagraph/src/cli/query.hpp
```diff
@@ -28,24 +29,43 @@ class Config;
...
+/**
+ * Construct a query graph
+ * @param anno_graph the input annotated de Bruijn graph
+ * @param call_sequences generate sequences to be queried against anno_graph
+ * @param discovery_fraction the minimum fraction at which a generated sequence
+ *                           must be present in anno_graph to be considered
+ * @param num_threads number of threads to use
+ * @param canonical if true, the returned query graph is a canonical graph
+ * @param sub_k also include k-mers whose suffixes of length sub_k match sub_k-mers
+ *              in the generated sequences. This is only applicable for DBGSuccinct
```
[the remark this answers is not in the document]
hmusta: There's a command line argument `--align-max-num-seeds-per-locus` which is set to 1 by default.

For a given k-mer, it finds the max k' s.t. there exist n > 0 nodes with a suffix match of length k'. If n > max-num-seeds-per-locus, then it ignores that k-mer and moves on (as a way to prevent too many seeds from being processed).

metagraph/src/cli/query.hpp
```diff
@@ -34,13 +34,13 @@ using StringGenerator = std::function<void(std::function<void(const std::string
  * @param anno_graph the input annotated de Bruijn graph
  * @param call_sequences generate sequences to be queried against anno_graph
  * @param discovery_fraction the minimum fraction at which a generated sequence
- *                           must be present in anno_graph to be included into the query graph
+ *                           must be present in anno_graph to be considered
```
karasikov: remove

metagraph/src/cli/query.cpp
```diff
@@ -224,6 +418,34 @@ slice_annotation(const AnnotatedDBG::Annotator &full_annotation,
...
+                                                 size_t num_threads = 1) {
+    BOSSConstructor constructor(full_dbg.get_k() - 1, canonical, 0, "", num_threads);
+    for (size_t i = 0; i < contigs.size(); ++i) {
+        const std::string &contig = contigs[i].first;
+        const auto &nodes_in_full = contigs[i].second;
+        constructor.add_sequences([&](const CallString &callback) {
+            auto it = nodes_in_full.begin();
+            while ((it = std::find_if(it, nodes_in_full.end(), [&](auto i) { return i; }))
+                    < nodes_in_full.end()) {
+                auto next = std::find(it, nodes_in_full.end(), 0);
```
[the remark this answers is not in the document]
hmusta: Though with the approach you suggested, several consecutive `std::find` calls are made until a matching k-mer is found. Do you think that's more efficient?

We could do something like this instead
```c++
do {
    end = std::find(begin, nodes_in_full.end(), DeBruijnGraph::npos);
    if (begin != end) {
        constructor.add_sequence(std::string_view(
            contig.data() + (begin - nodes_in_full.begin()),
            end - begin + k - 1
        ));
        begin = end + 1;
    } else {
        begin = std::find_if(begin + 1, nodes_in_full.end(),
                             [&](node_index i) {
                                 return i != DeBruijnGraph::npos;
                             });
    }
} while (end < nodes_in_full.end());
```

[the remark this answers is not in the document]
karasikov: I think they both a efficient enough.

The question is about readability and how easy it is to make a mistake in this code.

The simpler a thing looks, the harder it is to make a mistake there and easier to check.

metagraph/src/cli/config/config.cpp
```diff
@@ -184,6 +184,8 @@ Config::Config(int argc, char *argv[]) {
             align_both_strands = true;
         } else if (!strcmp(argv[i], "--align-edit-distance")) {
             alignment_edit_distance = true;
+        } else if (!strcmp(argv[i], "--max-hull-depth")) {
+            max_hull_depth = atoll(get_value(i++));
```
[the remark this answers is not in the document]
karasikov: It's not really a radius, I think this would be the most confusing

metagraph/src/cli/query.hpp
```diff
@@ -28,24 +29,43 @@ class Config;
...
+ *              in the generated sequences. This is only applicable for DBGSuccinct
+ * @param max_hull_forks when adding a hull to the query graph from anno_graph,
+ *                       restrict traversal to at most max_hull_forks forking nodes
+ * @param max_hull_depth nodes added to the hull must be within the max_hull_depth
+ *                       neighbourhood of k-mers in the generated sequences
+ * @param max_hull_depth_per_seq_char restrict the traversal around a generated
+ *                                    sequence's k-mers to its
+ *                                    len(seq) * max_hull_depth_per_seq_char
+ *                                    neighbourhood.
+ */
```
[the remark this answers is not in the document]
hmusta: The idea here is that I wanted to have the traversal depth/distance be a function of the length of the input sequences, though this is worst-case behaviour since the traversal stops if it finds a matching k-mer.

But if the user passes in a very long query sequence, I still wanted to have `max-hull-depth` as a hard cutoff on the traversal.

metagraph/src/graph/alignment/dbg_aligner.hpp
```diff
@@ -14,6 +14,9 @@
 
 
 namespace mtg {
+
+using mtg::common::logger;
```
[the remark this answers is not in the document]
hmusta: In general, yes. The problem in this specific case is that I didn't want to put the implementations of the `DBGAligner` methods into a cpp file to avoid having a bunch of instantiations for each combination of Seeder and Extender.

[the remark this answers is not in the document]
karasikov: then you shouldn't use `mtg::common::`, I guess

metagraph/tests/graph/test_canonical_dbg.cpp
```diff
@@ -616,6 +616,52 @@ TYPED_TEST(CanonicalDBGTest, CallUnitigsRegular) {
     }
 }
 
+TYPED_TEST(CanonicalDBGTest, CallUnitigsRegularCanonical) {
+    for (size_t num_threads = 1; num_threads < 5; num_threads += 3) {
```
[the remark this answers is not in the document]
hmusta: Since this shows up in a lot of places in existing unit tests, I'll make this change outside this pull request

metagraph/src/cli/query.cpp
```diff
@@ -99,6 +119,178 @@ std::string QueryExecutor::execute_query(const std::string &seq_name,
...
+
+        assert(!last_k);
+
+        dbg_succ.call_nodes_with_suffix(
+            std::string_view(contig.data() + i, k),
+            [&](node_index node, size_t seed_length) {
+                last_k = seed_length;
+                callback(dbg_succ.get_node_sequence(node), { node });
+            },
+            sub_k, 1 // max num nodes per suffix
```
karasikov, line 176:
```suggestion
            sub_k, 1 // fetch only a single node per suffix
```

metagraph/src/cli/query.cpp
```diff
@@ -99,6 +103,196 @@ std::string QueryExecutor::execute_query(const std::string &seq_name,
...
+
+    size_t last_k = 0;
+    for (size_t i = 0; i < nodes_in_full.size(); ++i) {
+        if (nodes_in_full[i] != DeBruijnGraph::npos) {
+            last_k = 0;
+            continue;
+        }
+
+        if (last_k >= sub_k) {
+            i += last_k - sub_k;
```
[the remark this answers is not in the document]
karasikov: I still don't understand why you think it's ok to skip these

metagraph/src/cli/query.cpp
```diff
@@ -99,6 +119,178 @@ std::string QueryExecutor::execute_query(const std::string &seq_name,
...
+    }
+}
+
+struct HullUnitig {
+    std::string unitig;
+    std::vector<node_index> path;
+    size_t fork_count;
+};
+
+typedef tsl::hopscotch_map<node_index, size_t> NodeToDistanceMap;
```
karasikov, line 187: no need to make this typedef, you use it just once

metagraph/src/cli/query.cpp
```diff
@@ -224,6 +416,30 @@ slice_annotation(const AnnotatedDBG::Annotator &full_annotation,
...
 
+template <class Contigs>
+std::shared_ptr<DBGSuccinct> convert_to_succinct(const Contigs &contigs,
+                                                 size_t k,
+                                                 bool canonical = false,
+                                                 size_t num_threads = 1) {
+    BOSSConstructor constructor(k - 1, canonical, 0, "", num_threads);
+    for (const auto &[contig, nodes_in_full] : contigs) {
+        auto begin = nodes_in_full.begin();
+        std::vector<DeBruijnGraph::node_index>::const_iterator end;
```
karasikov, line 427: More compact with indexes
```c++
        size_t begin = 0;
        size_t end;
```

karasikov (changes requested):

metagraph/tests/graph/test_canonical_dbg.cpp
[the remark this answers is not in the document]
karasikov: but only for such degenerate loops as this

metagraph/src/cli/query.cpp
```diff
@@ -31,7 +34,24 @@ using namespace mtg::annot::binmat;
...
+                             const graph::align::DBGAlignerConfig *aligner_config,
+                             ThreadPool &thread_pool)
+      : config_(config),
+        anno_graph_(anno_graph),
+        aligner_config_(aligner_config
+                        ? new graph::align::DBGAlignerConfig(*aligner_config)
+                        : nullptr),
+        thread_pool_(thread_pool) {
+    if (aligner_config_)
+        aligner_config_->forward_and_reverse_complement = false;
```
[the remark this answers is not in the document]
hmusta: There is a mode in the `align` regime where only the best alignment of the forward and reverse complement is reported (it can potentially speed things up because the best alignment score for the forward alignment is used as a lower bound for the reverse complement alignment), but this is kind of a legacy feature since it's not relevant for canonical graphs or `CanonicalDBG` graphs. Maybe I should just remove it...

metagraph/src/cli/config/config.hpp
```diff
@@ -83,10 +83,13 @@ class Config {
     unsigned int port = 5555;
     unsigned int bloom_max_num_hash_functions = 10;
     unsigned int num_columns_cached = 10;
+    unsigned int max_hull_forks = 0;
 
     unsigned long long int query_batch_size_in_bytes = 100'000'000;
     unsigned long long int num_rows_subsampled = 1'000'000;
     unsigned long long int num_singleton_kmers = 0;
+    unsigned long long int max_hull_depth
+        = std::numeric_limits<unsigned long long int>::max();
```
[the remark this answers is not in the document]
hmusta: By default, the `max_hull_depth` will be restricted to 2 * read_length because of the other parameters. I can change `max_hull_forks` to 4 by default.

hmusta:
I ended up rewriting the k-mer extraction parts of the query graph construction to reduce the amount of mapping that's done to create the hull.

Also, I noticed while testing the alignment in `dev` that it currently doesn't intersect the query graph with the full graph for non-canonical graphs, so the alignments were incorrect.

metagraph/src/cli/server.cpp
```diff
@@ -141,15 +141,16 @@ std::string process_search_request(const std::string &received_message,
...
+        aligner_config.reset(new graph::align::DBGAlignerConfig(
+            initialize_aligner_config(anno_graph.get_graph(), config)
+        ));
 
         // the fwd_and_reverse argument in the aligner config returns the best of
         // the forward and reverse complement alignments, rather than both.
         // so, we want to prevent it from doing this
-        auto &aligner_config = const_cast<graph::align::DBGAlignerConfig&>(aligner->get_config());
-        aligner_config.forward_and_reverse_complement = false;
+        aligner_config->forward_and_reverse_complement = false;
```
karasikov, line 153: why do you want to prevent this?

hmusta: this flag enables a mode in the aligner which returns the best alignment from the forward and reverse complement, which is different from the behaviour of query mode that returns both

karasikov: Why is it set to `true` by default?

karasikov: hold on, it is actually set to `false`, so we don't need to reset it here, right?

karasikov: I'll add an assert in QueryExecutor instead, so we make sure the value is correct

hmusta: actually, you're right, we don't need this at all. I think at one point, `forward_and_reverse_complement` was being set when the `--fwd-and-reverse` flag was set, but that was changed so it gets triggered by a different flag

hmusta: so yea, this isn't needed anymore

metagraph/src/cli/query.cpp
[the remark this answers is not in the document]
karasikov: Got it, it would have been clear if this there was an assertion here, so we expect that the config is compatible to the logic in this part.

metagraph/src/cli/query.cpp
```diff
@@ -279,224 +469,368 @@ construct_query_graph(const AnnotatedDBG &anno_graph,
...
+        std::vector<node_index> &nodes = contigs[i].second;
+        std::vector<node_index> &dual_nodes = rc_contigs[i].second;
+        std::reverse(dual_nodes.begin(), dual_nodes.end());
+        assert(nodes.size() == dual_nodes.size());
+
+        for (size_t j = 0; j < nodes.size(); ++j) {
+            if (!nodes[j]) {
+                nodes[j] = dual_nodes[j];
+            } else if (dual_nodes[j]) {
+                nodes[j] = std::min(nodes[j], dual_nodes[j]);
```
hmusta, line 746: `std::min` won't work here for the hash ordered graph, since it uses the min of the k-mer encodings rather than the node index to determine canonical k-mers

karasikov: Yes, thanks.

I thought we would use a generic `map_to_annotated_nodes` moved to `AnnotatedDBG`, but it's easier to call the existing method for now.

metagraph/src/cli/query.cpp
```diff
@@ -404,6 +403,14 @@ void add_nodes_with_suffix_matches(const DBGSuccinct &full_dbg,
...
                                     callback, sub_k, max_num_nodes_per_suffix);
 
+        if (check_reverse_complement && !full_dbg.is_canonical_mode()) {
+            std::string rev_contig = contig;
+            reverse_complement(rev_contig);
+            call_suffix_match_sequences(full_dbg, rev_contig,
+                                        map_sequence_to_nodes(full_dbg, rev_contig),
+                                        callback, sub_k, max_num_nodes_per_suffix);
+        }
+
```
karasikov, line 413: Maybe we should do this even if full_dbg is canonical? Because suffixes in dual nodes are prefixes in the forward ones.

hmusta: It should still be fine in most cases, but we should still do this for all graph types, since it will fail if the suffix match is to a dummy source k-mer.

karasikov: Hm, not sure I got it? Do you mean we should fix something here or you think it's good?

hmusta: I think that it should be fixed, as in, we should check the suffixes for dual nodes in all cases when the batch graph is canonical

hmusta: I just wanted to point out that it's a corner case

metagraph/src/cli/query.cpp
```diff
@@ -484,16 +827,41 @@ int query_graph(Config *config) {
...
-                                        anno_graph);
+inline void query_sequence(size_t id, std::string name, std::string seq,
+                           const AnnotatedDBG &anno_graph,
+                           const Config &config,
+                           const align::DBGAlignerConfig *aligner_config,
+                           const std::function<void(const std::string&)> &callback) {
+    if (aligner_config) {
+        auto matches = build_aligner(anno_graph.get_graph(), *aligner_config)->align(seq);
+        if (matches.size()) {
+            auto &match = matches[0];
```
karasikov, line 838: Does it always return multiple top matches, or we can tell the aligner that we want just the top alignment? Or the complexity is the same either way?

metagraph/src/cli/query.cpp
```diff
@@ -484,16 +827,41 @@ int query_graph(Config *config) {
...
-                                        config.suppress_unlabeled, config.num_top_labels,
-                                        config.discovery_fraction, config.anno_labels_delimiter,
-                                        anno_graph);
+inline void query_sequence(size_t id, std::string name, std::string seq,
+                           const AnnotatedDBG &anno_graph,
+                           const Config &config,
+                           const align::DBGAlignerConfig *aligner_config,
+                           const std::function<void(const std::string&)> &callback) {
+    if (aligner_config) {
+        auto matches = build_aligner(anno_graph.get_graph(), *aligner_config)->align(seq);
```
karasikov, line 836: should it be called `auto alignments = ...->align(seq)`?

metagraph/src/cli/query.cpp
```diff
@@ -540,10 +857,35 @@ int query_graph(Config *config) {
...
 inline std::string query_sequence(size_t id,
-                                  const std::string &name,
-                                  const std::string &seq,
+                                  std::string &name, std::string &seq,
                                   const AnnotatedDBG &anno_graph,
-                                  const Config &config) {
+                                  const Config &config,
+                                  const align::DBGAlignerConfig *aligner_config = nullptr) {
+    if (aligner_config) {
+        auto matches = build_aligner(anno_graph.get_graph(), *aligner_config)->align(seq);
```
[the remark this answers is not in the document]
karasikov: Hm, but it doesn't matter whether we pass `aligner_config` or an already constructed `aligner`, right?

Or we can call `align` only once?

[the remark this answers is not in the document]
karasikov: ah, never mind, it's indeed easier to pass `aligner_config`.

metagraph/src/cli/query.cpp
hmusta: You can tell the aligner how many top matches to return, but here we only take the top one. There's a check in `config` which tells the user about this and exits if they change the default number of returned alignments.

metagraph/src/cli/query.cpp
hmusta: sure, I called it `matches` just to be shorter

metagraph/src/cli/query.cpp
karasikov: So we can add an assertion here?

karasikov (approved):

metagraph/src/cli/query.cpp
```diff
@@ -811,6 +812,10 @@ int query_graph(Config *config) {
     assert(config->infbase_annotators.size() == 1);
 
     auto graph = load_critical_dbg(config->infbase);
+
+    if (config->canonical)
+        graph = std::make_shared<CanonicalDBG>(graph);
```
karasikov: do we need this if alignment is off?

hmusta: this is an old commit. now, it only does this if the underlying graph is not canonical

karasikov: ok, but could you check it if this is not going to make it slower for k-mer query without alignment?

metagraph/src/graph/annotated_dbg.cpp
```diff
@@ -348,6 +350,11 @@ ::call_annotated_nodes(const std::string &label,
 }
 
 bool AnnotatedSequenceGraph::check_compatibility() const {
+    if (dynamic_cast<const CanonicalDBG*>(graph_.get())) {
+        return dynamic_cast<const CanonicalDBG&>(*graph_).get_graph().max_index()
+            == annotator_->num_objects();
+    }
```
karasikov, line 356: maybe we can redefine max index: let it return maximum node index in map_to_nodes, that is, only for annotated nodes.
Then we wouldn't need to do this hack

metagraph/src/cli/query.cpp
```diff
@@ -451,33 +811,27 @@ int query_graph(Config *config) {
     assert(config->infbase_annotators.size() == 1);
 
     auto graph = load_critical_dbg(config->infbase);
+
+    if (config->canonical && !graph->is_canonical_mode())
+        graph = std::make_shared<CanonicalDBG>(graph);
```
karasikov, line 816: Search without alignment can be done without this wrapper.
Though it's ok to have it if it doesn't slow it down. Did you benchmark?

metagraph/src/graph/annotated_dbg.cpp
hmusta: Maybe a cleaner approach would be to add a method `max_annotated_index` to `AnnotatedDBG`

hmusta: That'll cause confusion, though, since the max index in a wrapped graph will no longer be correct.

Maybe a less confusion approach (which would in effect be moving this hack somewhere else) would be to add a method `max_annotated_index` to `AnnotatedDBG`

metagraph/src/cli/query.cpp
hmusta: comparing with align --map --count-kmers on the 750 subset with the metasub sample

Canonical graph
Elapsed (wall clock) time (h:mm:ss or m:ss): 0:04.16

Primary graph --fwd-and-reverse
Elapsed (wall clock) time (h:mm:ss or m:ss): 0:06.35

Primary graph --canonical
Elapsed (wall clock) time (h:mm:ss or m:ss): 0:06.15

Primary graph --canonical --primary-kmers
Elapsed (wall clock) time (h:mm:ss or m:ss): 0:06.03

metagraph/src/graph/annotated_dbg.cpp
karasikov: It should definitely be in the same class as `map_to_nodes`. It's better if they both move to AnnotatedDBG, but this is not a completely trivial refactoring. So I think for now it's better to keep them both in graphs.

metagraph/src/cli/query.hpp
```diff
@@ -18,6 +18,7 @@ namespace graph {
     class AnnotatedDBG;
     namespace align {
         class IDBGAligner;
```
hmusta, line 20: no longer needed
```suggestion
```

metagraph/src/graph/annotated_dbg.cpp
karasikov: Turns out, `max_index()` is already used in many places as a bound on indexes in `map_to_nodes_sequentially`, so we can't easily redefine it.

In the future, we need to add another method that returns max index called in `map_to_nodes`.

PR #194, opened 2020-09-01
Add a new annotation format: row_diff and the corresponding converter from RowCompressed

danieldanciu:


metagraph/src/annotation/binary_matrix/row_diff/row_diff.cpp
```diff
@@ -0,0 +1,129 @@
...
+        assert(boss_edge > 1 && w != 0);
+
+        // fwd always selects the last outgoing edge for a given node
+        boss_edge = boss.fwd(boss_edge, w % boss.alph_size);
+        row = graph::AnnotatedSequenceGraph::graph_to_anno_index(
+                graph_->boss_to_kmer_index(boss_edge));
+        merge(&result, get_diff(row));
+    };
+    return result;
+}
```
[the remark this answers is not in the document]
hmusta: We wanted to keep AnnotatedDBG as simple as possible, so we can keep it as-is (using it simply to map sequences to nodes then call `get_rows` or `count_labels`), then implement everything in those downstream calls.

metagraph/src/graph/representation/base/sequence_graph.hpp
```diff
@@ -170,14 +170,18 @@ class DeBruijnGraph : public SequenceGraph {
...
+     * (canonical/non-canonical). That is, skip a k-mer if its reverse-complement has been
+     * extracted.
+     * @param select_last_edge if true, when reaching a bifurcation during traversal,
+     * either traverse the last edge (if not visited), or stop the sequence.
      */
     virtual void call_sequences(const CallPath &callback,
                                 size_t num_threads = 1,
-                                bool kmers_in_single_form = false) const;
+                                bool kmers_in_single_form = false,
+                                bool select_last_edge = false) const;
```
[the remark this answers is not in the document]
hmusta: This change actually makes paths shorter. Its purpose was to have a guarantee that if an internal node in a path is a branching node, then the next node taken in the path has to be the last outgoing k-mer.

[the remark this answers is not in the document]
hmusta: Though I don't know how much storage this saves relative to just having all unitig end-points as termini for storing full annotation rows.

metagraph/src/graph/representation/base/sequence_graph.cpp
```diff
@@ -349,7 +349,8 @@ void call_sequences(const DeBruijnGraph &graph,
 
 void DeBruijnGraph::call_sequences(const CallPath &callback,
                                    size_t num_threads,
-                                   bool kmers_in_single_form) const {
+                                   bool kmers_in_single_form,
+                                   bool /* select_last_edge */) const {
```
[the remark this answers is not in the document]
danieldanciu: I didn't bother implementing it as this code path was never used. I implemented first-edge traversal for this class, too, as it turns out that first-edge traversal does have a significant impact on the annotation size.

metagraph/src/annotation/binary_matrix/row_vector/vector_row_binmat.cpp
```diff
@@ -14,7 +14,7 @@ namespace annot {
 namespace binmat {
 
 template <typename RowType>
-VectorRowBinMat<RowType>::VectorRowBinMat(std::vector<RowType>&& rows,
+VectorRowBinMat<RowType>::VectorRowBinMat(Vector<RowType>&& rows,
```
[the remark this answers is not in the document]
karasikov: right, I confused it, the problem will be if we do the other way around: change something when folly is off, and then see that it doesn't compile with folly on

metagraph/src/annotation/annotation_converters.cpp
```diff
@@ -978,5 +980,136 @@ void convert_to_row_annotator(const ColumnCompressed<std::string> &source,
...
+            [&](const std::string &, const std::vector<uint64_t> &path) {
+                assert(!path.empty());
+                std::vector<uint64_t> anno_ids;
+                for (const uint64_t node_id : path) {
+                    assert(node_id <= nnodes);
+                    anno_ids.push_back(
+                            graph::AnnotatedSequenceGraph::graph_to_anno_index(node_id));
+                }
+                std::vector<Vector<uint64_t>> rows
+                        = annotation.get_matrix().get_rows(anno_ids);
```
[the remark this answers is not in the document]
karasikov: Correct, if you eventually need to get the rows as a bunch of vectors,  eventually, there is no need to call slice_rows, `get_rows` is the exact function for this.

Another thing is that often we can save time by working with batches of rows and keeping all them in a single vector

metagraph/src/graph/representation/base/sequence_graph.hpp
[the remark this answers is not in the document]
karasikov: ok, let's work with contigs then.

One thing I suggest to change in this case: remove the flag `bool select_last_edge` and make this behavior the default for succinct graphs (there is no overhead from it, right?). So the signature won't be changed and there will be no dummy flag in other graph representation. And the algorithm will rely on the correct order of traversal for succinct graphs.

metagraph/src/cli/load/load_annotated_graph.cpp
```diff
@@ -20,16 +21,33 @@ std::unique_ptr<AnnotatedDBG> initialize_annotated_dbg(std::shared_ptr<DeBruijnG
...
+        }
+        const Config::AnnotationType input_anno_type
+                = parse_annotation_type(config.infbase_annotators.at(0));
+        // row_diff annotation is special, as it must know the graph structure
+        if (input_anno_type == Config::AnnotationType::RowDiff) {
+            auto dbg_graph = dynamic_cast<DBGSuccinct *>(graph.get());
+            if (!dbg_graph) {
+                logger->error("Only dbg graphs are supported for row_diff annotations");
+                std::exit(1);
+            }
```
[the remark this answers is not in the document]
karasikov: In this case, write `Only succinct de Bruijn graph representations are supported ...`

metagraph/src/annotation/binary_matrix/bin_rel_wt/bin_rel_wt.hpp
```diff
@@ -21,18 +21,18 @@ class BinRelWT : public BinaryMatrix {
 
     BinRelWT(std::vector<std::unique_ptr<bit_vector>>&& columns);
 
-    uint64_t num_columns() const;
-    uint64_t num_rows() const;
+    uint64_t num_columns() const override;
+    uint64_t num_rows() const override;
```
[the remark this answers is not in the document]
karasikov: I agree, sometimes this may be helpful, but if we write this everywhere, we waste a lot of horizontal space...

metagraph/tests/annotation/test_converters.cpp
```diff
@@ -25,15 +25,20 @@ const std::string test_dump_basename_row_compressed_to_rowflat = test_dump_basen
...
-        initial_annotation->add_labels({ 3 }, {"Label1", "Label2", "Label8"});
-        initial_annotation->add_labels({ 4 }, {"Label2"});
+        initial_annotation->add_labels({ 0 }, { "Label0", "Label2", "Label8" });
+        initial_annotation->add_labels({ 2 }, { "Label1", "Label2" });
+        initial_annotation->add_labels({ 3 }, { "Label1", "Label2", "Label8" });
+        initial_annotation->add_labels({ 4 }, { "Label2" });
+
+        graph = new graph::DBGSuccinct(3);
+        graph->add_sequence("ACGTCAC");
+        graph->mask_dummy_kmers(1, false);
```
karasikov, line 41: Why did you change this test? Please add new tests but don't change the old ones, unless it's some kind of refactoring. But this change actually makes a conceptual difference.

In practice, we never mask out the dummy k-mers in our graphs. This is a legacy feature.

karasikov (changes requested):

metagraph/src/annotation/annotation_converters.cpp
```diff
@@ -978,5 +980,111 @@ void convert_to_row_annotator(const ColumnCompressed<std::string> &source,
...
+#pragma omp critical
+            {
+                offset = node_diffs.size();
+                node_diffs.insert(node_diffs.end(), diffs.begin(), diffs.end());
+            }
+            for (uint32_t i = 0; i < rows.size(); ++i) {
+                pos[anno_ids[i]].first += offset;
+            }
+
+            terminal_count.fetch_add(1, std::memory_order_relaxed);
```
karasikov, line 1048: this is a really negligible computation, I would just use atomic variables to make it easier to read

metagraph/src/annotation/annotation_converters.cpp
```diff
@@ -978,5 +980,111 @@ void convert_to_row_annotator(const ColumnCompressed<std::string> &source,
...
+            terminal[anno_ids.back()] = 1;
+            // add the last row plus one termination bit for each row
+            boundary_size.fetch_add(diffs.size() + rows.size(), std::memory_order_relaxed);
```
karasikov, line 1051: If the last node of a contig is a merge, it's not necessary to make it an anchor.

An anchor must be created only if there are no outgoing edges.

This should bring about the same reduction as you got by fetching contigs instead of unitigs (which essentially translates the same idea to forks, and I assume there are about the same number of forks and merges in graphs).

karasikov: Hm, I'm thinking how this can be implemented, and think both forks and merges can be handled the same way here. So you can fetch unitigs from graph and then check if there are outgoing (incoming) edge from the last (first) node.

Why fetch unitigs instead of contigs -- no need to make assumptions about how the contigs are traversed (no need to add that flag `bool select_first_edge`), and also, `pull_unitigs` allows for more efficient parallelization and will have a more scalable implementation in the future.

metagraph/tests/annotation/test_converters.cpp
danieldanciu: I didn't change the test, in the sense that the previous tests are not in any way affected by my change. I simply added a graph to the test so that we can also test row-diff annotations.

karasikov: I think the graph had already been there, and I guess used by other annotations as well. You added the mask, which changed the graph, so effectively changed the tests too. Did I miss something?

danieldanciu: Yes, you missed the fact that I added the graph :)

karasikov: oh, indeed. For some reason, I thought the graph had been there.

metagraph/src/graph/representation/base/sequence_graph.cpp
```diff
@@ -113,7 +113,8 @@ void call_sequences_from(const DeBruijnGraph &graph,
                          ProgressBar &progress_bar,
                          bool call_unitigs,
                          uint64_t min_tip_size,
-                         bool kmers_in_single_form) {
+                         bool kmers_in_single_form,
+                         bool select_first_edge) {
```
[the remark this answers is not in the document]
karasikov: my point is not to change first to last (either is fine, it's equivalent), but to make this a default and remove `bool select_first_edge`

metagraph/src/annotation/annotation_converters.cpp
hmusta: I suggested this change because there are several atomic variables in this section which are being incremented frequently, so using `operator++` causes their memory order to be `std::memory_order_seq_cst`. In some contexts, this can lead to performance loss because we are unnecessarily forcing all of these operations to be performed in order.

In this context, however, the graph traversal, `get_rows`, and `sort` calls dominate the runtime, so most likely this change won't lead to a performance boost. Using `fetch_add` just gives a hint to the reader that we there are no dependences between these increments.

metagraph/src/graph/representation/base/sequence_graph.cpp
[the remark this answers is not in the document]
danieldanciu: As discussed last night, we can't just remove `select_first/last_edge` without affecting the length of the contigs. If we make it a default, we'll be shortening contigs that could be otherwise continued. Am I missing something?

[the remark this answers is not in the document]
karasikov: I see now. Ok, in this case, let's copy the entire function, so instead of changing all the interfaces we add a new method to boss.cpp.

It's fine that the code is copied, this method is suboptimal anyway and the function will be removed as soon as we figure out the best way of assigning anchor nodes.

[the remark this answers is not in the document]
karasikov: I just don't like to further complicate the interfaces for the sake of a new suboptimal feature, which will be changed in the future.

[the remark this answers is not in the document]
karasikov: also, then there will be no confusion about the other graph representations, it will be immediately clear that they don't support this annotation (again, will be fixed in the future).

metagraph/src/annotation/annotation_converters.cpp
karasikov: There is an efficient parallel algorithm for this, can discuss in the next meeting.

Basically, we start parallel traversal from random nodes (first, all forks, then all other, as usual), and make every 100-th node an anchor. Also, we mark each traversed node in another bitmap (which indicates the nodes within <=100 steps to the next anchor). And we terminate the traversal once we reach a node which is marked in this extra bitmap.

So we never traverse the same node twice (thanks to the first bitmap of visited nodes), hence the time complexity is the same as for `call_sequences`.
The second bitmap, on the other hand, ensures that there are no paths of size >200 without an anchor.

With some other tricks, we can probably guarantee a lower bound on average too, but this is a bit more tricky.

Also, in fact, we don't need two traversals. Everything can be done in one. So we can implement this in this PR. Basically, this will be a slightly modified version of the current algorithm, with an extra masking bitmap and a few more checks. So creating a variation of the original traversal algorithm as a separate function seems to be the best approach.

karasikov (approved):
I would make a copy of BOSS::call_paths already now, as it's clear that the algorithm should be improved, and trying to make everything as a single function with many parameters doesn't make much sense.

metagraph/src/annotation/annotation_converters.cpp
danieldanciu: I prefer having the option to not set maximum path lengths, as the compression is significantly better if there is no max path length (50% by max length 100). If we query not too short sequences, the cost of the path length will be amortized very quickly, so in many applications this makes sense. 

On the other hand, if I am not mistaken, the algorithm you proposed doesn't guarantee that the maximum path length is 200 - in the worst case the maximum path length can be n, where n is the number of edges in the graph (if we keep chaining paths into paths that were already visited).

karasikov: of course, it's a parameter, I used 100 and 200 as an example

hmusta (approved):

metagraph/src/annotation/annotation_converters.cpp
danieldanciu: What I meant was that I don't think this method can provide an upper bound for the length of a path, no matter what the value of the parameter is. In the worst case, you can always have a path of length n (total number of edges), even if you  set the maximum path length to 1.

karasikov: It does, that's why bitmaps, otherwise we would use just one.
If we reach a node marked in the second bitmap while traversing a path, we don't mark the nodes of this path in that second bitmap, because they don't necessarily satisfy the closeness condition (<= 100)

karasikov: > I prefer having the option to not set maximum path lengths, as the compression is significantly better if there is no max path length (50% by max length 100). If we query not too short sequences, the cost of the path length will be amortized very quickly, so in many applications this makes sense.

In this case just set n to a large number, like 1000. But a threshold must be set, otherwise, there are no reasonable guarantees on the query complexity.

PR #195, opened 2020-09-07
Fix race condition in counting dummy sink and dummy source k-mers

danieldanciu:


metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
```diff
@@ -473,20 +470,22 @@ generate_dummy_1_kmers(size_t k,
...
-            num_sink++;
         }
     }
 
+    uint64_t num_sink = 0;
+    uint64_t num_source = 0;
     for (TAlphabet i = 0; i < alphabet_size; ++i) {
         dummy_sink_chunks[i].finish();
         dummy_l1_chunks[i].finish();
+        num_sink += dummy_sink_chunks[i].size();
```
karasikov, line 487: I think now it's a bit inconsistent for dummy sink k-mers (the redundant dummy sink k-mers are skipped in chunks but not in counts)

danieldanciu: I don't understand what you mean. Can you elaborate?

karasikov: never mind, all good

karasikov (approved):

PR #197, opened 2020-09-08
Add checkpointing to metagraph build

danieldanciu:
As discussed via chat, there are multiple checkpoints after each major operation: generate kmers, generate reverse complements, splitting into chunks, generate dummy1, concatenate chunks, generate dummy2..k.

The client interface exposes only 2 phases: 
  - phase 1 builds everything up to and including dummy k-mers
  - phase 2 finishes the build

I tested by inserting exit statements in the code after each phase and making sure the build is successful on resume.
I also added functional tests for building in phases.


karasikov:
Could you add a test where it builds one graph, kills the process, so it fails to build to the end.
Then it does the same thing for a different graph. Then restarts building both graphs in parallel, and checks that both graphs are correctly generated.

So we check that there are no collisions here and you can correctly generate multiple graphs in parallel on the same machine.

danieldanciu:
I can add a test that builds 2 graphs in parallel (using phases), but I
won't kill the process as that causes the test to be undeterministic.

danieldanciu:
Ok, I added a test that builds 2 graphs at the same time. First it builds
phase 1 for both graphs, then it builds phase2 for both and asserts that
the graphs are correctly built.

metagraph/src/graph/representation/succinct/boss_chunk_construct.hpp
```diff
@@ -28,7 +29,8 @@ class IBOSSChunkConstructor : public IGraphChunkConstructor<BOSS::Chunk> {
                double memory_preallocated = 0,
                mtg::kmer::ContainerType container_type = mtg::kmer::ContainerType::VECTOR,
                const std::filesystem::path &swap_dir = "/tmp/",
-               size_t max_disk_space_bytes = 1e9);
+               size_t max_disk_space_bytes = 1e9,
+               const BuildCheckpoint& checkpoint = BuildCheckpoint(false, "", 2));
```
karasikov, line 33:
```suggestion
               const BuildCheckpoint &checkpoint = BuildCheckpoint(false, "", 2));
```

karasikov: Maybe make the empty prefix mean it's disabled, so you don't have to pass true/false?

metagraph/src/common/sorted_sets/sorted_set_disk_base.cpp
```diff
@@ -91,7 +93,7 @@ void SortedSetDiskBase<T>::start_merging_async() {
     async_worker_.enqueue([file_names, this]() {
         std::function<void(const T &)> on_new_item
                 = [this](const T &v) { merge_queue_.push(v); };
-        merge_files(file_names, on_new_item);
+        merge_files(file_names, on_new_item, false);
```
karasikov, line 96: Who is responsible for cleaning these files up? Does it mean that it never removes the old temp files until all the k-mers are sorted, and thus, the disk usage has grown a lot?

metagraph/src/common/elias_fano.cpp
```diff
@@ -14,29 +14,29 @@
 namespace mtg {
 namespace common {
 
-void concat(const std::vector<std::string> &files, const std::string &result) {
+std::vector<std::string> concat(const std::vector<std::string> &files, const std::string &result) {
     if (files.empty())
-        return;
+        return {};
+    std::vector<std::string> original_files;
```
karasikov, line 20: Please add an empty line after returns
```suggestion
        return {};

    std::vector<std::string> original_files;
```

metagraph/src/cli/config/config.cpp
```diff
@@ -751,6 +772,7 @@ void Config::print_usage(const std::string &prog_name, IdentityType identity) {
             fprintf(stderr, "\t-p --parallel [INT] \tuse multiple threads for computation [1]\n");
             fprintf(stderr, "\t   --disk-swap [STR] \tdirectory to use for temporary files [off]\n");
             fprintf(stderr, "\t   --disk-cap-gb [INT] \tmax temp disk space to use before forcing a merge, in GB [20]\n");
+            fprintf(stderr, "\t   --phase [INT] \tmax where to stop the computation (1=generate kmers, 2= build all) [2]\n");
```
karasikov, line 775: for consistency
```suggestion
            fprintf(stderr, "\t   --phase [INT] \tmax where to stop the computation (1=generate kmers, 2=build all) [2]\n");
```

metagraph/src/cli/build.cpp
```diff
@@ -115,15 +119,24 @@ int build_graph(Config *config) {
...
 
             boss::BOSS::Chunk *next_chunk = constructor->build_chunk();
+
+            if (checkpoint.phase() < 2) { // phase 1 stops after generating dummy k-mers
+                assert(next_chunk == nullptr);
+                return 0;
+            }
             logger->trace("Graph chunk with {} k-mers was built in {} sec",
                           next_chunk->size() - 1, timer.elapsed());
-
```
karasikov, line 126: please keep this empty line, it makes it easier to read

metagraph/integration_tests/test_build.py
```diff
@@ -337,6 +337,76 @@ def test_build_chunks_from_kmc_canonical(self, build):
...
+                                '--disk-swap {tmp_dir} -o {outfile} {input}'.format(
+                exe=METAGRAPH,
+                repr=representation,
+                tmp_dir=tmp_dir,
+                outfile=self.tempdir.name + '/' + name,
+                input=TEST_DATA_DIR + ('/transcripts_1000.fa' if name == 'graph1' else '/transcripts_100.fa')
+            )
+            res = subprocess.run([construct_command], shell=True)
+            self.assertEqual(res.returncode, 0)
+            self.assertTrue(os.path.isfile(self.tempdir.name + '/' + name + '.checkpoint'))
```
karasikov, line 390: please add a check that the checkpoint file is removed after the graph is constructed

metagraph/src/cli/build.cpp
```diff
@@ -115,15 +119,24 @@ int build_graph(Config *config) {
...
+            } else {
+                logger->info("Skipping parsing sequences from input file(s)");
+            }
 
             boss::BOSS::Chunk *next_chunk = constructor->build_chunk();
+
+            if (checkpoint.phase() < 2) { // phase 1 stops after generating dummy k-mers
+                assert(next_chunk == nullptr);
+                return 0;
+            }
```
karasikov, line 137:
```suggestion
            }

```

metagraph/src/cli/build.cpp
```diff
@@ -140,6 +153,7 @@ int build_graph(Config *config) {
             } else {
                 graph_data.reset(next_chunk);
             }
+            checkpoint.done();
```
karasikov, line 156: It's not clear what this method does when I look at its name. Maybe cleanup? Then call it `.remove_checkpoint()`

metagraph/src/cli/config/config.cpp
```diff
@@ -521,6 +523,25 @@ Config::Config(int argc, char *argv[]) {
     if (identity == COMPARE && fnames.size() != 2)
         print_usage_and_exit = true;
 
+    if (identity != BUILD && phase != 2) {
+        std::cerr << "Error: Phases are only supported for building. Remove --phase"
+                  << std::endl;
+        print_usage_and_exit = true;
+    }
```
karasikov, line 530: please add a check that phase is either 1 or 2, and print an error message otherwise

metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
```diff
@@ -347,22 +348,30 @@ using Decoder = common::EliasFanoDecoder<T>;
 template <typename T_REAL>
 std::vector<std::string> split(size_t k,
                                const std::filesystem::path &dir,
-                               const ChunkedWaitQueue<T_REAL> &kmers) {
+                               const ChunkedWaitQueue<T_REAL> &kmers,
+                               BuildCheckpoint* checkpoint) {
```
karasikov, line 352:
```suggestion
                               BuildCheckpoint *checkpoint) {
```

metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
```diff
@@ -347,22 +348,30 @@ using Decoder = common::EliasFanoDecoder<T>;
...
     std::vector<Encoder<T_INT_REAL>> sinks;
     std::vector<std::string> names(chunk_count);
     for (size_t i = 0; i < names.size(); ++i) {
         names[i] = dir/("real_F_W_" + std::to_string(i));
+    }
+
+    if (checkpoint->checkpoint() > 2) {
+        logger->info("Skipping splitting k-mers into chunks");
+        return names;
+    }
```
karasikov, line 368: I got confused here. Please print that the k-mers have already been split into chunks in the last checkpoint.
Otherwise the logging message is misleading.

karasikov: also, add `assert(checkpoint->checkpoint() >= 2)` in the beginning of the function. It can't be 1, right? Similar thing for the other functions with next checkpoints.

metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
```diff
@@ -612,73 +676,125 @@ template <class KmerCollector, typename T_REAL, typename T>
...
                          ChunkedWaitQueue<T_REAL> &kmers,
                          ChunkedWaitQueue<T> *kmers_out,
-                         ThreadPool &async_worker) {
-    using KMER_REAL = get_first_type_t<T_REAL>; // 64/128/256-bit KmerBOSS
-    using T_INT_REAL = get_int_t<T_REAL>; // either KMER_INT or <KMER_INT, count>
-
-    using KMER = get_first_type_t<T>; // 64/128/256-bit KmerBOSS with sentinel $
-    using KMER_INT = typename KMER::WordType; // 64/128/256-bit integer
+                         ThreadPool &async_worker,
+                         BuildCheckpoint* checkpoint) {
```
karasikov, line 680:
```suggestion
                         BuildCheckpoint *checkpoint) {
```

metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
```diff
@@ -612,73 +676,125 @@ template <class KmerCollector, typename T_REAL, typename T>
...
+        checkpoint->set_kmer_dir(kmer_collector.tmp_dir());
+        checkpoint->set_checkpoint(1);
+    }
 
     size_t k = kmer_collector.get_k() - 1;
-    const std::filesystem::path dir = kmer_collector.tmp_dir();
+    const std::filesystem::path dir = checkpoint->kmer_dir();
     size_t num_threads = kmer_collector.num_threads();
 
+    if (previous_checkpoint == 1) {
```
karasikov, line 697: rename `previous_checkpoint` -> `last_checkpoint`?

metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
```diff
@@ -703,7 +819,7 @@ void recover_dummy_nodes(const KmerCollector &kmer_collector,
                     return kmer::transform<KMER>(reinterpret_cast<const KMER_REAL &>(v), k + 1) + kmer_delta;
                 }
             },
-            real_split_by_W, true
+            real_split_by_W, false /* remove sources */
```
karasikov, line 822: looks like a variable name commented out
```suggestion
            real_split_by_W, false // remove sources
```

metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
```diff
@@ -770,7 +886,8 @@ class BOSSChunkConstructor : public IBOSSChunkConstructor {
                          size_t num_threads,
                          double memory_preallocated,
                          const std::filesystem::path &tmp_dir,
-                         size_t max_disk_space)
+                         size_t max_disk_space,
+                         const BuildCheckpoint& checkpoint)
```
karasikov, line 890:
```suggestion
                         const BuildCheckpoint &checkpoint)
```

metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
```diff
@@ -779,26 +896,45 @@ class BOSSChunkConstructor : public IBOSSChunkConstructor {
...
+    template <typename KMER, typename T, typename Container>
+    BOSS::Chunk *build_chunk_2bit(Container &kmers) {
+        logger->trace("Reconstructing all required dummy source k-mers...");
+
+        Timer timer;
+        ChunkedWaitQueue<utils::replace_first_t<KMER, T>> queue(ENCODER_BUFFER_SIZE);
+        recover_dummy_nodes(kmer_collector_, kmers, &queue, async_worker_, &checkpoint_);
+        logger->trace("Dummy source k-mers were reconstructed in {} sec", timer.elapsed());
+        if (checkpoint_.phase() == 1) {
+            logger->info("Finished building phase 1");
```
karasikov, line 927:
```suggestion
            logger->info("Phase 1 finished");
```

karasikov (changes requested):

karasikov:
It's better to have non-deterministic tests than no tests.
The expected result is well determined. The second run of metagraph started after a previous run that was killed, must construct a valid graph.
So it's a perfectly defined test, isn't it?

danieldanciu:
Flakey/nondeterministic tests are a pest. There is general agreement about
this in the testing community, just google "flakey tests".
Your argument that the expected result is well determined doesn't mean that
the test is not flakey - all tests have a determined expected result. It's
not the expected result that makes a test flakey, but the actual result. A
test that sometimes passes and sometimes not is harmful. It erodes the
developers' trust in the testing system and often gets ignored with a
"well, it works for me" shrug. Flakey tests are also extremely hard to
debug, as they are not reproducible. There's hoards of literature teaching
developers how not to write non-deterministic tests, so writing a test that
is intentionally non-deterministic is really against the trend here.

If we were writing aviation software where people would die if a smooth
recovery was not always possible, then we'd figure out a way to run such
tests (e.g. introduce exit points at well-defined points in the code). But
in our case, recovering a failed computation is a best effort thing, so if
there is a hidden bug that causes the computation to be unrecoverable if
the crash happens exactly at that point, well, then so be it.

karasikov:
> A test that sometimes passes and sometimes not is harmful.

I agree that we should never write such tests. But I don't get what does it have to do with our case.
The second run of metagraph must always build the graph, so the test must always pass. This is what I mean by the deterministic expected result.

Can you explain why do you think this would be "Flakey"?

danieldanciu:
Flakey is a test that:
 - passes most of the time, but in relatively rare conditions it fails
 - the conditions under which it fails are unknown and cannot be (easily)
reproduced

Let's assume that there is the following completely plausible problem in my
current code: if the execution of the program is stopped right after the
chunks finished merging, but before the buffer of the reverse complements
was flushed out to disk, the checkpointing won't work, as the data is in an
inconsistent state: old chunks were deleted, and new reverse complements
are not yet flushed out. Depending on the size of the graph, the chances
that we stop the program right in this point can be one in a thousand or
one in a trillion. Let's say one in a thousand, since the test graphs are
small. So we'll have a test that passes 999 times out of 1000 tries. That's
a flakey test. Even if it fails, developers tend to ignore it, because it
passes most of the time, so they just assume that something was wrong on
the file system or something and move on. So the test brings no value.
There is little incentive to try to fix it, as the cost is very high (hard
to reproduce) and the benefit is very low (well, it passes most of the time
anyway). At Google there are entire teams of test engineers whose role,
among others, is to make sure that tests are reliable and either always
pass or they reproducibly fail.

In actual fact, I am almost convinced that although the current code will
work in most cases when the computation dies (because I inserted random
breakpoints in the code and tried it), there are still some rare
circumstances under which the data is inconsistent and the computation
can't be resumed.

karasikov:
I thought this can be used to resume crashed builds, so there is a guarantee that any run which finishes with exit code 0 builds a valid graph.

But you sound like this is not reliable enough to be used and one should start from scratch if a run fails.

So maybe we should use this to split the build into stages, but not use this to resume crashed runs?

karasikov:
Or maybe it was not meant to work in these scenarios (resume a crashed build) in the first place?

danieldanciu:
This discussion is getting surreal. That was just a thought experiment.
To the best of my knowledge, any build that finishes with exit code 0
builds a valid graph. As you could see in the code review, I took extra
care to make sure that we never leave the data in an inconsistent state or
that we never delete data too early. What I am saying is that there is a
very small chance that if the program crashes in a very specific way we
might to restart the build. And that is acceptable. Even if out of 1000
crashed builds, we can save 999, this code is pretty valuable.

To make an analogy, it already is the case that there is a small chance
that we give the program an input and it crashes while building the graph.
So `metagraph build` does suffer from some flaw that is on my list to
investigate. Out of maybe 50'000 graphs built, there is one graph where it
just crashes, and it's not because of insufficient memory. Should we just
delete metagraph and build graphs by hand because of this ? No, of course
not. The same is true for this code. I simply don't understand why we are
even having this discussion.

karasikov:
First of all, let me express my opinion about this

>Even if out of 1000 crashed builds, we can save 999, this code is pretty valuable.

This is not acceptable. If we allow commits with bugs in 1 out of 1000 runs, this will very soon accumulate and every second run will have a bug. Are we on the same page here?

> To the best of my knowledge, any build that finishes with exit code 0 builds a valid graph.

Up to this moment, this has been true, and we do check this in integration tests, by the way. However, if this PR introduces the feature for resuming a crashed run and starts with some pre-generated output, this is no longer obvious that the result will be a valid graph even if the exit code is zero. The test I'm proposing is supposed to check exactly this (exit code 0 => the graph is correct).

> To make an analogy, it already is the case that there is a small chance
that we give the program an input and it crashes while building the graph.
So `metagraph build` does suffer from some flaw that is on my list to
investigate. Out of maybe 50'000 graphs built, there is one graph where it
just crashes, and it's not because of insufficient memory. Should we just
delete metagraph and build graphs by hand because of this ? No, of course
not. The same is true for this code. I simply don't understand why we are
even having this discussion.

I wasn't aware of this. This might be a bug, which we have to find and fix.
I don't think this example can be used as an analogy with this case. There is nothing in common.

The thing I want to avoid is that we claim that a failed run can be resumed, then someone does this, and the second run of metagraph finishes with exit code 0, but the graph it's constructed is not correct (has missing k-mers, or invalid at all). So if we claim that the resuming is supported, we must be able to add the test. Otherwise you say that one can resume a failed run even though the result sometimes can be incorrect and there is no way to check it except rebuilding the graph again from scratch and comparing the two graphs. I hope it's clear that this is not a good thing.

karasikov:
Just to clarify, I'm proposing to add a test that checks the generated graph only if the last run finishes with exit code 0. If it fails too, the test would pass. So it's ok to have some stated from which we can't continue, but we must be able to detect this case, and the test would be supposed to check this.

danieldanciu (commented):
Is there anything else that needs to be addressed, or can we merge this?

metagraph/src/cli/build.cpp
```diff
@@ -115,15 +119,25 @@ int build_graph(Config *config) {
...
+            if (checkpoint.checkpoint() == 0) {
+                push_sequences(files, *config, timer, constructor.get());
+            } else {
+                logger->info("Skipping parsing sequences from input file(s)");
+            }
 
             boss::BOSS::Chunk *next_chunk = constructor->build_chunk();
+
+            if (checkpoint.phase() < 2) { // phase 1 stops after generating dummy k-mers
+                assert(next_chunk == nullptr);
```
hmusta, line 135: Maybe put a log message here to say why it's ending early?

metagraph/src/graph/representation/succinct/build_checkpoint.hpp
```diff
@@ -0,0 +1,42 @@
...
+#include <cstdint>
+#include <filesystem>
+#include <fstream>
+
+namespace mtg {
+namespace graph {
+namespace boss {
+
+/**
+ * Stores checkpointing and phase information for resuming disk-based building of succinct
```
hmusta, line 11: The code shows that there are >4 checkpoints, but the help message in config only mentions two possible values for phase. It would be useful to describe here how checkpoints and phases are related.

metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
```diff
@@ -388,25 +401,74 @@ void skip_same_suffix(const KMER &el, Decoder &decoder, size_t suf) {
...
+
+    std::vector<std::string> real_split_by_W(alphabet_size);
+    std::string dummy_sink_name = dir / "dummy_sink";
+    for (TAlphabet W = 0; W < alphabet_size; ++W) {
+        real_split_by_W[W] = dir/("real_split_by_W_" + std::to_string(W));
+    }
+
+    assert(checkpoint->checkpoint() >= 4);
+    if (checkpoint->checkpoint() > 4) {
+        return { real_split_by_W, dummy_sink_name };
```
hmusta, line 419: It would be useful to have a logger message written whenever a step is skipped due to a checkpoint being present, rather than lacking a message.

I see that you did this already in a few places, so maybe you missed this one...

hmusta (changes requested):

PR #199, opened 2020-09-15
various improvements in parallel traversal

karasikov:
- minor fixes and improvements
- made the critical section smaller: scales up to 48 and more threads
- reduced redundant traversal operations


hmusta:
I pushed a small change which wraps the unnecessary rotate between an `#if 0 ... #endif` block

hmusta (approved):

PR #200, opened 2020-09-16
Primary contig assembly: only lock when a conflict with another contig is detected

hmusta:


karasikov:
Great idea! I cleaned it up a bit

PR #201, opened 2020-09-16
only store a fixed number of decoded k-mers while traversing

hmusta:


metagraph/src/graph/representation/succinct/boss.cpp
```diff
@@ -1912,7 +1912,7 @@ void BOSS::call_paths(Call<std::vector<edge_index>&&,
     auto enqueue_start = [&](ThreadPool &thread_pool, edge_index start) {
         thread_pool.enqueue([&,start]() {
             ::mtg::graph::boss::call_paths(
-                    *this, { Edge(start, get_node_seq(start)) }, callback,
```
karasikov, line 1915: hm, but even empty std::string is quite large

hmusta: I tried this out with `std::variant` and it uses more RAM than the version in `dev` because `std::variant` has some overhead. To illustrate:

```c++
#include <iostream>
#include <vector>
#include <variant>
#include <utility>
#include <optional>
#include <cstdint>

int main() {
    std::cout << sizeof(std::pair<size_t, std::vector<uint8_t>>) << " "
              << sizeof(std::variant<size_t, std::pair<size_t, std::vector<uint8_t>>>) << " "
              << sizeof(std::pair<size_t, std::variant<std::vector<uint8_t>, std::nullopt_t>>) << "\n";
    return 0;
}
```
outputs `32 40 40`

hmusta: since `std::variant` stores uninitialized copies of each object it could be, along with an additional variable to act as an identifier

karasikov: Why not use std::variant of the entire vector?
`std::variant<std::vector<edge_index>, std::vector<std::pair<...>>>`

hmusta: With the current implementation, you can do a hybrid approach where you can store decoded k-mers, then switch to only storing only edge indices if you've used too much RAM. If the entire thing is an `std::variant`, then you have to do one or the other.

karasikov: Either way, we shouldn't replace vector with empty vector and hope that all works well, it wastes 24 bytes per element.

You can create two variables `std::vector<edge_index>, std::vector<std::pair<...>>` and push to `std::vector<edge_index>` if `std::vector<std::pair<...>>`, so the second vector is always limited, while the first can potentially grow without limits.

hmusta: One other thing, since the lengths of the `std::vector<TAlphabet>` are fixed and known, we could also store them as `std::unique_ptr<TAlphabet[]>` to save even more space.

karasikov (approved):

PR #203, opened 2020-09-18
get rid of bitmap fetched in graph traversal

karasikov:


metagraph/src/graph/representation/succinct/boss.cpp
```diff
@@ -2035,27 +2035,16 @@ void BOSS::call_paths(Call<std::vector<edge_index>&&,
...
-        }
-
-        for (edge_index edge : path) {
-            if (!fetch_and_set_bit(visited.data(), edge, async))
-                ++progress_bar;
+        // Ensures that call_path is called only once for each cycle
+        edge_index rep = *std::min_element(path.begin(), path.end());
+        if (!fetch_bit(visited.data(), rep, async)) {
+            // TODO: avoid get_node_seq(rep), and maybe pass the whole path
+            ::mtg::graph::boss::call_paths(
```
hmusta, line 2042: why `call_paths`? at this point, all non-cyclic parts of the graph should have been covered

metagraph/src/graph/representation/succinct/boss.cpp
```diff
@@ -2356,70 +2349,61 @@ call_path(const BOSS &boss,
...
     std::vector<size_t> breakpoints;
-    breakpoints.reserve(path.size());
+    breakpoints.reserve(dual_visited.size());
 
-    for (size_t i = 0; i < path.size(); ++i) {
-        if (fetched[path[i]]) {
-            assert(fetched[dual_path[i]]
-                    && "if a k-mer is fetched, its rev-compl must be too");
-            breakpoints.push_back(i);
+    // then lock all threads and resolve all compeeting nodes from dual_visited
```
hmusta, line 2378:
```suggestion
    // then lock all threads and resolve all competing nodes from dual_visited
```

hmusta (changes requested):

metagraph/src/graph/representation/succinct/boss.cpp
karasikov: To make it the processing the same as for other unitigs, so there are no special cases. What if not `call_caths`?

hmusta: it's a little bit overkill because of the extra incoming/outgoing checking

hmusta: `process_cycle` already does the same traversal as `call_paths` if you assume that the edge you're on is part of a cycle, the only difference is the order in which they set bits in `visited`. Before, we used to first set the bit at the representative, then set each bit in the path order.

To make it the same as with other unitigs, we'd have to first rotate the cycle so that it starts from the representative edge, then set bits in `visited` in order.

hmusta: This way, the order in which bits are set in `visited` will correspond to paths in the graph for cycles as well.

karasikov: But I start the traversal from the minimizer, so it is indeed the same as any other unitig/contig.

The first thing `call_paths` does is marks that minimizer as visited, so no cycle is called twice as a forward path, and all assumptions are satisfied.

I see, however, that this is overkill, especially taking into account that the entire cycle has already been traversed, so there is no need to do it again. But I just wanted to use a routine that already exists. I guess we need to write another simplified version of `call_paths` adjusted for this case: use the existing path without traversing the graph, but with checks for `visited`.

hmusta: or just iterate through in that order without rotating the path

hmusta: maybe it'll be enough to do something like
```c++
auto rep = std::min_element(path.begin(), path.end());
std::rotate(path.begin(), rep, path.end());
std::rotate(sequence.begin(), sequence.begin() + (rep - path.begin()), sequence.end());
auto begin = path.begin();
while (rep != path.end()) {
    if (fetch_and_set_bit(visited.data(), *rep, async)) {
        if (begin != rep) {
            // do stuff, then call_path
        }
        begin = rep;
    }
    ++rep;
}
```

metagraph/tests/graph/succinct/test_boss.cpp
```diff
@@ -1076,7 +1076,8 @@ TEST(BOSS, CallUnitigsTwoBigLoops) {
             num_kmers += path.size();
         }, num_threads, 0, true);
 
-        EXPECT_EQ(2, num_sequences);
+        // There is no guarantee on the number of contigs in the primary mode
+        // EXPECT_EQ(2, num_sequences);
         EXPECT_EQ(sequences[0].size() - k - 1 + sequences[1].size() - k - 1 - 1,
                   num_kmers);
```
hmusta, line 1082: there's something strange here now. If `num_sequences` is no longer 2, then the next test should either fail as well, or we're outputting duplicate k-mers

metagraph/src/graph/representation/succinct/boss.cpp
karasikov: it's not enough, we also need to start traversal for `dual_endpoints` and so on

metagraph/tests/graph/succinct/test_boss.cpp
karasikov: no, it's ok. `sequences` contains the input sequences, there are just two of them.

metagraph/src/graph/representation/succinct/boss.cpp
karasikov: I pushed a patch, but not sure this worth it. Maybe you want to benchmark...?

metagraph/tests/graph/succinct/test_boss.cpp
hmusta: ah, sorry, my bad. For some reason, I read this as if it were referring to the sequences extracted from the graph.

metagraph/src/graph/representation/succinct/boss.cpp
hmusta: One thing to note is that by the time we get to `process_cycles`, in both cases, all of the remaining k-mers are present in cycles. So when checking cycles in parallel, there are three cases to cover:

1. non-primary mode or none of the cycle's dual k-mers are in the graph:
mark the minimizer as visited, then mark the rest and call

2. all of the cycle's dual k-mers are in the graph:
mark the minimum of the path's and the dual path's minimizers as visited, then mark the rest in both strands and call only the forward path

3. some cycles with partial overlaps (may also overlap itself)
lock, rotate, then go through the path marking both the canonical and non-canonical forms of its k-mers. Use the result of fetch_and_set_bit on the canonical k-mer to decide where to split the cycle for output

I implemented this strategy in a76cfd288b8d7519cdc7d1360dc6633a245eaa53 but I haven't benchmarked it on any large data sets yet.

hmusta: I'll benchmark both approaches and let you know how it goes.

metagraph/tests/graph/succinct/test_boss.cpp
karasikov: yeah, at first I also thought so, I guess it won't hurt if we rename it

PR #205, opened 2020-09-21
start only one http server

karasikov:
The second server sometimes fails to immediately start on the same port afterr the first one, so here we start a single server and load the graph in a separate thread.


izcram:
I just did a small change making things maybe more readable. I also tried to address the inherent race condition we have in the test setup (https://github.com/ratschlab/projects2014-metagenome/blob/server/metagraph/integration_tests/test_api.py#L33) by reading from stdout of the server process. However, doing this correctly seemed relatively involved. I may revisit this decision when I have a minute again. Note, that by waiting we loose perhaps 2s in total, as this doesn't need to be done for every single api test.

PR #206, opened 2020-09-29
fast batch queries for row-diff annotation

karasikov:
* It never traverses the same nodes in the graph multiple times to reconstruct the full annotation from deltas.
* Also, queries all diff rows in one batch with `get_rows` on the base matrix.

Makes batch queries in row diff a few times faster.


metagraph/src/annotation/binary_matrix/row_diff/row_diff.cpp
```diff
@@ -42,6 +47,60 @@ BinaryMatrix::SetBitPositions RowDiff::get_row(Row row) const {
...
+RowDiff::get_rows(const std::vector<Row> &row_ids) const {
+    std::vector<SetBitPositions> rows(row_ids.size());
+
+    tsl::ordered_map<graph::boss::BOSS::edge_index, SetBitPositions> cached_rows;
+    cached_rows.reserve(rows.size() * 2);
+
+    for (size_t i = 0; i < row_ids.size(); ++i) {
+        // reset the cache if it's too large
+        if (cached_rows.size() > MIN_ROW_CACHE_CAPACITY
+                                    + row_ids.size() * ROW_CACHE_CAPACITY_FACTOR)
```
hmusta, line 60: why not use a circular buffer for this?

hmusta (changes requested):

metagraph/src/annotation/binary_matrix/row_diff/row_diff.cpp
```diff
@@ -42,6 +47,60 @@ BinaryMatrix::SetBitPositions RowDiff::get_row(Row row) const {
...
+            cached_rows.clear();
```
hmusta, line 61: it seems wasteful to clear the cache like this, could we not use one of the caches from the external library for this?

hmusta (changes requested):

metagraph/src/annotation/binary_matrix/row_diff/row_diff.cpp
karasikov: That would take more code and it's unlikely that it would work any faster.
if we assume the locality, like `row_ids[i + 1]` is faster to query if the results from `row_ids[i]` are cached, both approaches work equally well. If there is no locality, on the other hand, no caching strategy will work.

So if we pass `row_ids` in a nice order, it doesn't matter how large is the cache and how we reinitialize it.

karasikov: but maybe it makes sense to use a cache lib, to get the thread-safety easily, so we can run this loop with multiple threads. I'll just need to check if that doesn't make it slower with 1 thread.

karasikov: I figured it's better to assume the batch is small enough so we don't need to reset the buffer.

If a batch is too large and the indexes are shuffled, the cache will bring very little benefit.

So by never resetting the cache, we make sure that the entire batch is processed in the most efficient way.

metagraph/src/annotation/binary_matrix/row_diff/row_diff.hpp
```diff
@@ -148,6 +152,70 @@ BinaryMatrix::SetBitPositions RowDiff<BaseMatrix>::get_row(Row row) const {
     return result;
 }
 
+template <class BaseMatrix>
+std::vector<BinaryMatrix::SetBitPositions>
+RowDiff<BaseMatrix>::get_rows(const std::vector<Row> &row_ids) const {
+    // diff rows annotating nodes along the row-diff paths
+    std::vector<Row> rd_ids;
+    rd_ids.reserve(row_ids.size() * 2);
```
hmusta, line 160: Let's make this a compile-time constant for now.

metagraph/src/annotation/binary_matrix/row_diff/row_diff.hpp
```diff
@@ -148,6 +152,70 @@ BinaryMatrix::SetBitPositions RowDiff<BaseMatrix>::get_row(Row row) const {
...
+    // map row index to its index in |rd_rows|
+    VectorMap<Row, size_t> node_to_rd;
+    node_to_rd.reserve(row_ids.size() * 2);
```
hmusta, line 163: see above

metagraph/src/annotation/binary_matrix/row_diff/row_diff.hpp
```diff
@@ -148,6 +152,70 @@ BinaryMatrix::SetBitPositions RowDiff<BaseMatrix>::get_row(Row row) const {
...
+
+        graph::boss::BOSS::edge_index boss_edge = graph_->kmer_to_boss_index(
+                graph::AnnotatedSequenceGraph::anno_to_graph_index(row));
+
+        while (true) {
+            row = graph::AnnotatedSequenceGraph::graph_to_anno_index(
+                    graph_->boss_to_kmer_index(boss_edge));
+
+            auto [it, is_new] = node_to_rd.try_emplace(row, rd_ids.size());
+            rd_paths[i].push_back(it.value());
```
hmusta, line 181:
```suggestion
            rd_paths[i].push_back(*it);
```
you only need `it.value()` to access a non-const reference. for other cases, dereferencing should be good enough

metagraph/src/annotation/binary_matrix/row_diff/row_diff.hpp
```diff
@@ -148,6 +152,70 @@ BinaryMatrix::SetBitPositions RowDiff<BaseMatrix>::get_row(Row row) const {
...
+    // reconstruct annotation rows from row-diff
+    std::vector<SetBitPositions> rows(row_ids.size());
+
+    for (size_t i = 0; i < row_ids.size(); ++i) {
+        SetBitPositions &result = rows[i];
+        // propagate back to reconstruct full annotations for predecessors
+        for (auto it = rd_paths[i].rbegin(); it != rd_paths[i].rend(); ++it) {
+            merge(&result, rd_rows[*it]);
+            rd_rows[*it] = result;
+        }
```
hmusta, line 213: For the case where the values in `rows` correspond to the rowdiff path, this is kind of wasteful since the same merge operations get repeated.

hmusta (changes requested):

metagraph/src/annotation/binary_matrix/row_diff/row_diff.hpp
karasikov: You mean `2`?

I think this is a super minor thing. For large batches, this should be negligible. Maybe we should just remove these reserves?

hmusta: It's hard to predict a value for this since the graph topology is so variable. All we know really is that we can bound it above by the max row diff path length.

metagraph/src/annotation/binary_matrix/row_diff/row_diff.hpp
karasikov: These are truncated row-diff paths, see line 183, so nothing is merged multiple times.

metagraph/src/annotation/binary_matrix/row_diff/row_diff.hpp
karasikov: Yeah, but that could add an overhead of the worst case. I will just remove these reserves. They are insignificant anyway. Most of the complexity is in traversing and querying diff rows.

metagraph/src/annotation/binary_matrix/row_diff/row_diff.hpp
karasikov: Shouldn't it be `it->second`?
For `*it`, it is not quite clear what it returns, could return a pair.

metagraph/src/annotation/binary_matrix/row_diff/row_diff.hpp
karasikov: Never mind, why not to reserve, we know for sure it's not going to be smaller than `row_ids.size()`.

I'll add a global constant.

metagraph/src/annotation/binary_matrix/row_diff/row_diff.hpp
hmusta: yes, true. not that important anyway

hmusta (approved):

PR #207, opened 2020-09-29
properly align to primary graphs in query

karasikov:


metagraph/src/cli/query.cpp
```diff
@@ -461,7 +462,13 @@ int query_graph(Config *config) {
         assert(config->alignment_num_alternative_paths == 1u
                 && "only the best alignment is used in query");
 
-        aligner = build_aligner(*graph, *config);
+        if (!graph->is_canonical_mode() && config->canonical) {
+            // wrapp the primary graph into a canonical one
```
hmusta, line 466:
```suggestion
            // wrap the primary graph into a canonical one
```

hmusta:
One thing that's missing here is that the alignments to the graph will be in the CanonicalDBG index space, so after each alignment is done, we need to remap the nodes using `get_base_node`

hmusta:
ah, never mind, it ends up adding both orientations to the query graph, so they still get annotated

hmusta:
this looks fine, let's merge once the CI passes

PR #208, opened 2020-09-30
Memory efficient construction for row_diff and support for BRWT conversion.

danieldanciu:


metagraph/src/annotation/binary_matrix/row_diff/row_diff.hpp
```diff
@@ -9,92 +9,175 @@
...
+
+    const graph::DBGSuccinct *graph() const { return graph_; }
 
     /**
      * Returns the number of set bits in the matrix.
      */
-    uint64_t num_relations() const override { return num_relations_; }
-
-    uint64_t num_rows() const override { return terminal_.size(); }
+    uint64_t num_relations() const override { return diffs_.num_relations(); }
```
karasikov, line 61: move the implementation out of the class to the bottom of the file.
e.g.
```c++
template <class BaseMatrix>
inline bool RowDiff<BaseMatrix>::get(Row row, Column column) const {
```

metagraph/src/annotation/binary_matrix/row_diff/row_diff.hpp
```diff
@@ -9,92 +9,175 @@
...
+ * Retrieving data from RowDiff requires the associated #graph_. In order to get the
+ * annotation for  i-th row, we start traversing the node corresponding to i in #graph_
+ * and accumulate the values in #diffs until we hit a terminal node, which is stored in
+ * full.
  */
+//NOTE: Clang aggressively abuses the clause in the C++ standard (14.7.1/11) that allows
+// virtual methods in template classes to not be instantiated if unused and mistakenly
+// does not instantiate the virtual methods in this class, so I had to move definitions
+// to the header (gcc works fine)
+ template <class BaseMatrix>
```
karasikov, line 40:
```suggestion
template <class BaseMatrix>
```

metagraph/src/annotation/binary_matrix/row_diff/row_diff.hpp
```diff
@@ -9,92 +9,175 @@
...
+ *   3. #graph_ the graph that was used to determine adjacent rows for sparsification
+ * Retrieving data from RowDiff requires the associated #graph_. In order to get the
+ * annotation for  i-th row, we start traversing the node corresponding to i in #graph_
+ * and accumulate the values in #diffs until we hit a terminal node, which is stored in
+ * full.
  */
+//NOTE: Clang aggressively abuses the clause in the C++ standard (14.7.1/11) that allows
+// virtual methods in template classes to not be instantiated if unused and mistakenly
+// does not instantiate the virtual methods in this class, so I had to move definitions
+// to the header (gcc works fine)
```
karasikov, line 39: hm, but this is not the first class where we want to implement methods of a template class in .cpp file. How does it work in there?

metagraph/src/annotation/binary_matrix/row_diff/row_diff.hpp
```diff
@@ -9,92 +9,175 @@
...
-            : num_columns_(num_columns),
-              num_relations_(num_relations),
-              graph_(graph),
-              diffs_(diffs),
-              boundary_(boundary),
-              terminal_(terminal) {
-        sdsl::util::init_support(sboundary_, &boundary_);
+    RowDiff(const graph::DBGSuccinct *graph,
+               BaseMatrix &&diffs,
+               const std::string &terminal_file)
```
karasikov, line 47: terminal_file sounds ambiguous.

`terminal_file` -> `anchors_filename`

metagraph/src/annotation/binary_matrix/row_diff/row_diff.cpp
```diff
@@ -1,119 +1,31 @@
...
-    f.write(reinterpret_cast<const char *>(&num_columns_), 8);
-    f.write(reinterpret_cast<const char *>(&num_relations_), 8);
-    diffs_.serialize(f);
-    boundary_.serialize(f);
-    terminal_.serialize(f);
-}
-
-void RowDiff::serialize(const std::string &name) const {
+template <class BaseMatrix>
+void RowDiff<BaseMatrix>::serialize(const std::string &name) const {
```
karasikov, line 10: hm, so it does work. Why not move other methods to .cpp as well?

metagraph/src/annotation/binary_matrix/column_sparse/column_major.cpp
```diff
@@ -126,6 +132,12 @@ uint64_t ColumnMajor::num_relations() const {
     return num_set_bits;
 }
 
+void ColumnMajor::call_columns(const ColumnCallback &callback) {
+    for(uint64_t i = 0; i < data_.size(); ++i) {
+        callback(i, std::move(data_[i]));
+    }
+}
```
karasikov, line 139: Why not just return a const reference to `data_`?

metagraph/src/annotation/binary_matrix/column_sparse/column_major.cpp
```diff
@@ -109,8 +111,12 @@ bool ColumnMajor::load(std::istream &in) {
 void ColumnMajor::serialize(std::ostream &out) const {
     serialize_number(out, columns_->size());
 
-    for (const auto &column : *columns_) {
+    for (auto &column : *columns_) {
         assert(column.get());
+        // conversion pilfers the converted object, so we place the result back into
+        // #column, to ensure the ColumnMajor instance is valid after serialization
+        const_cast<std::unique_ptr<bit_vector> &>(column)
+                = std::make_unique<bit_vector_sd>(column->convert_to<bit_vector_sd>());
```
karasikov, line 119: This should serialize the given vectors and load them from their original representations (we need a loader `static std::unique_ptr<bit_vector> bit_vector::load(filename)`

Not necessary to implement it now, just add a TODO message.

metagraph/src/annotation/binary_matrix/row_diff/row_diff.cpp
```diff
@@ -1,119 +1,31 @@
...
     std::ofstream f(name, ios::binary);
     serialize(f);
     f.close();
 }
 
-bool RowDiff::load(const std::string &name) {
+template <class BaseMatrix>
+bool RowDiff<BaseMatrix>::load(const std::string &name) {
```
karasikov, line 17: name -> filename

metagraph/src/annotation/binary_matrix/row_diff/row_diff.hpp
```diff
@@ -9,92 +9,175 @@
...
-  private:
-    static void merge(Vector<uint64_t> *result, const Vector<uint64_t> &diff2);
+    Vector<uint64_t> get_diff(uint64_t node_id) const { return diffs_.get_row(node_id); }
 
-    uint64_t num_columns_ = 0;
+  private:
+    static void load_terminal(const std::string &filename, sdsl::rrr_vector<> *terminal) {
+        std::ifstream f(filename, ios::binary);
+        if (!f.good()) {
+            common::logger->error("Could not open terminal file {}", filename);
```
karasikov, line 138: it's not a file which is terminal
```suggestion
            common::logger->error("Could not open file  '{}' with anchors", filename);
```

metagraph/src/annotation/binary_matrix/row_diff/row_diff.hpp
```diff
@@ -9,92 +9,175 @@
...
+        terminal->load(f);
+        f.close();
+    }
 
-    uint64_t num_relations_ = 0;
+    static void merge(Vector<uint64_t> *result, const Vector<uint64_t> &diff2) {
+        if (diff2.empty()) {
+            return;
+        }
+        Vector<uint64_t> diff1;
```
karasikov, line 149: reserve space max(diff1.size(), diff2.size())

metagraph/src/annotation/binary_matrix/row_diff/row_diff.hpp
```diff
@@ -9,92 +9,175 @@
...
+        }
+        terminal->load(f);
+        f.close();
+    }
 
-    uint64_t num_relations_ = 0;
+    static void merge(Vector<uint64_t> *result, const Vector<uint64_t> &diff2) {
+        if (diff2.empty()) {
+            return;
+        }
```
karasikov, line 148: assert that both vectors are sorted
assert(std::is_sorted(...))

metagraph/src/annotation/binary_matrix/row_diff/row_diff.hpp
```diff
@@ -9,92 +9,175 @@
...
+        std::swap(*result, diff1);
+        uint64_t idx1 = 0;
+        uint64_t idx2 = 0;
+        while (idx1 < diff1.size() && idx2 < diff2.size()) {
+            if (diff1[idx1] == diff2[idx2]) {
+                idx1++;
+                idx2++;
+            } else if (diff1[idx1] < diff2[idx2]) {
+                result->push_back(diff1[idx1]);
+                idx1++;
```
karasikov, line 159:
```suggestion
                result->push_back(diff1[idx1++]);
```

metagraph/src/annotation/binary_matrix/row_diff/row_diff.hpp
```diff
@@ -9,92 +9,175 @@
...
+            } else {
+                result->push_back(diff2[idx2]);
+                idx2++;
```
karasikov, line 162:
```suggestion
                result->push_back(diff2[idx2++]);
```

metagraph/src/annotation/binary_matrix/row_diff/row_diff.hpp
```diff
@@ -9,92 +9,175 @@
...
+            }
+        }
+        while (idx1 < diff1.size()) {
+            result->push_back(diff1[idx1]);
+            idx1++;
```
karasikov, line 167:
```suggestion
            result->push_back(diff1[idx1+=]);
```

metagraph/src/annotation/binary_matrix/row_diff/row_diff.hpp
```diff
@@ -9,92 +9,175 @@
...
+        }
+        while (idx2 < diff2.size()) {
+            result->push_back(diff2[idx2]);
+            idx2++;
```
karasikov, line 171:
```suggestion
            result->push_back(diff2[idx2++]);
```

metagraph/src/annotation/annotation_converters.cpp
```diff
@@ -275,12 +304,28 @@ convert_to_greedy_BRWT<MultiBRWTAnnotator, std::string>(ColumnCompressed<std::st
...
+convert_to_greedy_BRWT(RowDiffAnnotator &&annotation,
+                       size_t num_parallel_nodes,
+                       size_t num_threads,
+                       uint64_t num_rows_subsampled) {
+    return convert_row_diff_to_BRWT(
+            std::move(annotation),
+            [num_threads,num_rows_subsampled](const auto &columns) {
+              std::vector<sdsl::bit_vector> subvectors
+                      = random_submatrix(columns, num_rows_subsampled, num_threads);
+              return greedy_matching(subvectors, num_threads);
```
karasikov, line 317:
```suggestion
                std::vector<sdsl::bit_vector> subvectors
                      = random_submatrix(columns, num_rows_subsampled, num_threads);
                return greedy_matching(subvectors, num_threads);
```

metagraph/src/annotation/annotation_converters.cpp
```diff
@@ -289,6 +334,17 @@ convert_to_simple_BRWT<MultiBRWTAnnotator, std::string>(ColumnCompressed<std::st
     );
 }
 
+std::unique_ptr<BRWTRowDiffAnnotator>
+convert_to_simple_BRWT(RowDiffAnnotator &&annotation,
+                                                      size_t grouping_arity,
+                                                      size_t num_parallel_nodes,
+                                                      size_t num_threads) {
```
karasikov, line 341: align

metagraph/src/annotation/annotation_converters.cpp
```diff
@@ -275,12 +304,28 @@ convert_to_greedy_BRWT<MultiBRWTAnnotator, std::string>(ColumnCompressed<std::st
...
+            num_threads
+    );
+}
+
 std::unique_ptr<MultiBRWTAnnotator>
-convert_to_simple_BRWT<MultiBRWTAnnotator, std::string>(ColumnCompressed<std::string>&& annotation,
-                                                        size_t grouping_arity,
-                                                        size_t num_parallel_nodes,
-                                                        size_t num_threads) {
+convert_to_simple_BRWT(ColumnCompressed<std::string> &&annotation,
```
karasikov, line 325:
```suggestion
convert_to_simple_BRWT(ColumnCompressed<std::string>&& annotation,
```

metagraph/src/annotation/annotation_converters.cpp
```diff
@@ -983,134 +1082,454 @@ void convert_to_row_annotator(const ColumnCompressed<std::string> &source,
...
+        if (!std::filesystem::exists(outfbase + suffix)) {
+            logger->trace("Building and writing  successor/predecessor/terminal files");
+            must_build = true;
+            break;
+        }
+    }
+    if (!must_build) {
+        logger->trace("Using existing pred/succ/terminal files in {}", outfbase);
+        return false;
+    }
```
karasikov, line 1100: move 1085:1511 to another file `src/annotation/row_diff_transform.cpp/hpp`.

karasikov (changes requested):

metagraph/src/cli/transform_annotation.cpp
```diff
@@ -434,26 +417,45 @@ int transform_annotation(Config *config) {
...
+                    "row_diff");
+            exit(1);
+        }
+        if(config->anno_type == Config::BRWTRowDiff) {
+            std::unique_ptr<BRWTRowDiffAnnotator> brwt_annotator;
+            if (config->infbase.empty()) { // load all columns in memory and compute linkage on the fly
+                logger->trace("Loading annotation from disk...");
+                auto annotator = std::make_unique<RowDiffAnnotator>();
+                if (!annotator->merge_load(files))
+                    std::exit(1);
```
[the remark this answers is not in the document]
danieldanciu: `merge_load` prints a more informative message, and say exactly which annotation file couldn't be loaded.

metagraph/tests/graph/succinct/test_boss.cpp
```diff
@@ -952,9 +952,10 @@ TEST(BOSS, CallSequencesRowDiff_EmptyGraph) {
             BOSS empty(k);
 
             sdsl::bit_vector terminal;
+            sdsl::bit_vector dummy;
             empty.call_sequences_row_diff([&](const auto &, const std::optional<uint64_t> &) {
                 FAIL() << "Empty graph should not have any sequences!";
-            }, num_threads, 1, &terminal);
+            }, num_threads, 1, &terminal, &dummy);
```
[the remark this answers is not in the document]
danieldanciu: That's one reason, the other reason is that mapping terminal indices from BOSS indices to annotation indices (lines 130-139 in the updated `row_dfif.cpp`) would be problematic when dummy k-mers are masked. I think you mentioned earlier that masking is sort of deprecated, so maybe that's not an issue.

metagraph/integration_tests/test_query.py
```diff
@@ -32,10 +34,17 @@
...
+    for graph in graph_types:
+        for anno in anno_types:
+            if graph == 'succinct' or (anno != 'row_diff' and anno != 'brwt_row_diff'):
+                result.append((graph, anno))
+    return result
 
 def build_annotation(graph_filename, input_fasta, anno_repr, output_filename, extra_params=''):
     target_anno = anno_repr
-    if anno_repr in {'rb_brwt', 'brwt'}:
+    if anno_repr in {'rb_brwt', 'brwt', 'row_diff', 'brwt_row_diff'}:
```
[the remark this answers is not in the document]
karasikov: I think it's better when the names of things from the same group share their prefix and not suffix.
For example, String1, String2, ... instead of FirstString, SecondString, etc.

The grouping here would be row_diff, which specifies what type of matrix is stored in the annotation.

metagraph/src/annotation/representation/annotation_matrix/annotation_matrix.hpp
```diff
@@ -39,6 +42,8 @@ class StaticBinRelAnnotator : public MultiLabelEncoded<Label> {
 
     const BinaryMatrixType& get_matrix() const override { return *matrix_; };
 
+    BinaryMatrixType&& release_matrix()  { return std::move(*matrix_); };
```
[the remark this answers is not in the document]
karasikov: Not sure I understand what you mean.

With
```c++
BinaryMatrixType* release_matrix()  { return matrix_.release(); };
```
you would do `std::unique_ptr<BinaryMatrixType> new_matrix(annotator.release_matrix());`

And you don't need to free anything.

How is this less readable?

metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
```diff
@@ -542,9 +542,18 @@ const bitmap& ColumnCompressed<Label>::get_column(const Label &label) const {
 template <typename Label>
 const binmat::ColumnMajor& ColumnCompressed<Label>::get_matrix() const {
     flush();
+    annotation_matrix_view_.point_to(bitmatrix_);
```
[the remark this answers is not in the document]
karasikov: The internal state of `annotation_matrix_view_` is changed, so this is indeed a re-initialization.

But it's a minor point. I'm ok either way.

If you don't want `initialize`, maybe rename it to `change_pointer` or `update_pointer`, so it's clear that it changes the internal state and not supposed to return anything, as I would assume from the name `point_to()`

metagraph/src/cli/transform_annotation.cpp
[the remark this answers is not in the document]
karasikov: Not all annotators do it. I saw `annotator`, which doesn't specify the type.
Some annotators may not print error messages inside `load()`, they just return `false`.

So if we rely on certain behavior in this annotator, would be good to call it accordingly, e.g., `row_diff_anno` instead of `annotator`.

But anyway, adding an extra error message won't hurt (but it's ok not to add if you indeed rely on logging inside the method)

metagraph/src/annotation/binary_matrix/row_diff/row_diff.hpp
karasikov: It's good to be able to see the pure interface. Currently, it's crowded with the implementations which make it hard to view the list of all methods in the class. This won't change the performance, just make .hpp more structured.

metagraph/src/annotation/binary_matrix/row_diff/row_diff.cpp
karasikov: yeah, I see the problem. Though `override` is optional so maybe we could do without specifying it. The code will be easier to read and maybe it will compile faster too. I think it's worth neglecting `override` in this case, and I'm not a huge fan of it anyway.

metagraph/src/annotation/binary_matrix/column_sparse/column_major.cpp
karasikov: ah, I missed that it moves them. In this case, return the object itself `std::vector<...>`.

The implementation is simpler. Also, maybe in some cases, the target thing you need is this vector. So if we implement it with a callback, you'd have to put all these bit vectors back to an std::vector in the callback, which is a bit stupid because they were initially in an std::vector.

So the idea here is to reduce the total number of conversions which may happen.
If we return the entire `data_`, we will never do extra overhead for sure.
Essentially, this loop `for ...` will be moved to `annotation_converters.cpp`, where you need to construct a callback.

metagraph/src/annotation/binary_matrix/column_sparse/column_major.cpp
karasikov: It not cool to convert a given representation to `bit_vector_sd`.
It should dump its original representation, which could be bit_vector_stat, bit_vector_rrr, etc.

So the problem here is not in serialization but in loading (we can't load a vector if we don't know its representation).
But this should be solved in the future when we can load a bit vector without necessarily knowing its representation: `bit_vector::load(filename)`. So I asked to add a TODO message to fix this serialization once this loader is implemented.

Why -- to avoid blind transforms to `bit_vector_sd`, which may be inappropriate for certain cases.

metagraph/src/annotation/binary_matrix/row_diff/row_diff.hpp
karasikov: It does work. If you reserve space for `diff1`, it will be swapped with `result`, into which you push elements below.
So `result` will have a reserved array and will never reallocate.

karasikov (changes requested):

metagraph/src/annotation/binary_matrix/row_diff/row_diff.hpp
karasikov: you can also call reserve for `result` after swap. That is equivalent.

metagraph/src/annotation/representation/annotation_matrix/annotation_matrix.hpp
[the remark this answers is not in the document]
danieldanciu: Maybe I should say "more verbose" rather than "less readable". Here is what the caller currently looks like:
```  
std::vector<std::unique_ptr<bit_vector>> columns 
    = annotator.release_matrix().diffs().release_columns();
```

If I made this change, the caller would have to construct a unique_ptr:

```  
std::vector<std::unique_ptr<bit_vector>> columns
            = std::unique_ptr<binmat::RowDiff<binmat::ColumnMajor>>(annotator.release_matrix())
                      .diffs()
                      .release_columns();
```

metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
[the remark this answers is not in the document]
danieldanciu: `update_pointer` sounds good

metagraph/src/cli/transform_annotation.cpp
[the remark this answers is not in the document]
danieldanciu: Renamed to `row_diff_anno`

metagraph/src/annotation/annotation_converters.cpp
```diff
@@ -1263,7 +1261,7 @@ void convert_batch_to_row_diff(const graph::DBGSuccinct &graph,
         set_rows[i].resize(sources[i]->num_labels());
         for(uint64_t j = 0; j < sources[i]->num_labels(); ++j) {
             const std::filesystem::path tmp_dir = tmp_path/std::to_string(i / 100)
-                    /("col_" + std::to_string(i) + "_" + std::to_string(j));
+                    /fmt::format("col_{}_{}", i, j);
```
karasikov: put i/100 inside fmt as well?

metagraph/src/annotation/binary_matrix/column_sparse/column_major.cpp
```diff
@@ -132,10 +134,8 @@ uint64_t ColumnMajor::num_relations() const {
     return num_set_bits;
 }
 
-void ColumnMajor::call_columns(const ColumnCallback &callback) {
-    for(uint64_t i = 0; i < data_.size(); ++i) {
-        callback(i, std::move(data_[i]));
-    }
+std::vector<std::unique_ptr<bit_vector>>&& ColumnMajor::release_columns() {
+    return std::move(data_);
 }
```
karasikov: move to .hpp?

metagraph/integration_tests/test_query.py
```diff
@@ -55,23 +64,38 @@ def build_annotation(graph_filename, input_fasta, anno_repr, output_filename, ex
...
+
 
 
 @parameterized_class(('graph_repr', 'anno_repr'),
-    input_values=list(itertools.product(
+    input_values=product(
         GRAPH_TYPES + ['succinct_bloom', 'succinct_mask'],
         ANNO_TYPES
-    )),
+    ),
```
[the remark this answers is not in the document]
danieldanciu: Ah, I had to revert this fix, as simply adding the pair here is not enough. It needs to be added to the global `anno_file_extension` map, as `build_annotation` is using that map in order to build the annotation file name.

metagraph/src/annotation/representation/annotation_matrix/annotation_matrix.hpp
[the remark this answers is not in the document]
karasikov: Hm, I see. Then maybe return `std::unique_ptr<BinaryMatrixType>`, so you can still write
```c++
std::vector<std::unique_ptr<bit_vector>> columns 
    = annotator.release_matrix()->diffs().release_columns();
```

metagraph/integration_tests/test_query.py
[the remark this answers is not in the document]
karasikov: yeah, I would put the row-diff/no-row-diff on the higher place because that defined the content of the matrix. The representation is conceptually less important, it just defines compression.

Like `bit_vector_sd`, and not `sd_bit_vector`.
Don't ask me about `SuccinctDBG`... :)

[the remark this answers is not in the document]
karasikov: Ah, actually it's `DBGSuccinct`, so all food. The representation/compression is in the suffix.

metagraph/src/graph/representation/succinct/boss.cpp
```diff
@@ -2671,25 +2671,29 @@ void BOSS::call_sequences_row_diff(
...
+            }
         });
     };
 
     // start traversal from the dummy source edges first ($..$X)
     //  .____
     for (edge_index i = succ_last(1); i >= 1; --i) {
         if (!fetch_bit(visited.data(), i, async))
-            enqueue_start(thread_pool, i);
+            enqueue_start(thread_pool, {i});
```
karasikov, line 2687:
```suggestion
            enqueue_start(thread_pool, { i });
```

metagraph/src/graph/representation/succinct/boss.cpp
```diff
@@ -2671,25 +2671,29 @@ void BOSS::call_sequences_row_diff(
     ThreadPool thread_pool(std::max(num_threads, 1UL), TASK_POOL_SIZE);
     constexpr bool async = true;
 
-    // TODO: Accumulate more edges and start each task for a bunch of nodes
-    auto enqueue_start = [&](ThreadPool &thread_pool, edge_index start) {
+    auto enqueue_start = [&](ThreadPool &thread_pool, const std::vector<edge_index> &start) {
         thread_pool.enqueue([&, start]() {
-            call_row_diff_path(*this, start, callback, max_length, &visited, terminal,
-                               &near_terminal, dummy, progress_bar);
+            for (const auto& edge : start) {
```
karasikov, line 2676:
```suggestion
            for (edge_index edge : start) {
```

metagraph/src/annotation/annotation_converters.cpp
```diff
@@ -983,134 +1080,339 @@ void convert_to_row_annotator(const ColumnCompressed<std::string> &source,
...
+                   const std::string &graph_fname,
+                   const std::vector<std::unique_ptr<annot::ColumnCompressed<>>> &sources,
+                   const std::function<void()> &before_chunk,
+                   const CallOnes &call_ones,
+                   const std::function<void(uint64_t start, uint64_t size)> &after_chunk) {
+    const uint32_t num_threads = get_num_threads();
+    const uint64_t num_rows = graph.num_nodes();
+
+    sdsl::int_vector_buffer succ(graph_fname + ".succ", std::ios::in, 1024*1024);
+    sdsl::int_vector_buffer pred(graph_fname + ".pred", std::ios::in, 1024*1024);
```
karasikov, line 1105: Please move this to another file like row_diff_builder.cpp, this annotation_converters.cpp is too long.

metagraph/src/annotation/annotation_converters.cpp
```diff
@@ -983,134 +1080,339 @@ void convert_to_row_annotator(const ColumnCompressed<std::string> &source,
...
+                                    uint64_t start_idx,
+                                    uint64_t source_idx,
+                                    uint64_t col_idx,
+                                    const std::vector<uint64_t> &succ_chunk,
+                                    const std::vector<uint64_t> &pred_chunk,
+                                    const std::vector<uint64_t> &pred_chunk_idx)>;
+
+constexpr uint64_t chunk_size = 1 << 20;
+
+void traverse_anno_chunked(const std::string &name,
```
karasikov, line 1094: what is `name`? `output_filename`?

metagraph/src/annotation/annotation_converters.cpp
```diff
@@ -983,134 +1080,339 @@ void convert_to_row_annotator(const ColumnCompressed<std::string> &source,
...
+                   const graph::DBGSuccinct &graph,
+                   const std::string &graph_fname,
```
karasikov, line 1096: why do you need to pass both graph and its filename?

metagraph/src/annotation/annotation_converters.cpp
```diff
@@ -983,134 +1080,339 @@ void convert_to_row_annotator(const ColumnCompressed<std::string> &source,
...
+                   const std::vector<std::unique_ptr<annot::ColumnCompressed<>>> &sources,
```
karasikov, line 1097: `column_annotators`?

metagraph/src/annotation/binary_matrix/row_diff/row_diff.cpp
```diff
@@ -1,118 +1,154 @@
...
+
+    // traverse BOSS table in parallel processing num_threads chunks of size 1'000'000
+    constexpr uint64_t chunk_size = 1'000'000;
+    const uint64_t batch_size = chunk_size * num_threads;
+    const uint64_t batch_count = (graph.num_nodes() - 1) / batch_size + 1;
+
+    ProgressBar progress_bar(batch_count, "Compute successors", std::cerr,
+                             !common::get_verbose());
+
+    for(uint64_t batch = 0; batch < batch_count; ++batch) {
```
karasikov, line 92:
```suggestion
    for (uint64_t batch = 0; batch < batch_count; ++batch) {
```

metagraph/src/annotation/binary_matrix/row_diff/row_diff.cpp
```diff
@@ -1,118 +1,154 @@
...
+                    outfbase);
+            must_build = true;
+            break;
         }
     }
-    while (idx1 < diff1.size()) {
-        result->push_back(diff1[idx1]);
-        idx1++;
+    if (!must_build) {
+        common::logger->trace("Using existing pred/succ/terminal files in {}.*", outfbase);
```
karasikov, line 49:
```suggestion
        common::logger->trace("Using existing pred / succ / terminal files in {}.*", outfbase);
```
otherwise looks like a path

metagraph/src/annotation/binary_matrix/row_diff/row_diff.cpp
```diff
@@ -1,118 +1,154 @@
...
+    ProgressBar progress_bar(batch_count, "Compute successors", std::cerr,
+                             !common::get_verbose());
+
+    for(uint64_t batch = 0; batch < batch_count; ++batch) {
+        std::vector<std::vector<uint64_t>> pred_buf(num_threads);
+        std::vector<std::vector<uint64_t>> succ_buf(num_threads);
+        std::vector<std::vector<bool>> pred_boundary_buf(num_threads);
+
+        #pragma omp parallel for num_threads(num_threads) schedule(static, 1)
+        for (uint32_t chunk = 0; chunk < std::max((uint32_t)1, num_threads); ++ chunk) {
```
karasikov, line 98:
```suggestion
        for (uint32_t chunk = 0; chunk < std::max((uint32_t)1, num_threads); ++chunk) {
```

metagraph/src/annotation/binary_matrix/row_diff/row_diff.cpp
```diff
@@ -1,118 +1,154 @@
...
+        }
+        ++progress_bar;
     }
+    succ.close();
+    pred.close();
+
+    std::ofstream fpred_boundary(outfbase + ".pred_boundary");
+    sdsl::rrr_vector rpred_boundary(to_sdsl(pred_boundary));
+    rpred_boundary.serialize(fpred_boundary);
+    fpred_boundary.close();
```
karasikov, line 148: This can be quite large, I think, why not use delimiters in `pred`? It will be about twice larger, but we read it with streaming which may be ok. What do you think?

metagraph/src/annotation/annotation_converters.cpp
```diff
@@ -983,134 +1080,339 @@ void convert_to_row_annotator(const ColumnCompressed<std::string> &source,
...
+                },
+                [&](uint64_t chunk_start, uint64_t chunk_size) {
+                    for (uint64_t i = 0; i < chunk_size; ++i) {
+                        if (sparse_ones[i] > 2 && sparse_ones[i] / 2 > orig_ones[i]) {
+                            new_terminal[i + chunk_start] = 1;
+                        } else {
+                            new_terminal[i + chunk_start] = terminal[i + chunk_start];
+                        }
+                    }
+                });
```
karasikov, line 1235: use node_index instead of uint64_t where possible, so it's easier to read so see what refers to a node index and what is a simple chunk/file index

metagraph/src/annotation/annotation_converters.cpp
```diff
@@ -983,134 +1080,339 @@ void convert_to_row_annotator(const ColumnCompressed<std::string> &source,
...
+                     p_idx < pred_chunk_idx[chunk_idx + 1]; ++p_idx) {
+                    if (!source_col[pred_chunk[p_idx]] && !terminal[pred_chunk[p_idx]])
+                        set_rows[source_idx][col_idx].push_back(pred_chunk[p_idx]);
                 }
-                pos[anno_ids[i]] = { start_idx, diffs.size() - start_idx };
-            }
+            },
+            [&](uint64_t /* start */, uint64_t /* chunk_size */) {
+                for (uint32_t source_idx = 0; source_idx < sources.size(); ++source_idx) {
+                    for (uint32_t col_idx = 0; col_idx < set_rows[source_idx].size(); ++col_idx) {
```
karasikov, line 1303: use size_t to index stl containers.

karasikov (approved):

metagraph/src/annotation/representation/annotation_matrix/annotation_matrix.hpp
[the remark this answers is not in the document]
karasikov: It requires move constructors to be implemented.

For you example, you don't have to call the second release(), you just need to dereference and move
```c++
return std::make_unique<RowDiffBRWTAnnotator>(
            std::make_unique<RowDiff<BRWT>>(nullptr, std::move(*annotator->release_matrix()), ""),
            annotator->get_label_encoder());
```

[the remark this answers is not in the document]
karasikov: also, if you need a pointer, this
```
std::make_unique<BRWT>(std::move(annotator->release_matrix()));
```
becomes simply this
```
annotator->release_matrix();
```

metagraph/src/annotation/annotation_converters.cpp
karasikov: ah, then find a more appropriate name, like `logging_header`. I was sure this is a filename but had no idea which one.

metagraph/src/annotation/annotation_converters.cpp
karasikov: Could also be `anchors_filename`, or something. I wanted to point out that this doesn't have to be in the same directory as the graph. They could be stored in different locations. So the name should reflect that. `outfbase` is fine too

metagraph/src/annotation/binary_matrix/row_diff/row_diff.cpp
karasikov: But I think at some point you convert it to rrr, which requires the full bit vector to be stored in RAM. This is 1 bit per k-mer -- unnecessary overhead.

metagraph/src/annotation/annotation_converters.cpp
karasikov: `column_annotations` is ok

metagraph/src/annotation/representation/annotation_matrix/annotation_matrix.hpp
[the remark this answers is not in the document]
karasikov: It is generally easier to get a reference from a pointer than the other way around. So if we already have a pointer, it's good to keep it as a pointer without reinitializing the object with move constructors.
Then, if a reference is needed, the user will be responsible to convert it.

metagraph/src/annotation/annotation_converters.cpp
karasikov: uint32_t can be 4 billion.

In fact, we use `size_t` exactly for this purpose -- to emphasize that the max value is quite small so we don't have to use uint64_t/uint32_t.

metagraph/src/annotation/binary_matrix/row_diff/row_diff.cpp
danieldanciu: I had removed that conversion. Now it's just a plain old bit set.

metagraph/src/annotation/annotation_converters.cpp
danieldanciu: I know how big uint32_t can be. It's quite unusual to signal that a value is small by using largest possible type...

karasikov: size_t is not the largest possible type

danieldanciu: From cppreference.com:
> size_t can store the maximum size of a theoretically possible object of any type (including array).

So we are using the maximum possible size of an array to indicate that the array is small.  :)

Yes, if we start splitting hairs, we can say that uint128_t is larger, but that can never be the size of an array and it's not supported on many platforms (e.g. ARM), so I didn't even consider it.

Anyway, I really don't care one way or the other, I just found the convention amusing.

karasikov: You are considering only arrays, but this doesn't have to refer to an array. It could be a more general class.

size_t can be smaller than 64 bits and it can even be smaller than 32.

It must be at least 16 bits, so theoretically when you write `size_t`, you implicitly show that the thing you index doesn't have more than 2^16 elements, or that you index something that is supposed to be indexed by size_t, like any stl container.

PR #209, opened 2020-10-02
Adapt server for primary/canonical graphs

karasikov:
* Always construct canonical query graphs
* Test with both canonical and primary

### DIFF
--- a/metagraph/integration_tests/test_api.py
+++ b/metagraph/integration_tests/test_api.py
@@ -39,7 +39,7 @@ class TestAPIBase(TestingBase):
         cls.server_process.kill()
 
     def _start_server(self, graph, annotation):
-        construct_command = '{exe} server_query -i {graph} -a {annot} --port {port} --address {host} -p {threads}'.format(
+        construct_command = '{exe} server_query -i {graph} -a {annot} --port {port} --address {host} -p {threads} --primary-kmers'.format(
             exe=METAGRAPH,
             graph=graph,
             annot=annotation,
--- a/metagraph/integration_tests/test_api.py
+++ b/metagraph/integration_tests/test_api.py
@@ -7,21 +7,23 @@ from subprocess import Popen
 import socket
 import requests
 from metagraph.client import GraphClientJson, MultiGraphClient
-from parameterized import parameterized
+from parameterized import parameterized, parameterized_class
 
 from base import TestingBase, METAGRAPH, TEST_DATA_DIR
 
 
 class TestAPIBase(TestingBase):
     @classmethod
-    def setUpClass(cls, fasta_path):
+    def setUpClass(cls, fasta_path, canonical=False, primary=False):
         super().setUpClass()
 
         graph_path = cls.tempdir.name + '/graph.dbg'
         annotation_path = cls.tempdir.name + '/annotation.column.annodbg'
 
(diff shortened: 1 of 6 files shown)


metagraph/integration_tests/test_api.py
```diff
@@ -183,15 +187,17 @@ def test_api_raw_search_empty_fasta_desc(self):
         self.assertEqual(ret[0]['seq_description'], '')
 
 
+@parameterized_class(('mode',), input_values=[('canonical',), ('primary',)])
 class TestAPIClient(TestAPIBase):
     graph_name = 'test_graph'
 
     sample_query = 'CCTCTGTGGAATCCAATCTGTCTTCCATCCTGCGTGGCCGAGGG'
-    sample_query_expected_cols = 98
+    sample_query_expected_cols = 99
```
hmusta, line 195: is this supposed to be the expected number of columns which have a match to the query? It would be useful to either name this `sample_query_expected_match_cols` or write a comment describing this. The value is so close to the total number of columns that this might be suspected as a bug.

karasikov: Makes sense, I was a bit confused at first as well. Feel free to change.

karasikov: btw, it changed from 98 to 99 because of the switch from the basic to canonical/primary graphs.

metagraph/integration_tests/test_api.py
```diff
@@ -39,7 +39,7 @@ def tearDownClass(cls):
         cls.server_process.kill()
 
     def _start_server(self, graph, annotation):
-        construct_command = '{exe} server_query -i {graph} -a {annot} --port {port} --address {host} -p {threads}'.format(
+        construct_command = '{exe} server_query -i {graph} -a {annot} --port {port} --address {host} -p {threads} --primary-kmers'.format(
```
karasikov, line 42: why do we need this flag?

metagraph/src/cli/query.cpp
```diff
@@ -463,8 +463,11 @@ int query_graph(Config *config) {
                 && "only the best alignment is used in query");
 
         if (!graph->is_canonical_mode() && config->canonical) {
-            // wrap the primary graph into a canonical one
-            graph.reset(new CanonicalDBG(graph, true));
+            // If the input graph is non-canonical and non-primary, and if there
+            // exist a pair of forward and reverse complement k-mers with different
+            // annotations, then we need to have a way of setting the second argument
+            // to false.
+            graph.reset(new CanonicalDBG(graph, config->kmers_in_single_form));
```
karasikov, line 470: kmers_in_single_form  = `false` should generalize the mode with kmers_in_single_form = true. Why is this essential to specify this?

metagraph/integration_tests/test_api.py
hmusta: When in primary mode, setting the `--primary-kmers` flag makes the querying faster since it enables some optimizations in CanonicalDBG. If the input graph is canonical, then this flag is ignored.

karasikov: It's easy to make mistake if we always have to pass this. So let's turn this on by default for now and then add an attribute to primary graphs, so this can be inferred automatically.

PR #212, opened 2020-10-06
Chain local alignments

hmusta:
This changes the default aligner behaviour so that instead of returning the best local alignment, the best-scoring disjoint set of alignments is returned.


metagraph/integration_tests/test_api.py
```diff
@@ -186,6 +186,50 @@ def test_api_raw_search_empty_fasta_desc(self):
...
+            graph=graph,
+            annot=annotation,
+            host=self.host,
+            port=self.port,
+            threads=2
+        )
+
+        return Popen(shlex.split(construct_command))
+
+    @parameterized.expand([(1,1), (3,1)])
```
karasikov, line 212:
```suggestion
    @parameterized.expand([(1, 1), (3, 1)])
```

metagraph/src/cli/align.cpp
```diff
@@ -330,9 +332,14 @@ int align_to_graph(Config *config) {
                         *out << "\t*\t*\t" << config->alignment_min_path_score
                              << "\t*\t*\t*";
                     } else {
+                        size_t sum = 0;
                         for (const auto &path : paths) {
+                            sum += path.get_score();
```
karasikov, line 337: better call it `total_score` or `score_sum`

metagraph/src/cli/query.cpp
```diff
@@ -800,7 +800,6 @@ construct_query_graph(const AnnotatedDBG &anno_graph,
     return std::make_unique<AnnotatedDBG>(graph, std::move(annotation));
 }
 
-
 int query_graph(Config *config) {
```
karasikov, line 803:
```suggestion

int query_graph(Config *config) {
```

karasikov (changes requested):

metagraph/src/graph/alignment/aligner_helper.hpp
```diff
@@ -855,25 +855,29 @@ class DPTable {
...
+                gap_count.reserve(new_size);
+
+                scores.insert(scores.begin(), shift, min_score_);
+                gap_scores.insert(gap_scores.begin(), shift, min_score_);
+                ops.insert(ops.begin(), shift, Cigar::CLIPPED);
+                prev_nodes.insert(prev_nodes.begin(), shift, 0);
+                gap_prev_nodes.insert(gap_prev_nodes.begin(), shift, 0);
+                gap_count.insert(gap_count.begin(), shift, 0);
+
+                scores.resize(new_size);
```
karasikov: you need to do `resize(new_size, min_score_);` if you don't want the new values to be uninitialized.

hmusta: the default allocator would initialize these to 0 (which is a mistake since it could be that `min_score_` is non-zero), but since these vectors are using the AlignedAllocator, I'm not sure if it does this by default

karasikov: so, this needs to be fixed, right?

hmusta: yes, I already fixed it

metagraph/src/graph/alignment/aligner_aggregator.hpp
```diff
@@ -37,28 +39,88 @@ class AlignmentAggregator {
...
+    if (!config_.chain_alignments) {
+        // no alignment chaining
+        if (path_queue_.size() < config_.num_alternative_paths) {
+            path_queue_.emplace(std::move(alignment));
+        } else if (alignment.get_score() > path_queue_.minimum().get_score()) {
+            path_queue_.update(path_queue_.begin(), std::move(alignment));
+        }
+    } else {
+        // alignment chaining
+        bool updated = false;
```
karasikov, line 71: to reduce the number of 'else' blocks...
```suggestion
    if (!config_.chain_alignments) {
        // no alignment chaining
        if (path_queue_.size() < config_.num_alternative_paths) {
            path_queue_.emplace(std::move(alignment));
        } else if (alignment.get_score() > path_queue_.minimum().get_score()) {
            path_queue_.update(path_queue_.begin(), std::move(alignment));
        }
        return;
    }

    // alignment chaining
    bool updated = false;
```

metagraph/src/graph/alignment/aligner_aggregator.hpp
```diff
@@ -67,12 +129,187 @@ ::call_alignments(const std::function<void(DBGAlignment&&)> &callback) {
...
+        }
+
+        return;
+    }
+
+    std::sort(cur_paths.begin(), cur_paths.end(),
+              [&](const auto &a, const auto &b) {
+                  return a.get_query_end() < b.get_query_end()
+                      || (a.get_query_end() == b.get_query_end() && a.get_query().data() < b.get_query().data());
+              });
```
karasikov, line 197:
```suggestion
              [&](const auto &a, const auto &b) {
                  return std::make_pair(a.get_query_end(), a.get_query().data())
                          < std::make_pair(b.get_query_end(), b.get_query().data());
              });
```

metagraph/src/graph/alignment/aligner_aggregator.hpp
```diff
@@ -67,12 +129,187 @@ ::call_alignments(const std::function<void(DBGAlignment&&)> &callback) {
...
+
+    std::pair<score_t, std::vector<DBGAlignment>> cur_chain(config_.min_path_score, {});
+    call_alignment_chains(cur_paths, [&](std::vector<DBGAlignment>&& chain,
+                                         score_t score) {
```
karasikov, line 201:
```suggestion
    call_alignment_chains(cur_paths, [&](std::vector<DBGAlignment>&& chain, score_t score) {
```

metagraph/src/graph/alignment/aligner_aggregator.hpp
```diff
@@ -67,12 +129,187 @@ ::call_alignments(const std::function<void(DBGAlignment&&)> &callback) {
...
+        assert(!i);
+        for (auto it = paths.begin(); it != paths.end(); ++it) {
+            if (cur_score + it->get_score() > best_score->at(it - paths.begin())) {
+                best_score->at(it - paths.begin()) = cur_score + it->get_score();
+                call_alignment_chains(paths, callback, it - paths.begin() + 1,
+                                      { *it }, cur_score + it->get_score(), best_score);
+            }
+        }
+
+        return;
```
karasikov, line 249:
```suggestion
        return;
```

metagraph/src/graph/alignment/aligner_helper.cpp
```diff
@@ -382,6 +382,460 @@ void Alignment<NodeType>::append(Alignment&& other) {
...
+    assert(cigar_rbegin_ <= cigar_it_);
+    assert(cigar_it_ <= cigar_rend_);
+
+    return *this;
+}
+
+template <typename NodeType>
+AlignmentPrefix<NodeType>& AlignmentPrefix<NodeType>::operator--() {
+    throw std::runtime_error("not implemented");
+    return *this;
```
karasikov, line 470: If this can't be implemented, it's better to remove this method.

Or, if this can be supported in the future, write "not implemented yet"

metagraph/src/graph/alignment/aligner_helper.hpp
```diff
@@ -48,6 +50,9 @@ class Cigar {
     // e.g., 3=1X2I3D for 3 matches, 1 mismatch, 2 insertions, 3 deletions
     explicit Cigar(const std::string &cigar_str);
 
+    explicit Cigar(const_iterator begin, const_iterator end)
+          : cigar_(begin, end) { }
```
karasikov, line 54:
```suggestion
    explicit Cigar(const_iterator begin, const_iterator end) : cigar_(begin, end) {}
```

metagraph/src/graph/alignment/aligner_helper.hpp
```diff
@@ -121,6 +123,137 @@ class Cigar {
...
+  public:
+    typedef Cigar::Operator value_type;
+    typedef const Cigar::Operator& reference;
+    typedef const Cigar::Operator* pointer;
+    typedef int difference_type;
+    typedef std::random_access_iterator_tag iterator_category;
+
+    CigarOpIterator(const Cigar &cigar) : CigarOpIterator(cigar, cigar.begin()) {}
+    CigarOpIterator(const Cigar &cigar, size_t offset)
+          : CigarOpIterator(cigar) {
```
karasikov, line 136:
```suggestion
    CigarOpIterator(const Cigar &cigar, size_t offset) : CigarOpIterator(cigar) {
```

metagraph/src/graph/alignment/aligner_helper.hpp
```diff
@@ -412,6 +561,157 @@ std::ostream& operator<<(std::ostream& out, const Alignment<NodeType> &alignment
     return out;
 }
 
+template <typename NodeType>
+class AlignmentSuffix {
+  public:
+    AlignmentSuffix(const Alignment<NodeType> &alignment, const DBGAlignerConfig &config,
+                    size_t k)
```
karasikov, line 568:
```suggestion
    AlignmentSuffix(const Alignment<NodeType> &alignment,
                    const DBGAlignerConfig &config,
                    size_t k)
```

metagraph/src/graph/alignment/aligner_helper.hpp
```diff
@@ -412,6 +561,157 @@ std::ostream& operator<<(std::ostream& out, const Alignment<NodeType> &alignment
...
+
+    Cigar::Operator get_front_op() const { return *cigar_it_; }
+    const CigarOpIterator& get_op_it() const {
+        assert(cigar_begin_ <= cigar_it_);
+        assert(cigar_it_ <= cigar_end_);
+        return cigar_it_;
+    }
+
+    bool eof() const { return cigar_it_ == cigar_end_; }
+    bool reof() const { return cigar_it_ == cigar_begin_; }
```
karasikov, line 607: if eof = "end of file", what is that file here?

Why not implement `pos()` and compare it to `size()`, or something like that?

metagraph/src/graph/alignment/aligner_helper.hpp
```diff
@@ -412,6 +561,157 @@ std::ostream& operator<<(std::ostream& out, const Alignment<NodeType> &alignment
...
+std::ostream& operator<<(std::ostream &out, const AlignmentSuffix<NodeType> &suffix) {
+    out << Alignment<NodeType>(suffix);
+    return out;
+}
+
+template <typename NodeType>
+class AlignmentPrefix {
+  public:
+    AlignmentPrefix(const Alignment<NodeType> &alignment, const DBGAlignerConfig &config,
+                    const DeBruijnGraph &graph)
```
karasikov, line 640: Easier to read if they are all in one list (vertical or horizontal).

When we call a function, we can neglect this rule for parameters, but declarations are read more often, so it's worth trading a bit of space to improve the readability.

```suggestion
    AlignmentPrefix(const Alignment<NodeType> &alignment,
                    const DBGAlignerConfig &config,
                    const DeBruijnGraph &graph)
```

metagraph/src/graph/alignment/aligner_helper.hpp
```diff
@@ -412,6 +561,157 @@ std::ostream& operator<<(std::ostream& out, const Alignment<NodeType> &alignment
...
+        cigar_it_ = cigar_rbegin_;
+
+        assert(cigar_rbegin_ <= cigar_rend_);
+        assert(cigar_it_ <= cigar_rend_);
+    }
+
+    AlignmentPrefix& operator++();
+    AlignmentPrefix& operator--();
+
+    auto get_score() const { return score_; }
```
karasikov, line 665:
```suggestion
    score_t get_score() const { return score_; }
```

karasikov (changes requested):
For aligner_aggregator.hpp and aligner_helper.cpp, I just quickly skimmed through them

metagraph/src/cli/query.cpp
```diff
@@ -846,20 +846,40 @@ void align_sequence(std::string &name, std::string &seq,
...
+            }
+            cigar.pop_back();
+            concat.pop_back();
+            std::swap(seq, concat);
+            name = fmt::format(ALIGNED_SEQ_HEADER_FORMAT, name, seq, score, cigar);
+
         } else {
-            seq = const_cast<std::string&&>(match.get_sequence());
-        }
+            auto &match = alignments[0];
```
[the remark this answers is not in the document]
hmusta: This is part of the `else` for the inner `if (aligner_config.chain_alignments && alignments.size() > 1)` loop, but it's still contained in the outer `if (alignments.size())` loop.

PR #216, opened 2020-10-20
moved build_successor to from row_diff.cpp to row_diff_builder.cpp

karasikov:
* moved build_successor to from row_diff.cpp to row_diff_builder.cpp
* cleaned up row_diff_builder.hpp

### DIFF
--- a/metagraph/src/annotation/row_diff_builder.cpp
+++ b/metagraph/src/annotation/row_diff_builder.cpp
@@ -10,14 +10,174 @@
 #include "common/vectors/bit_vector_sd.hpp"
 #include "graph/annotated_dbg.hpp"
 
+constexpr uint64_t chunk_size = 1 << 20;
+
 namespace mtg {
 namespace annot {
 
 using namespace mtg::annot::binmat;
 using mtg::common::logger;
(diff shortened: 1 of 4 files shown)


metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -243,13 +403,13 @@ void convert_batch_to_row_diff(const std::string &graph_fname,
     });
 
     // free memory occupied by sources
-    std::vector<std::unique_ptr<ColumnCompressed<>>>().swap(sources);
+    sources.clear();
```
danieldanciu, line 406: I don't think this guarantees that memory will be freed - only swapping with an empty vector, or calling `shrink_to_fit` afterwards does.

danieldanciu (approved):

metagraph/src/annotation/row_diff_builder.cpp
karasikov: Yeah, the internal array may not be released, but all objects will be destroyed, that's enough here. As a residual, we'll have just a buffer for pointers

PR #219, opened 2020-10-21
Atomic sdsl::int_vector<> operations. Ability to allocate aligned sdsl::int_vector<> and sdsl::bit_vector

hmusta:
This also allows for `fill_count_vector` in the differential assembly code to be multithreaded.

### DIFF
--- a/metagraph/src/common/vectors/vector_algorithm.hpp
+++ b/metagraph/src/common/vectors/vector_algorithm.hpp
@@ -16,17 +16,17 @@ sdsl::bit_vector to_sdsl(const std::vector<bool> &vector);
 sdsl::bit_vector to_sdsl(const std::vector<uint8_t> &vector);
 
 template <class Vector>
-inline sdsl::int_vector<> pack_vector(const Vector &vector, uint8_t bits_per_number) {
+inline sdsl::int_vector<> pack_vector(const Vector &vector, uint8_t width) {
     if constexpr(std::is_same_v<Vector, sdsl::int_vector<>>) {
-          if (bits_per_number == vector.width())
-              return vector;
-      }
+        if (width == vector.width())
+            return vector;
+    }
 
-      sdsl::int_vector<> packed(vector.size(), 0, bits_per_number);
-      for (uint64_t i = 0; i < vector.size(); ++i) {
-          packed[i] = vector[i];
-      }
-      return packed;
+    sdsl::int_vector<> packed(vector.size(), 0, width);
+    for (uint64_t i = 0; i < vector.size(); ++i) {
+        packed[i] = vector[i];
+    }
+    return packed;
 }
 
 sdsl::int_vector<> pack_vector(sdsl::int_vector<>&& vector, uint8_t width);
--- a/metagraph/src/common/vectors/vector_algorithm.hpp
+++ b/metagraph/src/common/vectors/vector_algorithm.hpp
@@ -40,6 +40,8 @@ sdsl::int_vector<> pack_vector(sdsl::int_vector<>&& vector, uint8_t width);
  * for more details.
  */
 
+// TODO: make bool atomic a template parameter instead
+
 inline bool fetch_and_set_bit(uint64_t *v,
                               uint64_t i,
                               bool atomic = false,
@@ -327,11 +329,11 @@ inline uint64_t atomic_fetch(const sdsl::int_vector<> &vector,
                              uint64_t i,
                              std::mutex &backup_mutex,
                              int mo) {
-    size_t width = vector.width();
-    size_t bit_pos = i * width;
-    uint64_t mask = (1llu << width) - 1;
+    const size_t width = vector.width();
+    const size_t bit_pos = i * width;
+    const uint64_t mask = (1llu << width) - 1;
     if (width + 7 > 64) {
-        // there is no way to reliably modify without a mutex
+        // there is no way to reliably fetch without a mutex
         std::lock_guard<std::mutex> lock(backup_mutex);
         return vector[i];
     } else if ((bit_pos & 0x3F) + width <= 64) {
@@ -345,7 +347,7 @@ inline uint64_t atomic_fetch(const sdsl::int_vector<> &vector,
         return (__atomic_load_n(word, mo) >> (bit_pos & 0x7F)) & mask;
 #endif
     } else {
-        uint8_t shift = bit_pos & 0x7;
+        const uint8_t shift = bit_pos & 0x7;
         const uint8_t *word = &reinterpret_cast<const uint8_t*>(vector.data())[bit_pos >> 3];
         if (shift + width <= 8) {
             // read from a byte
@@ -372,30 +374,30 @@ inline uint64_t atomic_fetch_and_add(sdsl::int_vector<> &vector,
                                      uint64_t val,
                                      std::mutex &backup_mutex,
                                      int mo) {
-    size_t width = vector.width();
-    size_t bit_pos = i * width;
-    uint64_t mask = (1llu << width) - 1;
+    const size_t width = vector.width();
+    const size_t bit_pos = i * width;
+    const uint64_t mask = (1llu << width) - 1;
     if (width + 7 > 64) {
         // there is no way to reliably modify without a mutex
         std::lock_guard<std::mutex> lock(backup_mutex);
-        uint64_t old_val = vector[i];
+        const uint64_t old_val = vector[i];
         vector[i] += val;
         return old_val;
     } else if ((bit_pos & 0x3F) + width <= 64) {
         // element fits in an aligned word
         uint64_t *word = &vector.data()[bit_pos >> 6];
-        uint8_t shift = bit_pos & 0x3F;
+        const uint8_t shift = bit_pos & 0x3F;
         return (__atomic_fetch_add(word, val << shift, mo) >> shift) & mask;
 #if defined(MODE_TI) && defined(__CX16__)
     } else if ((bit_pos & 0x7F) + width <= 128) {
         // element fits in an aligned double word
         __uint128_t *word = &reinterpret_cast<__uint128_t*>(vector.data())[bit_pos >> 7];
-        uint8_t shift = bit_pos & 0x7F;
-        // TODO: GCC only generates cmpxchg16b instruction only with older __sync functions
+        const uint8_t shift = bit_pos & 0x7F;
+        // TODO: GCC only generates cmpxchg16b instruction with older __sync functions
         return (__sync_fetch_and_add(word, __uint128_t(val) << shift) >> shift) & mask;
 #endif
     } else {
-        uint8_t shift = bit_pos & 0x7;
+        const uint8_t shift = bit_pos & 0x7;
         uint8_t *word = &reinterpret_cast<uint8_t*>(vector.data())[bit_pos >> 3];
         if (shift + width <= 8) {
             // read from a byte
@@ -422,22 +424,22 @@ inline uint64_t atomic_exchange(sdsl::int_vector<> &vector,
                                 uint64_t val,
                                 std::mutex &backup_mutex,
                                 int mo) {
-    size_t width = vector.width();
-    size_t bit_pos = i * width;
-    uint64_t mask = (1llu << width) - 1;
+    const size_t width = vector.width();
+    const size_t bit_pos = i * width;
+    const uint64_t mask = (1llu << width) - 1;
     if (width + 7 > 64) {
         // there is no way to reliably modify without a mutex
         std::lock_guard<std::mutex> lock(backup_mutex);
-        uint64_t old_val = vector[i];
+        const uint64_t old_val = vector[i];
         vector[i] = val;
         return old_val;
     } else if ((bit_pos & 0x3F) + width <= 64) {
         // element fits in an aligned word
         uint64_t *word = &vector.data()[bit_pos >> 6];
-        uint8_t shift = bit_pos & 0x3F;
+        const uint8_t shift = bit_pos & 0x3F;
         uint64_t desired;
         uint64_t exp = *word;
-        uint64_t inv_mask = ~(mask << shift);
+        const uint64_t inv_mask = ~(mask << shift);
         val <<= shift;
         do {
             desired = val | (inv_mask & exp);
@@ -447,12 +449,12 @@ inline uint64_t atomic_exchange(sdsl::int_vector<> &vector,
     } else if ((bit_pos & 0x7F) + width <= 128) {
         // element fits in an aligned double word
         __uint128_t *word = &reinterpret_cast<__uint128_t*>(vector.data())[bit_pos >> 7];
-        uint8_t shift = bit_pos & 0x7F;
+        const uint8_t shift = bit_pos & 0x7F;
         __uint128_t desired;
         __uint128_t exp;
         __uint128_t inv_mask = ~(__uint128_t(mask) << shift);
         __uint128_t big_val = __uint128_t(val) << shift;
-        // TODO: GCC only generates cmpxchg16b instruction only with older __sync functions
+        // TODO: GCC only generates cmpxchg16b instruction with older __sync functions
         do {
             exp = *word;
             desired = big_val | (inv_mask & exp);
@@ -460,13 +462,13 @@ inline uint64_t atomic_exchange(sdsl::int_vector<> &vector,
         return (exp >> shift) & mask;
 #endif
     } else {
-        uint8_t shift = bit_pos & 0x7;
+        const uint8_t shift = bit_pos & 0x7;
         uint8_t *word = &reinterpret_cast<uint8_t*>(vector.data())[bit_pos >> 3];
         if (shift + width <= 8) {
             // read from a byte
             uint8_t desired;
             uint8_t exp = *word;
-            uint8_t inv_mask = ~(mask << shift);
+            const uint8_t inv_mask = ~(mask << shift);
             val <<= shift;
             do {
                 desired = val | (inv_mask & exp);
@@ -477,7 +479,7 @@ inline uint64_t atomic_exchange(sdsl::int_vector<> &vector,
             uint16_t *this_word = (uint16_t*)word;
             uint16_t desired;
             uint16_t exp = *this_word;
-            uint16_t inv_mask = ~(mask << shift);
+            const uint16_t inv_mask = ~(mask << shift);
             val <<= shift;
             do {
                 desired = val | (inv_mask & exp);
@@ -488,7 +490,7 @@ inline uint64_t atomic_exchange(sdsl::int_vector<> &vector,
             uint32_t *this_word = (uint32_t*)word;
             uint32_t desired;
             uint32_t exp = *this_word;
-            uint32_t inv_mask = ~(mask << shift);
+            const uint32_t inv_mask = ~(mask << shift);
             val <<= shift;
             do {
                 desired = val | (inv_mask & exp);
@@ -499,7 +501,7 @@ inline uint64_t atomic_exchange(sdsl::int_vector<> &vector,
             uint64_t *this_word = (uint64_t*)word;
             uint64_t desired;
             uint64_t exp = *this_word;
-            uint64_t inv_mask = ~(mask << shift);
+            const uint64_t inv_mask = ~(mask << shift);
             val <<= shift;
             do {
                 desired = val | (inv_mask & exp);
--- a/metagraph/src/common/vectors/vector_algorithm.hpp
+++ b/metagraph/src/common/vectors/vector_algorithm.hpp
@@ -839,7 +839,7 @@ aligned_int_vector(size_t size, uint64_t val, uint8_t width, size_t alignment) {
         typename sdsl::int_vector<t_width>::int_width_type m_width;
     };
     static_assert(sizeof(sdsl::int_vector<t_width>) == sizeof(int_vector_access));
-    assert(!width || t_width == width);
+    assert(!t_width || t_width == width);
 
     sdsl::int_vector<t_width> v;
     auto &v_cast = reinterpret_cast<int_vector_access&>(v);
--- a/metagraph/src/common/vectors/vector_algorithm.hpp
+++ b/metagraph/src/common/vectors/vector_algorithm.hpp
@@ -3,6 +3,7 @@
 
 #include <functional>
 #include <cassert>
+#include <cstdlib>
 
 #include <sdsl/int_vector.hpp>
 #include <sdsl/select_support_scan.hpp>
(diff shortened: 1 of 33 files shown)


metagraph/src/common/vectors/vector_algorithm.hpp
```diff
@@ -109,6 +110,192 @@ inline void unset_bit(uint64_t *v,
...
+        // element fits in an aligned word
+        const uint64_t *word = &vector.data()[bit_pos >> 6];
+        return (__atomic_load_n(word, memorder) >> (bit_pos & 0x3F)) & mask;
+#if defined(MODE_TI) && defined(__CX16__)
+    } else if ((bit_pos & 0x7F) + width <= 128) {
+        // element fits in an aligned double word
+        const __uint128_t *word = &reinterpret_cast<const __uint128_t*>(vector.data())[bit_pos >> 7];
+        return (__atomic_load_n(word, memorder) >> (bit_pos & 0x7F)) & mask;
+#endif
+    } else {
```
karasikov, line 133: move implementation to the bottom of the file. Same for other functions.

metagraph/src/common/vectors/vector_algorithm.hpp
```diff
@@ -109,6 +110,192 @@ inline void unset_bit(uint64_t *v,
...
+            return (__atomic_fetch_add((uint64_t*)word, count << shift, memorder) >> shift) & mask;
+        } else {
+            assert(false);
+        }
+    }
+
+    return 0;
+}
+
+inline uint64_t atomic_exchange(sdsl::int_vector<> &vector, uint64_t i, uint64_t val,
```
karasikov, line 205:
```suggestion
inline uint64_t atomic_exchange(sdsl::int_vector<> &vector,
                                uint64_t i,
                                uint64_t val,
```

metagraph/src/common/vectors/vector_algorithm.hpp
```diff
@@ -489,6 +676,49 @@ prev_bit(const t_int_vec &v,
 }
 
 
+// Return an int_vector whose limbs are aligned to alignment bytes.
+// This is useful for algorithms requiring access to larger words in the underlying
+// vector (atomic increment, SIMD, etc.)
+template <int width = 0>
+inline sdsl::int_vector<width> aligned_int_vector(size_t size = 0, uint64_t val = 0,
+                                                  uint8_t var_width = 0,
```
karasikov, line 684: just `width`

metagraph/src/common/vectors/vector_algorithm.hpp
```diff
@@ -489,6 +676,49 @@ prev_bit(const t_int_vec &v,
...
+                                                  size_t alignment = 8) {
```
karasikov, line 685: does it make sense to specify if this are bits or bytes?

metagraph/src/common/vectors/vector_algorithm.hpp
```diff
@@ -489,6 +676,49 @@ prev_bit(const t_int_vec &v,
...
+
+    memset(v_cast.m_data, 0, capacity_bytes);
+
+    if (val)
+        sdsl::util::set_to_value(v, val);
+
+    return v;
+}
+
+inline sdsl::bit_vector aligned_bit_vector(size_t size = 0, bool val = false,
```
karasikov, line 716:
```suggestion
inline sdsl::bit_vector aligned_bit_vector(size_t size = 0, bool val = 0,
```

metagraph/tests/annotation/test_annotated_graph_algorithm.cpp
```diff
@@ -104,9 +104,12 @@ void test_mask_indices(double density_cutoff) {
...
 
 TYPED_TEST(MaskedDeBruijnGraphAlgorithm, MaskIndicesByLabel) {
-    for (double d = 0.0; d <= 1.0; d += 0.05) {
-        test_mask_indices<typename TypeParam::first_type,
-                          typename TypeParam::second_type>(d);
+    for (size_t num_threads : { 1, 4 }) {
+        for (double d = 0.0; d <= 1.0; d += 0.05) {
+            test_mask_indices<typename TypeParam::first_type, typename TypeParam::second_type>(
+                d, num_threads
+            );
```
karasikov, line 111:
```suggestion
            test_mask_indices<typename TypeParam::first_type,
                              typename TypeParam::second_type>(d, num_threads);
```

metagraph/tests/test_vector_algorithm.cpp
```diff
@@ -160,6 +160,118 @@ TEST(IntVector, call_nonzeros_sparse_every_4) {
...
+            sdsl::int_vector<> vector_atomic = aligned_int_vector(200, 0, w, 16);
+            sdsl::int_vector<> vector(200, 0, w);
+            ASSERT_LT(65u, vector.size());
+            std::atomic<size_t> counter = 0;
+
+            std::mutex mu;
+            std::mutex mu2;
+            #pragma omp parallel for num_threads(3) schedule(dynamic)
+            for (size_t j = 0; j < 3; ++j) {
+                for (size_t i = j; i < vector.size(); i += 3) {
```
karasikov, line 175:
```c++
            #pragma omp parallel for num_threads(3) schedule(static, 1)
            for (size_t i = 0; i < vector.size(); ++i) {
```
will test the same thing, you don't need two loops

metagraph/tests/test_vector_algorithm.cpp
```diff
@@ -160,6 +160,118 @@ TEST(IntVector, call_nonzeros_sparse_every_4) {
...
+                    if (i % 130 == 1) {
+                        uint64_t val = (++counter) % w;
+                        EXPECT_EQ(0u, atomic_exchange(vector_atomic, i, val, mu, memorder));
+                        std::lock_guard<std::mutex> lock(mu2);
+                        vector[i] = val;
+                    }
+                }
+            }
```
karasikov, line 183: Here each bit is changed only once by a single thread.
Do you also have a test where you write to the same bits?

karasikov: same comment for the other tests

metagraph/tests/test_vector_algorithm.cpp
karasikov: same comment for the other tests

metagraph/src/common/vectors/vector_algorithm.hpp
```diff
@@ -109,6 +110,192 @@ inline void unset_bit(uint64_t *v,
...
+        } else {
+            assert(false);
+        }
+    }
+
+    return 0;
+}
+
+inline uint64_t atomic_fetch_and_add(sdsl::int_vector<> &vector, uint64_t i,
+                                     uint64_t count,
```
karasikov, line 157: what is count?

I had to look at the code to get this.

Call it 'delta', or 'x', or 'val', ...

karasikov (commented):
checked interfaces

metagraph/src/graph/annotated_graph_algorithm.cpp
```diff
@@ -90,32 +92,44 @@ mask_nodes_by_unitig_labels(const AnnotatedDBG &anno_graph,
 
         return (in_count >= label_in_factor * path.size())
             && (other_count <= label_other_fraction * (in_count + out_count + other_count));
-    });
+    }, num_threads);
 }
 
 sdsl::int_vector<> fill_count_vector(const AnnotatedDBG &anno_graph,
```
karasikov, line 98: maybe `fill_diff_label_count_vector`?

metagraph/tests/test_vector_algorithm.cpp
hmusta: Since edits are made to the vector in blocks of variable width (ranging from 1 to 32 bytes), this test will fail if the operations are not atomic.

But you're right, it's also good to have another test where the same positions are written as well.

metagraph/src/common/vectors/vector_algorithm.hpp
hmusta: I'll call it val to be consistent with `atomic_exchange`

metagraph/tests/test_vector_algorithm.cpp
hmusta: I can also make the modulus smaller to make this test more rigourous.

karasikov: yes, the test should be different this is too simple. Obviously fetch will return 0 if you haven't done any midifications. You need to run something more complex. Otherwise, this test is useless.

karasikov: for instance, you can flip every bit 1001 times by every thread and in the end check that all bits were flipped from 0 to 1.

karasikov: similar for other functions, like fetch_and_add

karasikov: Also, remove all locks from tests where you test atomic operations, they add some synchronization which you don't want.

metagraph/src/common/vectors/vector_algorithm.hpp
hmusta: Usually alignment is specified in bytes

metagraph/src/graph/annotated_graph_algorithm.cpp
hmusta: there's no diff happening at this point, so how about just `construct_label_count_vector`?

karasikov: But there are in and out groups, so I thought about DIFFerential counts

hmusta: how about `construct_interleaved_label_count_vector`?

metagraph/tests/test_vector_algorithm.cpp
hmusta: I replaced the tests with those which should affect overlapping indexes in multiple threads.

metagraph/src/common/vectors/vector_algorithm.hpp
hmusta: Any particular reason why?

metagraph/tests/test_vector_algorithm.cpp
```diff
@@ -160,106 +160,66 @@ TEST(IntVector, call_nonzeros_sparse_every_4) {
...
+            std::atomic_thread_fence(std::memory_order_release);
             #pragma omp parallel for num_threads(3) schedule(static, 1)
-            for (size_t i = 0; i < vector.size(); ++i) {
-                if (i % 130 == 1) {
-                    EXPECT_EQ(0u, atomic_fetch_and_add(vector_atomic, i, 1, mu, memorder));
-                    std::lock_guard<std::mutex> lock(mu2);
-                    ++vector[i];
+            for (size_t j = 0; j < val; ++j) {
+                for (size_t i = 0; i < vector_atomic.size(); ++i) {
+                    atomic_fetch_and_add(vector_atomic, i, 1, mu, memorder);
```
karasikov, line 193: maybe also try updating only a single element of the vector `atomic_fetch_and_add(vector_atomic, 0, ...)` to increase the contention. Otherwise, it's quite unlikely that two threads are updating exactly the same element at the same time

metagraph/src/graph/annotated_graph_algorithm.cpp
karasikov: Hm, try something else?

metagraph/tests/test_vector_algorithm.cpp
```diff
@@ -160,6 +160,178 @@ TEST(IntVector, call_nonzeros_sparse_every_4) {
...
+
+                #pragma omp task
+                {
+                    EXPECT_EQ(val, atomic_fetch(vector_atomic, i, mu, __ATOMIC_ACQUIRE));
+                }
+            }
+        }
+
+        // ensure everything is synced before deallocating
+        std::atomic_thread_fence(std::memory_order_acquire);
```
karasikov, line 229: is this really needed? I don't think so

metagraph/src/graph/annotated_graph_algorithm.cpp
hmusta: or `construct_in_out_label_count_vector`? If not, we can do `construct_diff_label_count_vector`. This is not in the external API anyway, so it's name is not terribly important.

metagraph/tests/test_vector_algorithm.cpp
karasikov: please make sure each of these tests runs less than 1 sec, better 0.1 sec.

metagraph/tests/test_vector_algorithm.cpp
```diff
@@ -160,6 +160,172 @@ TEST(IntVector, call_nonzeros_sparse_every_4) {
...
+
+                            size_t r = 2;
+                            size_t i_min = i - std::min(i, r);
+                            size_t i_max = std::min(i + r, size_t(vector_atomic.size()));
+                            for (size_t k = i_min; k < i_max; ++k) {
+                                atomic_fetch_and_add(vector_atomic, k, 0, mu, memorder);
+                            }
+                        }
+                    }
+                }
```
karasikov, line 320: if this works with __ATOMIC_RELAXED, why test others?

hmusta: Originally I had it set up so that if any checks failed, it would output the memory order and I'd know if the issue was due to memory order or something else. Though in the future if issues do come up, one could just modify this test to output the memory order.

metagraph/src/common/vectors/vector_algorithm.hpp
karasikov: Minor point. I think it's just easier to read when it is in this context -- initialize a vector, even though it's binary. Otherwise feels like this may be another flag like `async`, or something...

metagraph/src/common/vectors/vector_algorithm.hpp
```diff
@@ -233,11 +628,11 @@ void call_zeros(const Bitmap &vector,
...
-void call_zeros(const sdsl::bit_vector &vector,
-                uint64_t begin, uint64_t end,
-                Callback callback,
-                bool atomic,
-                int mo = __ATOMIC_SEQ_CST) {
+inline void call_zeros(const sdsl::bit_vector &vector,
+                       uint64_t begin, uint64_t end,
+                       Callback callback,
+                       bool atomic,
+                       int mo) {
```
karasikov, line 635: Don't put too many inlines, it might take longer to compile and make the binary larger.

This function executes a large block from `begin` to `end`.

metagraph/src/common/vectors/vector_algorithm.hpp
```diff
@@ -15,26 +16,14 @@ sdsl::bit_vector to_sdsl(const std::vector<bool> &vector);
...
 
-sdsl::int_vector<> pack_vector(sdsl::int_vector<>&& vector,
-                               uint8_t bits_per_number);
+sdsl::int_vector<> pack_vector(sdsl::int_vector<>&& vector, uint8_t bits_per_number);
 
 
 /**
  * Atomic bit fetching, setting, and unsetting on packed vectors.
- * fetch_and_* return the old values. The default memorder __ATOMIC_SEQ_CST
+ * fetch_and_* return the old values. The default mo __ATOMIC_SEQ_CST
```
karasikov, line 26: memory order

metagraph/src/common/vectors/vector_algorithm.hpp
```diff
@@ -15,26 +16,14 @@ sdsl::bit_vector to_sdsl(const std::vector<bool> &vector);
...
-            return vector;
-    }
-
-    sdsl::int_vector<> packed(vector.size(), 0, bits_per_number);
-    for (uint64_t i = 0; i < vector.size(); ++i) {
-        packed[i] = vector[i];
-    }
-    return packed;
-}
+inline sdsl::int_vector<> pack_vector(const Vector &vector, uint8_t bits_per_number);
```
karasikov, line 19: don't overdo :) this was a small function, the implementation helps to see what is the complexity. Here it's better to keep it here.

metagraph/src/graph/annotated_graph_algorithm.cpp
```diff
@@ -184,9 +200,9 @@ mask_nodes_by_node_label(const AnnotatedDBG &anno_graph,
         // If at least one infrequent label exists, construct a count vector to
         // reduce calls to the annotator
         if (labels_in_infrequent.size() || labels_out_infrequent.size()) {
-            sdsl::int_vector<> counts = fill_count_vector(anno_graph,
-                                                          labels_in_infrequent,
-                                                          labels_out_infrequent);
+            sdsl::int_vector<> counts = construct_diff_label_count_vector(
+                anno_graph, labels_in_infrequent, labels_out_infrequent, num_threads
+            );
```
karasikov, line 205: was better

metagraph/src/common/vectors/vector_algorithm.hpp
hmusta: I'm not sure, but I think I at least need to make the initial declaration `inline`, otherwise it doesn't compile since it complains about having multiple instantiations.

metagraph/src/common/vectors/vector_algorithm.hpp
```diff
@@ -15,26 +16,25 @@ sdsl::bit_vector to_sdsl(const std::vector<bool> &vector);
...
-    return packed;
+          if (bits_per_number == vector.width())
+              return vector;
+      }
+
+      sdsl::int_vector<> packed(vector.size(), 0, bits_per_number);
+      for (uint64_t i = 0; i < vector.size(); ++i) {
+          packed[i] = vector[i];
+      }
+      return packed;
```
karasikov, line 29: fix indentation

metagraph/src/common/vectors/vector_algorithm.hpp
```diff
@@ -15,26 +16,25 @@ sdsl::bit_vector to_sdsl(const std::vector<bool> &vector);
...
 }
 
-sdsl::int_vector<> pack_vector(sdsl::int_vector<>&& vector,
-                               uint8_t bits_per_number);
+sdsl::int_vector<> pack_vector(sdsl::int_vector<>&& vector, uint8_t bits_per_number);
```
karasikov, line 32:
```suggestion
sdsl::int_vector<> pack_vector(sdsl::int_vector<>&& vector, uint8_t width);
```

karasikov (approved):

danieldanciu (commented):
All looks fine, my only concern is the design decision to make atomic a parameter rather than a compile time decision in the `set/fetch` operations.

metagraph/src/common/vectors/vector_algorithm.hpp
```diff
@@ -78,42 +301,221 @@ inline bool fetch_and_unset_bit(uint64_t *v,
...
+            return (__atomic_fetch_add((uint64_t*)word, val << shift, mo) >> shift) & mask;
+        } else {
+            assert(false);
+        }
+    }
+
+    return 0;
+}
+
+inline uint64_t atomic_exchange(sdsl::int_vector<> &vector,
```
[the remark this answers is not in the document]
hmusta: In the GCC builtins, setting a new value and returning the old one is called `exchange` instead of `set`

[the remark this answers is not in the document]
hmusta: The closely related methods implemented for bit vectors are named fetch_and_(un)set because there's no variable passed with which the stored value should be swapped.

metagraph/src/common/vectors/vector_algorithm.hpp
```diff
@@ -78,42 +301,221 @@ inline bool fetch_and_unset_bit(uint64_t *v,
...
 
+inline uint64_t atomic_fetch(const sdsl::int_vector<> &vector,
+                             uint64_t i,
+                             std::mutex &backup_mutex,
+                             int mo) {
+    size_t width = vector.width();
+    size_t bit_pos = i * width;
+    uint64_t mask = (1llu << width) - 1;
+    if (width + 7 > 64) {
+        // there is no way to reliably modify without a mutex
```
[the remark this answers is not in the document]
hmusta: oops, this comment was copied from the other methods. Yes, this only does a fetch and doesn't modify anything

metagraph/src/common/vectors/vector_algorithm.hpp
```diff
@@ -78,42 +301,221 @@ inline bool fetch_and_unset_bit(uint64_t *v,
     }
 }
 
-inline bool fetch_bit(const uint64_t *v,
-                      uint64_t i,
-                      bool atomic = false,
-                      int mo = __ATOMIC_SEQ_CST) {
+inline bool fetch_bit(const uint64_t *v, uint64_t i, bool atomic, int mo) {
```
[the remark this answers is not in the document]
hmusta: It was done this way because we originally thought that the atomic operations would introduce too much overhead in the traversal operations, so we wanted to make them variable so that a non-atomic version would be used in the single threaded case. However, it turned out in the end that the overhead is negligible relative to the rest of the operations and we figured that the branch predictor would learn that the same branch is always taken here.

metagraph/src/common/vectors/vector_algorithm.hpp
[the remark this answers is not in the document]
karasikov: I like `atomic_fetch_and_set`, but `exchange` is also fine.
I agree that `atomic_set` doesn't make it clear what is returned.

metagraph/src/common/vectors/vector_algorithm.hpp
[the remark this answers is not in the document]
danieldanciu: I am not sure if branch prediction is also able to amortize the effects of the jmp. Maybe it is. But it still seems to me that we can easily avoid that overhead, so why don't we? :)

danieldanciu (approved):

metagraph/src/common/vectors/vector_algorithm.hpp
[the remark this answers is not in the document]
hmusta: I'll put a TODO so we can handle this in a different pull request later, since it'll require changing more stuff and merging might get messy.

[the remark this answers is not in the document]
karasikov: Doesn't jmp cost zero if branch prediction works?

Do you know if there is any overhead of using atomic operations at all? Suppose you have a single thread, so no atomicity is needed, but you still call these atomic operations.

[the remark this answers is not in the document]
danieldanciu: I don't know what the cost is if the prediction works, the documentation was unclear. It seemed to indicate that the overhead of jmp is still there, but maybe because it's done ahead of time is negligible dunno.

Regarding atomic operations: as far as I know, atomic operations simply block the cache line to ensure atomicity, so they cause an overhead if there is another processor wanting to use that cache line. No idea what that involves in practice.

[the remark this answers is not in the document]
hmusta: When I was parallelizing the sequence extraction method, I benchmarked it with and without atomic operations and the overhead was negligible (<1%), so the other graph traversal operations dominate.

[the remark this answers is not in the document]
hmusta: But I've read that the overhead can be 10x in cases of high contention.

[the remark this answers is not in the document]
karasikov: Yeah, I agree. In graph traversal we cal always use atomic calls. But I am not sure we should remove this feature from the interface completely. Maybe it will be useful in other cases? But I don’t know.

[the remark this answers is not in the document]
hmusta: Regarding branch prediction: there is some overhead because the if statement still has to be evaluated eventually for every call, but there won't be any branch misprediction overhead.

PR #220, opened 2020-10-21
cleanup in row-diff build_successor

karasikov:
just a cleanup and minor optimization (don't write predecessors for dummy nodes, which are not annotated)

### DIFF
--- a/metagraph/src/annotation/row_diff_builder.cpp
+++ b/metagraph/src/annotation/row_diff_builder.cpp
@@ -88,7 +88,7 @@ void build_successor(const std::string &graph_fname,
         std::vector<std::vector<uint64_t>> succ_buf(num_threads);
         std::vector<std::vector<bool>> pred_boundary_buf(num_threads);
 
-        // use static scheduling to make thread run on ordered contiguous blocks
+        // use static scheduling to make threads process ordered contiguous blocks
         #pragma omp parallel for num_threads(num_threads) schedule(static)
         for (uint64_t i = start; i < std::min(start + batch_size, graph.num_nodes() + 1); ++i) {
             const size_t r = omp_get_thread_num();
--- a/metagraph/src/annotation/row_diff_builder.cpp
+++ b/metagraph/src/annotation/row_diff_builder.cpp
@@ -100,11 +100,11 @@ void build_successor(const std::string &graph_fname,
                 uint64_t next = d ? graph.boss_to_kmer_index(boss.fwd(boss_idx, d)) : 0;
                 succ_buf[r].push_back(next);
             }
-            BOSS::TAlphabet d = boss.get_node_last_value(boss_idx);
-            uint64_t back_idx = boss.bwd(boss_idx);
             // ignore predecessors if boss_idx is not the last outgoing
             // edge (bc. we only traverse the last outgoing at a bifurcation)
-            if (!dummy[boss_idx] && boss.fwd(back_idx, d) == boss_idx) {
+            if (!dummy[boss_idx] && boss.get_last(boss_idx)) {
+                BOSS::TAlphabet d = boss.get_node_last_value(boss_idx);
+                uint64_t back_idx = boss.bwd(boss_idx);
                 boss.call_incoming_to_target(back_idx, d,
                     [&](BOSS::edge_index pred) {
                         // terminal and dummy predecessors are ignored
--- a/metagraph/src/annotation/row_diff_builder.cpp
+++ b/metagraph/src/annotation/row_diff_builder.cpp
@@ -81,7 +81,11 @@ void build_successor(const std::string &graph_fname,
     ProgressBar progress_bar(graph.num_nodes(), "Compute successors", std::cerr,
                              !common::get_verbose());
 
+#ifndef NDEBUG
     const uint64_t batch_size = 10'000'000;
+#else
+    const uint64_t batch_size = 1'000;
+#endif
     // traverse BOSS table in parallel processing blocks of size |batch_size|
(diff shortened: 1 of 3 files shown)


metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -77,60 +78,54 @@ void build_successor(const std::string &graph_fname,
...
-    constexpr uint64_t chunk_size = 1'000'000;
-    const uint64_t batch_size = chunk_size * num_threads;
-    const uint64_t batch_count = (graph.num_nodes() - 1) / batch_size + 1;
-
-    ProgressBar progress_bar(batch_count, "Compute successors", std::cerr,
+    ProgressBar progress_bar(graph.num_nodes(), "Compute successors", std::cerr,
                              !common::get_verbose());
 
-    for (uint64_t batch = 0; batch < batch_count; ++batch) {
+    const uint64_t batch_size = 10'000'000;
```
danieldanciu, line 84: I would keep this at 1M unless you notice a speedup. I think I tried 10M and didn't notice a difference.

If you make it 10M none of the tests will be able to test the case when there is more than one batch. I believe the transcripts_1000 file has just a bit over 1M nodes.

danieldanciu (approved):

metagraph/src/annotation/row_diff_builder.cpp
karasikov: it was 1M * num_threads, so I made it larger. But yeah, I was hesitant to merge it yet because we don't have proper tests for this. (We also must build it using multiple threads, which is not done, I think, for transcripts_1000).

danieldanciu: That's true, though it was still testable, because we were only using 1 thread in transform_anno. So at least we tested the multiple batches, even if we didn't test the multithreading.
One thing people do in such cases is to add a method `set_batch_size_for_testing` that makes it clear that this is intended for testing only. We can then use this method to make the batch size 10 in unit tests. Won't work for integration tests, but maybe that's ok.

metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -77,60 +78,54 @@ void build_successor(const std::string &graph_fname,
...
+        // use static scheduling to make threads process ordered contiguous blocks
+        #pragma omp parallel for num_threads(num_threads) schedule(static)
+        for (uint64_t i = start; i < std::min(start + batch_size, graph.num_nodes() + 1); ++i) {
+            const size_t r = omp_get_thread_num();
+            uint64_t boss_idx = graph.kmer_to_boss_index(i);
+            if (dummy[boss_idx] || terminal[boss_idx]) {
+                succ_buf[r].push_back(0);
+            } else {
+                const BOSS::TAlphabet d = boss.get_W(boss_idx) % boss.alph_size;
+                uint64_t next = d ? graph.boss_to_kmer_index(boss.fwd(boss_idx, d)) : 0;
```
hmusta, line 100: Instead of called `fwd` on all non-dummy, non-terminal indices, won't it be faster to just call `rank(start, s)` for every character `s`, then keep updating a rank counter as you iterate through (since you already compute `get_W`)?

hmusta: So this way, the ranks will be precomputed and you'll only have to compute `select` for each iteration.

hmusta: One problem is that every time you hit a dummy or a terminal node, you'll have to resync the counts either with extra `get_W` or `rank` calls, which will slow it down a bit.

karasikov: It sounds like significantly more code to me, but I don't know how faster is this going to make the whole function. (We also call bwd, etc.)

karasikov: Also, this whole function is not critical to the application. If we can easily make it faster, we should do it, but if it's something more than straightforward, I don't it's worth it.

metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -77,60 +78,54 @@ void build_successor(const std::string &graph_fname,
...
+                succ_buf[r].push_back(next);
+            }
+            BOSS::TAlphabet d = boss.get_node_last_value(boss_idx);
+            uint64_t back_idx = boss.bwd(boss_idx);
+            // ignore predecessors if boss_idx is not the last outgoing
+            // edge (bc. we only traverse the last outgoing at a bifurcation)
+            if (!dummy[boss_idx] && boss.fwd(back_idx, d) == boss_idx) {
```
hmusta, line 107: Isn't this sufficient?
```suggestion
            // ignore predecessors if boss_idx is not the last outgoing
            // edge (bc. we only traverse the last outgoing at a bifurcation)
            if (!dummy[boss_idx] && boss.select_last(boss_idx)) {
                BOSS::TAlphabet d = boss.get_node_last_value(boss_idx);
                uint64_t back_idx = boss.bwd(boss_idx);
```

hmusta: (the diff highlighting is weird for some reason...)

karasikov: yes

metagraph/src/annotation/row_diff_builder.cpp
karasikov: I'll make the batch size equal to 1000, so integration tests in Debug mode will cover this case.

PR #221, opened 2020-10-22
precompute dummy k-mers in call_sequences_row_diff

karasikov:
This should make the traversal slightly faster.
Also prepares for another optimization in row-diff where we start traversal from the sink nodes and must have source dummy nodes precomputed to know where to stop.

### DIFF
--- a/metagraph/src/graph/representation/succinct/boss.cpp
+++ b/metagraph/src/graph/representation/succinct/boss.cpp
@@ -31,6 +31,7 @@ namespace graph {
 namespace boss {
 
 using utils::remove_suffix;
+using common::logger;
 
 #define CHECK_INDEX(idx) \
     assert(idx < W_->size()); \
@@ -1420,44 +1421,46 @@ void BOSS::edge_DFT(edge_index start,
  * Traverse the entire dummy subtree
  * and find all redundant dummy edges.
  */
-template <typename Array, typename U>
+template <class Callback>
 void traverse_dummy_edges(const BOSS &graph,
                           edge_index subtree_root,
                           size_t check_depth,
-                          Array *redundant_mask,
-                          Array *traversed_mask,
-                          U *num_dummy_traversed,
-                          bool verbose) {
+                          const Callback &call_edge,
+                          sdsl::bit_vector *redundant_mask,
+                          sdsl::bit_vector *traversed_mask) {
     assert(!redundant_mask || redundant_mask->size() == graph.get_W().size());
     assert(!traversed_mask || traversed_mask->size() == graph.get_W().size());
 
     if (!check_depth)
         return;
 
+    constexpr bool async = true;
+
     std::vector<bool> redundant_path;
-    uint64_t num_edges_traversed = 0;
 
     // start traversal in the given node
     graph.edge_DFT(subtree_root,
         [&](edge_index) {
             redundant_path.push_back(true);
-            num_edges_traversed++;
         },
         [&](edge_index edge) {
             assert(graph.get_W(edge) < graph.alph_size);
 
+            call_edge(edge, redundant_path.size());
+
+            // TODO: remove this mask and use |call_edge| instead
             if (traversed_mask)
-                (*traversed_mask)[edge] = 1;
+                set_bit(traversed_mask->data(), edge, async, __ATOMIC_RELAXED);
 
             if (redundant_path.back())
-                (*redundant_mask)[edge] = 2;
+                set_bit(redundant_mask->data(), edge, async);
 
             redundant_path.pop_back();
         },
         [&](edge_index edge) {
             if (redundant_path.size() == check_depth) {
                 if (!redundant_mask
-                        || (!(*redundant_mask)[edge]
+                        || (!fetch_bit(redundant_mask->data(), edge, async)
                                 && graph.is_single_incoming(edge, graph.get_W(edge)))) {
                     // the last dummy edge is not redundant and hence the
                     // entire path has to remain in the graph
@@ -1470,11 +1473,6 @@ void traverse_dummy_edges(const BOSS &graph,
             }
         }
     );
-    *num_dummy_traversed += num_edges_traversed;
-    if (verbose) {
-        std::cout << "Source dummy edges traversed: "
-                        + std::to_string(*num_dummy_traversed) + "\n" << std::flush;
-    }
     assert(redundant_path.empty());
 }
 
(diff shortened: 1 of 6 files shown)


metagraph/src/graph/representation/succinct/boss.cpp
```diff
@@ -1486,43 +1484,40 @@ uint64_t traverse_dummy_edges(const BOSS &graph,
...
     assert(!redundant_mask || redundant_mask->size() == graph.get_W().size());
     assert(!traversed_mask || traversed_mask->size() == graph.get_W().size());
 
+    // assume the main dummy source node already traversed
+    std::atomic<uint64_t> num_dummy_traversed { 1 };
+    on_dummy(1, 0); // call the main dummy source edge
     if (traversed_mask)
         (*traversed_mask)[1] = true;
 
     if (graph.get_last(1))
```
danieldanciu, line 1498: maybe add a short comment, as I kept wondering how this thing works for cycles until I saw this if:
`if (graph.get_last(1)) // graph has no dummy k-mers`

metagraph/src/graph/representation/succinct/boss.cpp
```diff
@@ -2373,10 +2364,8 @@ void update_terminal_bits(
         set_bit(terminal->data(), path[i + max_length - 1], async);
     }
 
-    if (path.size() % max_length == 0) { // last node is terminal
-        callback(std::move(path), nullopt);
+    if (path.size() % max_length == 0) // last node is terminal
```
danieldanciu, line 2367: Hmm, I think we should also set the previous nodes as near_terminal here.
Actually, I think the way I assigned near terminals in the entire method is not optimal. 
Every node on the path should have the near-terminal bit set, because it is by definition near a terminal. This way, if we later merge into a node on this path, we won't need to set a terminal node. I can make this change later if you add a todo, should be pretty easy.

danieldanciu (approved):

metagraph/src/graph/representation/succinct/boss.cpp
```diff
@@ -1486,43 +1484,40 @@ uint64_t traverse_dummy_edges(const BOSS &graph,
...
                               sdsl::bit_vector *traversed_mask,
                               size_t num_threads,
-                              bool verbose) {
+                              const std::function<void(edge_index, size_t)> &on_dummy
+                                        = [](edge_index, size_t) {}) {
     assert(!redundant_mask || redundant_mask->size() == graph.get_W().size());
     assert(!traversed_mask || traversed_mask->size() == graph.get_W().size());
 
+    // assume the main dummy source node already traversed
+    std::atomic<uint64_t> num_dummy_traversed { 1 };
```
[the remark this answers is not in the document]
hmusta: It's a minor point, but doing `= 1` uses an atomic operation to set the value, whereas brace initialization or `num_dummy_traversed(1)` avoids this.

metagraph/src/graph/representation/succinct/boss.cpp
```diff
@@ -2649,52 +2626,71 @@ void BOSS::call_sequences(Call<std::string&&, std::vector<edge_index>&&> callbac
...
     // keep track of the edges that have been reached
-    sdsl::bit_vector visited(W_->size(), false);
-    sdsl::bit_vector near_terminal(W_->size(), false);
+    sdsl::bit_vector visited = *dummy;
+
+    constexpr bool async = true;
+
+    // traverse all dummy source k-mers but leave last source dummy edges
+    // as not visited
+    traverse_dummy_edges(*this, NULL, NULL, num_threads,
```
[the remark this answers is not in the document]
karasikov: It still traversal, so for large graphs, you better use as many threads as you have.
This should be a bit faster than normal traversal. But I have never actually benchmarked it.

metagraph/src/graph/representation/succinct/boss.cpp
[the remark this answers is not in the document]
karasikov: Yeah, this looks weird. Maybe I thought it doesn't have an assignment constructor... Or maybe I copied this from some other place where it's necessary. I'll make it `= 1`.

metagraph/src/graph/representation/succinct/boss.cpp
karasikov: But if every node is a near terminal, this "near terminal" doesn't mean anything and will propagate forever, so there will be no anchors assigned. Right?

metagraph/src/graph/representation/succinct/boss.cpp
karasikov: will do

metagraph/src/graph/representation/succinct/boss.cpp
karasikov: What we can do though is start backward traversal from this position. And process all diff-paths that merge into this anchor. I implemented it for starting from sink nodes (see soon in the next PR).
That processes a large portion of the graph, but the loopy tricky parts are handled by this approach.
So what we can do is to invoke the same backward traversal thing from the newly created anchors.

Hm, actually then we don't need the near-terminal thing at all, it seems.
This thing was supposed to mark all nodes that merge into an anchor soon (say, this was path X of near-anchor nodes). So that if there is another path (call it Y) that merges somewhere into X, we don't create an anchor in the merge because we know that it's close to an anchor in X.

But, with the backward traversal thing, we could cover all such Y from X (if Y merges into X, Y will be completely covered by a backward traversal from X).

Will implement this in the next PR then.

metagraph/src/graph/representation/succinct/boss.cpp
```diff
@@ -1486,43 +1484,40 @@ uint64_t traverse_dummy_edges(const BOSS &graph,
...
         (*traversed_mask)[1] = true;
 
     if (graph.get_last(1))
         return 1;
 
-    if (num_threads <= 1 || graph.get_k() <= 1) {
-        // assume the main dummy source node already traversed
-        uint64_t num_dummy_traversed = 1;
+    if (graph.get_k() <= 1) {
         // start traversal in the main dummy source node
```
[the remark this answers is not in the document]
karasikov: yeah, moved the comment up, so it's less confusing

metagraph/src/graph/representation/succinct/boss.cpp
danieldanciu: Yes, if we do backward traversal, near terminals are not needed.

Note that they don't propagate forever. A merge into a near terminal doesn't make the merge nodes near-terminal. So the max distance of propagation is still 2 * max_depth

karasikov: Maybe I misinterpreted your
> Every node on the path should have the near-terminal bit set, because it is by definition near a terminal.

On a different note, I realized that even with backward traversal running this with concurrently with multiple threads is tricky, so I decided to keep `near_terminal` for now.

danieldanciu: Right, more precisely I should say:
 > Every node on the path that is succeeded by a terminal node should have the near-terminal bit set, because it is by definition near a terminal.
So the only nodes that are *not* near terminal are the nodes between the last terminal node and the end of the path.

PR #222, opened 2020-10-23
traverse row-diff paths backwards for initial anchor assignment

karasikov:
This should significantly reduce the number of initially assigned anchors and make it closer to the desired value.

---
Result of an experiment:
The backward traversal processes more than 98% of the graph (24750k subset of Microbe), which is very good, because for this part the anchors are assigned optimally.
```
...
[2020-10-25 22:25:19.928]  98.7%, 6170180163/6251449537
[2020-10-25 22:25:20.676] [info] Backward traversal from sinks is done. Finishing up the rest...
[2020-10-25 22:25:21.572]  98.8%, 6176431612/6251449537
...
[2020-10-25 22:26:03.158] 100.0%, 6251449537/6251449537
```

### DIFF
--- a/metagraph/src/graph/representation/succinct/boss.cpp
+++ b/metagraph/src/graph/representation/succinct/boss.cpp
@@ -2819,24 +2819,16 @@ void BOSS::row_diff_traverse(size_t num_threads,
         if (fetch_and_set_bit(visited.data(), rep, async))
             return;
 
-        ++progress_bar;
-
         for (edge_index idx : path) {
             std::ignore = idx;
             assert(idx == rep || !fetch_bit(visited.data(), idx, async));
             set_bit(visited.data(), idx, async);
-            ++progress_bar;
         }
 
-        std::optional<edge_index> anchor;
-        for (uint64_t i = 0; i + max_length <= path.size(); i += max_length) {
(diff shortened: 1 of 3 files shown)


danieldanciu (approved):
This looks great!

karasikov:
Just finished an experiment: this actually processes more than 98% of the nodes.
The rest is processed by the old algorithm with marking near-anchor nodes.

metagraph/src/graph/representation/succinct/boss.cpp
```diff
@@ -2623,6 +2624,72 @@ void BOSS::call_sequences(Call<std::string&&, std::vector<edge_index>&&> callbac
...
+        }
+    );
+
+    while (queue.size()) {
+        T dist_to_anchor;
+        std::tie(edge, dist_to_anchor) = queue.back();
+        queue.pop_back();
+
+        // mark as visited
+        if (fetch_and_set_bit(visited->data(), edge, async))
```
danieldanciu, line 2661: BTW, is it ever possible that this returns true?

karasikov: yes, it could be a loop

PR #223, opened 2020-10-25
Annotate k-mer counts

karasikov:
This is a first PR, contains only construction of count annotation.

The annotation is constructed in two steps:
1. First, the usual construction of columns.
2. Then, map all counted input k-mers to the respective annotation column and write their counts to the positions given by `rank` on the column.

### DIFF
--- a/metagraph/src/annotation/representation/base/annotation.hpp
+++ b/metagraph/src/annotation/representation/base/annotation.hpp
@@ -51,6 +51,7 @@ class MultiLabelAnnotation
 
     virtual void add_labels(const std::vector<Index> &indices,
                             const VLabels &labels) = 0;
+    // for each label and index 'indices[i]' add count 'counts[i]'
     virtual void add_label_counts(const std::vector<Index> &indices,
                                   const VLabels &labels,
                                   const std::vector<uint32_t> &counts);
--- a/metagraph/src/annotation/representation/base/annotation.hpp
+++ b/metagraph/src/annotation/representation/base/annotation.hpp
@@ -51,6 +51,9 @@ class MultiLabelAnnotation
 
     virtual void add_labels(const std::vector<Index> &indices,
                             const VLabels &labels) = 0;
+    virtual void add_label_counts(const std::vector<Index> &indices,
+                                  const VLabels &labels,
+                                  const std::vector<uint32_t> &counts);
 
     virtual bool has_label(Index i, const Label &label) const = 0;
     virtual bool has_labels(Index i, const VLabels &labels) const = 0;
--- a/metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
+++ b/metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
@@ -5,6 +5,8 @@
 #include <algorithm>
 #include <stdexcept>
 
+#include <sdsl/enc_vector.hpp>
+
 #include "common/serialization.hpp"
 #include "common/utils/string_utils.hpp"
 #include "common/logger.hpp"
@@ -20,7 +22,8 @@ namespace annot {
 using utils::remove_suffix;
 using mtg::common::logger;
 
-size_t kNumElementsReservedInBitmapBuilder = 10'000'000;
+const size_t kNumElementsReservedInBitmapBuilder = 10'000'000;
+const uint8_t kCountWidth = 8;
 
 
 template <typename Label>
@@ -73,6 +76,42 @@ void ColumnCompressed<Label>::add_labels(const std::vector<Index> &indices,
 }
 
 template <typename Label>
+void ColumnCompressed<Label>::add_label_counts(const std::vector<Index> &indices,
+                                               const VLabels &labels,
+                                               const std::vector<uint32_t> &counts) {
+    assert(indices.size() == counts.size());
+
+    const auto &columns = get_matrix().data();
+
+    if (relation_counts_.size() != columns.size())
+        relation_counts_.resize(columns.size());
+
+    for (const auto &label : labels) {
+        const auto j = label_encoder_.insert_and_encode(label);
+
+        if (!relation_counts_[j].size()) {
+            relation_counts_[j] = sdsl::int_vector<>(columns[j]->num_set_bits(), 0, kCountWidth);
+
+        } else if (relation_counts_[j].size() != columns[j]->num_set_bits()) {
+            logger->error("Binary relation matrix was changed while adding relation counts");
+            exit(1);
+        }
+
+        for (size_t i = 0; i < indices.size(); ++i) {
+            if (uint64_t rank = columns[j]->conditional_rank1(indices[i])) {
+                uint64_t count = std::min(counts[i], (uint32_t)sdsl::bits::lo_set[kCountWidth]);
+                sdsl::int_vector_reference<sdsl::int_vector<>> ref = relation_counts_[j][rank - 1];
+                ref = std::min((uint64_t)ref, sdsl::bits::lo_set[kCountWidth] - count) + count;
+
+            } else {
+                logger->warn("Trying to add count {} for non-annotated object {}",
+                             counts[i], indices[i]);
+            }
+        }
+    }
+}
+
+template <typename Label>
 bool ColumnCompressed<Label>::has_label(Index i, const Label &label) const {
     try {
         return get_column(label)[i];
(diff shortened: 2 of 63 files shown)


metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
```diff
@@ -72,6 +73,42 @@ void ColumnCompressed<Label>::add_labels(const std::vector<Index> &indices,
...
+            relation_counts_[j] = sdsl::int_vector<>(columns[j]->num_set_bits(), 0, kCountWidth);
+
+        } else if (relation_counts_[j].size() != columns[j]->num_set_bits()) {
+            logger->error("Binary relation matrix was changed while adding relation counts");
+            exit(1);
+        }
+
+        for (size_t i = 0; i < indices.size(); ++i) {
+            if (uint64_t rank = columns[j]->conditional_rank1(indices[i])) {
+                uint64_t count = std::min(counts[i], (uint32_t)sdsl::bits::lo_set[kCountWidth]);
```
hmusta, line 100: since `kCountWidth` is fixed, why not store this upper bound as `kMaxCount`?

metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
```diff
@@ -72,6 +73,42 @@ void ColumnCompressed<Label>::add_labels(const std::vector<Index> &indices,
...
+                sdsl::int_vector_reference<sdsl::int_vector<>> ref = relation_counts_[j][rank - 1];
+                ref = std::min((uint64_t)ref, sdsl::bits::lo_set[kCountWidth] - count) + count;
+
+            } else {
+                logger->warn("Trying to add count {} for non-annotated object {}",
```
hmusta, line 105: Why not say something along the lines of "ignoring count" instead? Saying "trying" makes it sound like there's some attempt to get the operation to work.

metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
hmusta: also, why `uint64_t count` when we know that it can be at most 32-bits?

metagraph/src/annotation/representation/base/annotation.hpp
```diff
@@ -51,6 +51,9 @@ class MultiLabelAnnotation
 
     virtual void add_labels(const std::vector<Index> &indices,
                             const VLabels &labels) = 0;
+    virtual void add_label_counts(const std::vector<Index> &indices,
+                                  const VLabels &labels,
+                                  const std::vector<uint32_t> &counts);
```
hmusta, line 56: Is it significantly slower if you use `sdsl::int_vector<>` here with a width of 32 instead of `std::vector<uint32_t>`?

metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
```diff
@@ -72,6 +73,42 @@ void ColumnCompressed<Label>::add_labels(const std::vector<Index> &indices,
...
+        } else if (relation_counts_[j].size() != columns[j]->num_set_bits()) {
+            logger->error("Binary relation matrix was changed while adding relation counts");
+            exit(1);
+        }
+
+        for (size_t i = 0; i < indices.size(); ++i) {
+            if (uint64_t rank = columns[j]->conditional_rank1(indices[i])) {
+                uint64_t count = std::min(counts[i], (uint32_t)sdsl::bits::lo_set[kCountWidth]);
+                sdsl::int_vector_reference<sdsl::int_vector<>> ref = relation_counts_[j][rank - 1];
+                ref = std::min((uint64_t)ref, sdsl::bits::lo_set[kCountWidth] - count) + count;
```
hmusta, line 102: If the current strategy of storing counts after finishing with annotation is going to be kept, this could probably be implemented atomically using a CAS operation.

hmusta (changes requested):

metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
hmusta: as an example, for simplicity, if kCountWidth stays as 8, one can do
```
uint8_t *ref = ((uint8_t*)relation_counts_[j].data())[rank - 1];
uint8_t expected = *ref;
uint8_t desired;
do {
    desired = std::min(expected, kMaxCount - count) + count;
} while (__atomic_compare_exchange_n(ref, &expected, desired, true, memorder, __ATOMIC_RELAXED));
```

metagraph/src/graph/annotated_dbg.hpp
```diff
@@ -74,6 +74,11 @@ class AnnotatedDBG : public AnnotatedSequenceGraph {
 
     const DeBruijnGraph& get_graph() const { return dbg_; }
 
+    // add k-mer counts to the annotation
+    void add_kmer_counts(std::string_view sequence,
+                         const std::vector<Label> &labels,
+                         std::vector<uint32_t>&& kmer_counts);
```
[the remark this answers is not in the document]
karasikov: maybe in the future... Right now we always have variable counts for k-mers.
For reads, the overhead is always negligible.
So this may be useful only for very long sequences, for which you don't want to add `vector<uint32_t>` overhead.

metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
```diff
@@ -107,6 +144,44 @@ void ColumnCompressed<Label>::serialize(const std::string &filename) const {
...
+
+    if (!relation_counts_.size())
+        return;
+
+    outstream.close();
+
+    outstream.open(remove_suffix(filename, kExtension) + kExtension + ".counts",
+                   std::ios::binary);
+    if (!outstream.good())
+        throw std::ofstream::failure("Bad stream");
```
danieldanciu, line 156: I'd make this error message more useful by adding a `logger.trace("Could not open {}", filename);`
I sometimes get this "Bad stream" output in the integration tests, and it would be very useful if I knew which of the streams is actually bad.

metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
```diff
@@ -107,6 +144,44 @@ void ColumnCompressed<Label>::serialize(const std::string &filename) const {
...
+
+    uint64_t num_counts = 0;
+    uint64_t sum_counts = 0;
+
+    for (size_t j = 0; j < relation_counts_.size(); ++j) {
+        if (!relation_counts_[j].size()) {
+            sdsl::int_vector<>(bitmatrix_[j]->num_set_bits(), 0, kCountWidth).serialize(outstream);;
+
+        } else if (relation_counts_[j].size() != bitmatrix_[j]->num_set_bits()) {
```
danieldanciu, line 165: This can only happen because of a programming error, right? Wouldn't an `assert` be more appropriate?

danieldanciu (commented):
The change list looks fine so far, but seems unfinished. Did you mean to add some tests later?

metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
hmusta: we can maybe call this `atomic_fetch_and_adds`, where `adds` refers to addition with saturation

metagraph/src/annotation/representation/base/annotation.hpp
karasikov: I didn't try, but this doesn't make much sense, these vectors are limited in size, so there is no reason to pack them as much as possible.

metagraph/src/graph/annotated_dbg.cpp
```diff
@@ -67,6 +67,33 @@ ::annotate_sequence(std::string_view sequence,
...
+                                   const std::vector<Label> &labels,
+                                   std::vector<uint32_t>&& kmer_counts) {
+    assert(check_compatibility());
+    assert(kmer_counts.size() == sequence.size() - dbg_.get_k() + 1);
+
+    std::vector<row_index> indices;
+    indices.reserve(sequence.size());
+    size_t end = 0;
+
+    graph_->map_to_nodes(sequence, [&](node_index i) {
```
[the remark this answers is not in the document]
karasikov: done

metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
karasikov: done

metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
karasikov: done

metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
karasikov: Yeah, would be good to have this routine in `vector_algorithm.hpp`.
But in this specific place, this is not needed much, it takes much longer to map sequences to graph than incrementing these counts. So all threads should be busy.

metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
```diff
@@ -20,7 +20,8 @@ namespace annot {
 using utils::remove_suffix;
 using mtg::common::logger;
 
-size_t kNumElementsReservedInBitmapBuilder = 10'000'000;
+const size_t kNumElementsReservedInBitmapBuilder = 10'000'000;
+const uint8_t kCountWidth = 8;
```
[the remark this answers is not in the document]
karasikov: I read `kCountBits` as "bits per count".
`kCountWidth` makes a lot of sense too, so I think it should be either `kCountWidth` or `kCountBits`, but not mixed `kCountWidthBits`, which could be interpreted as "bits per count width".

Will change `kCountBits` to `kCountWidth`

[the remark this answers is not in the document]
karasikov: Ah, wait, it's already `kCountWidth`.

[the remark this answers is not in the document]
karasikov: Will rename it to `kCountBits` then.

metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
karasikov: Ideally, we should go through all such error handlers and fix them all, maybe appending the filename to the exception message instead of writing to the logger.

metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
karasikov: done

metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
hmusta: I started a new branch to add this function to vector_algorithm. I start a PR once it finishes compiling for me.

PR #225, opened 2020-10-26
Differential assembly support for canonical and primary graphs

hmusta:
This is a rebase of #188.


metagraph/src/cli/load/load_annotated_graph.cpp
```diff
@@ -22,7 +22,7 @@ using mtg::common::logger;
 std::unique_ptr<AnnotatedDBG> initialize_annotated_dbg(std::shared_ptr<DeBruijnGraph> graph,
                                                        const Config &config) {
     // TODO: check and wrap into canonical only if the graph is primary
-    if (config.canonical && !graph->is_canonical_mode())
+    if (config.canonical && !graph->is_canonical_mode() && config.identity != Config::ASSEMBLE)
```
karasikov, line 25: why not for assembly?

Normally you just don't pass `--canonical` to assembly, and this condition won't be satisfied anyway, right?

metagraph/src/common/vectors/bitmap.hpp
```diff
@@ -187,4 +187,40 @@ class bitmap_lazy : public bitmap {
...
+
+    bool operator[](uint64_t) const { throw std::runtime_error("Not implemented"); }
+    uint64_t get_int(uint64_t, uint32_t) const { throw std::runtime_error("Not implemented"); }
+
+    uint64_t size() const { return size_; }
+    uint64_t num_set_bits() const { return num_set_bits_; }
+
+    void add_to(sdsl::bit_vector *other) const;
+    void call_ones_in_range(uint64_t begin, uint64_t end,
+                            const VoidCall<uint64_t> &callback) const;
```
karasikov, line 218: add a comment that this function is inefficient and should not be called

karasikov: maybe also throw an exception unless end != size_ and begin != 0?

metagraph/src/common/vectors/bitmap.cpp
```diff
@@ -369,3 +369,52 @@ void bitmap_lazy::call_ones_in_range(uint64_t begin, uint64_t end,
...
+
+bitmap_generator
+::bitmap_generator(std::function<void(const VoidCall<uint64_t>&)>&& generator,
+                   size_t size,
+                   size_t num_set_bits) noexcept
+      : size_(size), num_set_bits_(num_set_bits), generator_(std::move(generator)) {}
+
+bitmap_generator
+::bitmap_generator(bitmap&& base, std::function<uint64_t(uint64_t)>&& index_transformer,
+                   size_t size, size_t num_set_bits) noexcept
```
karasikov, line 395: Hard to read. Make each on a new line
```suggestion
::bitmap_generator(bitmap&& base,
                   std::function<uint64_t(uint64_t)>&& index_transformer,
                   size_t size,
                   size_t num_set_bits) noexcept
```

metagraph/src/common/vectors/bitmap.hpp
```diff
@@ -187,4 +187,40 @@ class bitmap_lazy : public bitmap {
...
+
+    bitmap_generator(std::function<void(const VoidCall<uint64_t>&)>&& generator,
+                     size_t size,
+                     size_t num_set_bits = -1) noexcept;
+
+    bitmap_generator(bitmap&& base,
+                     std::function<uint64_t(uint64_t)>&& index_transformer
+                         = [](auto i) { return i; },
+                     size_t size = -1,
+                     size_t num_set_bits = -1) noexcept;
```
karasikov, line 208: I would remove this constructor, it looks too special

metagraph/src/common/vectors/bitmap.cpp
```diff
@@ -369,3 +369,52 @@ void bitmap_lazy::call_ones_in_range(uint64_t begin, uint64_t end,
...
+////////////////////////////////////////////////////////////////
+
+bitmap_generator::bitmap_generator(size_t size, bool value)
+      : bitmap_generator([value,size](const auto &index_callback) {
+                             if (value) {
+                                 for (size_t i = 0; i < size; ++i) {
+                                     index_callback(i);
+                                 }
+                             }
+                         }, size, size * value) {}
```
karasikov, line 385:
```suggestion
                         },
                         size, size * value) {}
```

metagraph/src/graph/representation/masked_graph.hpp
```diff
@@ -119,6 +119,7 @@ class MaskedDeBruijnGraph : public DeBruijnGraph {
     virtual void set_mask(bitmap *mask) { kmers_in_graph_.reset(mask); }
 
     virtual const bitmap& get_mask() const { return *kmers_in_graph_; }
+    virtual std::unique_ptr<bitmap>& get_mask_ptr() { return kmers_in_graph_; }
```
karasikov, line 122: Wouldn't it be better to construct a new masked graph, instead of doing these hacks with modifying an existing mask? (you can re-use the same graph, so the overhead will be zero).

If you modify the mask, it may break the internal properties of the graph (currently only `only_valid_nodes_in_mask_`).

metagraph/tests/annotation/test_annotated_dbg_helpers.cpp
```diff
@@ -12,14 +13,16 @@ namespace test {
...
+                                               const std::vector<std::string> &labels,
+                                               DBGMode mode) {
     assert(sequences.size() == labels.size());
-    auto graph = build_graph_batch<Graph>(k, sequences);
+    auto graph = build_graph_batch<Graph>(k, sequences, mode);
+    auto canonical = std::dynamic_pointer_cast<CanonicalDBG>(graph);
+    assert(!canonical || mode == DBGMode::CANONICAL_WRAPPER);
 
-    uint64_t max_index = graph->max_index();
+    uint64_t max_index = canonical ? canonical->get_graph().max_index() : graph->max_index();
```
karasikov, line 25:
```suggestion
    uint64_t max_index = canonical
                            ? canonical->get_graph().max_index()
                            : graph->max_index();
```

metagraph/src/annotation/binary_matrix/base/binary_matrix.hpp
```diff
@@ -21,6 +23,7 @@ class BinaryMatrix {
     typedef Vector<Column> SetBitPositions;
     typedef std::function<void(const SetBitPositions &)> RowCallback;
     typedef std::function<void(Row, Column)> ValueCallback;
+    typedef std::function<void(Column, bitmap&&)> ColumnCallback;
```
karasikov, line 26:
```suggestion
```

metagraph/src/annotation/binary_matrix/base/binary_matrix.hpp
```diff
@@ -32,9 +35,13 @@ class BinaryMatrix {
     virtual SetBitPositions get_row(Row row) const = 0;
     virtual std::vector<SetBitPositions> get_rows(const std::vector<Row> &rows) const;
     virtual std::vector<Row> get_column(Column column) const = 0;
+
     // get all selected rows appended with -1 and concatenated
     virtual std::vector<Column> slice_rows(const std::vector<Row> &rows) const;
 
+    virtual void slice_columns(const std::vector<Column> &columns,
+                               const ColumnCallback &callback) const;
```
karasikov, line 43:
```suggestion
    virtual void slice_columns(const std::vector<Column> &columns,
                               const std::function<void(Column, bitmap&&)> &callback) const;
```

karasikov: I assume this should be parallelized? (otherwise why would you call the indexes `Column`).

So, pass num_threads as another argument too?

karasikov (changes requested):

metagraph/src/cli/load/load_annotated_graph.cpp
hmusta: True, but metagraph will still run if there's a stray `--canonical` flag in the `ASSEMBLE` regime.

Maybe instead we can have metagraph print a warning in `config.cpp` if this case happens to warn that it'll be slower and add extra k-mers to the output.

hmusta: or we could print the warning here

metagraph/src/graph/representation/masked_graph.hpp
hmusta: But there is overhead since you have to create a new bit vector to populate

karasikov: This is an abstract bitmap, which is not necessarily modifiable, anyway.
Copying a bitmap is usually acceptable.

metagraph/src/cli/load/load_annotated_graph.cpp
karasikov: better unset it in config.cpp, so it's always false for ASSEMBLY.
Or just leave as it was. `--canonical` is not listed in usage for ASSEMBLY, so if pass it, the behavior is undefined.

metagraph/src/common/vectors/bitmap.hpp
hmusta: this is not used in this PR anyway, so I removed it

metagraph/src/cli/config/config.cpp
```diff
@@ -995,13 +996,17 @@ void Config::print_usage(const std::string &prog_name, IdentityType identity) {
...
             fprintf(stderr, "\n");
-            fprintf(stderr, "\t-a --annotator [STR] \t\t\tannotator to load []\n");
-            fprintf(stderr, "\t   --label-mask-in [STR] \t\tlabel to include in masked graph\n");
-            fprintf(stderr, "\t   --label-mask-out [STR] \t\tlabel to exclude from masked graph\n");
-            fprintf(stderr, "\t   --label-mask-in-fraction [FLOAT] \tminimum fraction of mask-in labels among the set of masked labels [1.0]\n");
-            fprintf(stderr, "\t   --label-mask-out-fraction [FLOAT] \tmaximum fraction of mask-out labels among the set of masked labels [0.0]\n");
-            fprintf(stderr, "\t   --label-other-fraction [FLOAT] \tmaximum fraction of other labels allowed [1.0]\n");
-            fprintf(stderr, "\t   --filter-by-kmer \t\t\tmask out graph k-mers individually [off]\n");
+            fprintf(stderr, "\t-a --annotator [STR] \t\tannotator to load []\n");
+            fprintf(stderr, "\t-c --canonical \t\t\ttreat the input graph as a canonical graph [off]\n");
```
karasikov, line 1000: Flag `--canonical` is now redundant

metagraph/src/graph/annotated_graph_algorithm.cpp
```diff
@@ -1,252 +1,505 @@
...
-                set_bit(unitig_mask.data(), node, atomic);
+    logger->trace("Filtering by unitig");
+
+    size_t num_labels = anno_graph.get_annotation().num_labels();
+    bool check_other = config.label_mask_other_unitig_fraction != 1.0;
+    counts.width(width);
+
+    update_masked_graph_by_unitig(*masked_graph,
+                                  [&](const auto &unitig, const auto &path)
+                                      -> std::vector<std::pair<size_t, size_t>> {
```
karasikov, line 138: This lambda is huge, so better define a function first and then call `update_masked_graph_by_unitig`.

```c++
auto ... = [&](...) {

};

update_masked_graph_by_unitig(*masked_graph, ...);

File: metagraph/src/annotation/binary_matrix/base/binary_matrix.cpp [hunk 12]
```diff
@@ -32,6 +34,17 @@ BinaryMatrix::slice_rows(const std::vector<Row> &row_ids) const {
...
 
+void BinaryMatrix::slice_columns(const std::vector<Column> &column_ids,
+                                 const std::function<void(Column, bitmap&&)> &callback,
+                                 size_t /* num_threads */) const {
+    size_t nrows = num_rows();
+
+    for (size_t k = 0; k < column_ids.size(); ++k) {
+        Column j = column_ids[k];
+        callback(j, bitmap_generator(get_column(j), nrows));
+    }
```
karasikov, line 45: why not enable `num_threads`?
```suggestion
                                 size_t num_threads) const {
    #pragma omp parallel for num_threads(num_threads) schedule(dynamic)
    for (size_t k = 0; k < column_ids.size(); ++k) {
        Column j = column_ids[k];
        callback(j, bitmap_generator(get_column(j), num_rows()));
    }
```

metagraph/src/cli/assemble.cpp
```diff
@@ -120,15 +224,56 @@ int assemble(Config *config) {
...
+                                       config->kmers_in_single_form);
+                } else {
+                    graph.call_sequences([&](const auto &seq, auto&&) {
+                                             std::lock_guard<std::mutex> lock(write_mutex);
+                                             writer.write(seq);
+                                         },
+                                         num_threads_per_traversal,
+                                         config->kmers_in_single_form);
+                }
+            },
```
karasikov, line 271: The is written as a critical section. So you run only one call_sequences at a time? If so, why do you use only `num_threads_per_traversal` threads and not all `get_num_threads()`?

What is the parallelization scheme here?

karasikov (changes requested):

metagraph/tests/annotation/test_annotated_dbg.cpp
```diff
@@ -9,6 +9,18 @@
...
+#include "graph/representation/hash/dbg_hash_string.hpp"
+#include "graph/representation/hash/dbg_hash_ordered.hpp"
+#include "graph/representation/hash/dbg_hash_fast.hpp"
+#include "graph/representation/bitmap/dbg_bitmap.hpp"
+
+#include "annotation/representation/column_compressed/annotate_column_compressed.hpp"
+
+// this next #include includes AnnotatedDBG. we need access to its protected
+// members to modify the underlying annotator
+#define protected public
```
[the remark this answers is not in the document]
hmusta: in `ExtendGraphWithSimplePath`, for example, `annotator_` is accessed to insert rows after the graph is modified

metagraph/tests/annotation/test_annotated_dbg_helpers.cpp
```diff
@@ -2,65 +2,68 @@
...
-    uint64_t max_index = canonical
-                            ? canonical->get_graph().max_index()
-                            : graph->max_index();
+    uint64_t max_index = canonical ? canonical->get_graph().max_index()
+                                   : graph->max_index();
 
     auto anno_graph = std::make_unique<AnnotatedDBG>(
         graph,
-        std::make_unique<annot::ColumnCompressed<>>(max_index)
+        std::make_unique<ColumnCompressed<>>(max_index)
```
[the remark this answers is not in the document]
hmusta: `AnnotatedDBG::check_compatibility` uses `max_index` from the underlying graph, so we need to set the size this way

metagraph/src/cli/assemble.cpp
hmusta: The goal in the end is to have several assembly experiments running in parallel and have them all write their results to the same output file.

One approach is to just have one output file per experiment.

The other approach (keeping the single-file output) would involve either modifying FastaWriter to have a header associated with each sequence, or do the current approach where each assembly that finishes locks the output files to append.

metagraph/src/annotation/binary_matrix/base/binary_matrix.cpp
hmusta: The problem here is that `get_column` can return a huge vector, so if a large number of threads is passed, this can easily blow up the RAM usage.

metagraph/tests/annotation/test_annotated_dbg_helpers.cpp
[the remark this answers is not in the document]
karasikov: Then let's fix it there, so the wrapper returns the same max_index as its base graph.

metagraph/src/annotation/binary_matrix/base/binary_matrix.cpp
karasikov: You can always set num_threads to 1 in the caller if memory is an issue.

metagraph/src/cli/assemble.cpp
karasikov: Are you sure we need these two levels of parallelization?

Why do we need to run multiple assemblies at the same time, if each assembly already happens in parallel?

karasikov (changes requested):
Could you also add some integration tests?

metagraph/src/cli/assemble.cpp
hmusta: Since different assembly experiments are independent, it scales better than doing one with multiple threads

hmusta: Though this is theoretical, I don't remember if I actually measured it to see if it makes a difference

metagraph/tests/annotation/test_annotated_dbg_helpers.cpp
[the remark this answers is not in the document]
hmusta: Having access to `max_index` is still needed for masked graphs (in case you use to mask a CanonicalDBG), so maybe it would be better to implement either `max_canonical_index`, or `get_base_graph` so one could call `get_base_graph().max_index()`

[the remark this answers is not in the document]
hmusta: I'd say that latter is more useful, since it provides a nice way to check the underlying type of a wrapped graph without having to guess how to unwrap it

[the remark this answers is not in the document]
karasikov: do we actually allow masking wrapped graphs? Doesn't it make more sense to mask the original primary graphs, and wrap the result when required?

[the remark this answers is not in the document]
hmusta: Even though we technically can mask a CanonicalDBG, in practice, the code checks for this case and unwraps the graph.

metagraph/src/cli/assemble.cpp
hmusta: Thinking about this again, it's probably not worth it. If two assemblies are done at the same time, then you have to lock and you end up using far fewer threads than you intended

metagraph/tests/annotation/test_annotated_dbg_helpers.cpp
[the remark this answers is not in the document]
hmusta: I added `DeBruijnGraph::get_base_graph()`. This should simplify things quite a bit.

metagraph/src/annotation/binary_matrix/column_sparse/column_major.hpp
```diff
@@ -27,6 +27,10 @@ class ColumnMajor : public BinaryMatrix {
     // get all selected rows appended with -1 and concatenated
     std::vector<Column> slice_rows(const std::vector<Row> &rows) const override;
 
+    void slice_columns(const std::vector<Column> &columns,
+                       const std::function<void(Column, bitmap&&)> &callback,
+                       size_t num_threads = 1) const override;
```
karasikov, line 32: Name `call_columns` would suit it better.

metagraph/src/annotation/binary_matrix/row_diff/row_diff.hpp
```diff
@@ -194,7 +196,7 @@ RowDiff<BaseMatrix>::get_rows(const std::vector<Row> &row_ids) const {
         graph::boss::BOSS::edge_index boss_edge = graph_->kmer_to_boss_index(
                 graph::AnnotatedSequenceGraph::anno_to_graph_index(row));
 
-        while (true) {
+        while (boss_edge) {
```
karasikov, line 199: Why change this?

metagraph/src/annotation/binary_matrix/row_diff/row_diff.hpp
```diff
@@ -154,6 +154,8 @@ BinaryMatrix::SetBitPositions RowDiff<BaseMatrix>::get_row(Row row) const {
 
     while (!anchor_[row]) {
         boss_edge = boss.row_diff_successor(boss_edge, rd_succ);
+        if (!boss_edge)
+            return result;
```
karasikov, line 158: This must never happen, why did you need to add this?

metagraph/src/cli/assemble.cpp
```diff
@@ -17,64 +21,194 @@ using mtg::common::logger;
...
+            detected_missing_labels = true;
+            logger->trace("Label {} is not found in annotation", label);
+        }
+    }
+
+    if (detected_missing_labels)
+        exit(1);
+
+    std::sort(label_set.begin(), label_set.end());
+}
```
karasikov, line 41: Maybe rename to `check_and_sort_labels`?

metagraph/src/cli/assemble.cpp
```diff
@@ -17,64 +21,194 @@ using mtg::common::logger;
...
+std::pair<std::vector<std::string>, unsigned int> parse_diff_file(std::ifstream &fin) {
+    std::string line;
+    std::vector<std::string> lines;
+    unsigned int num_experiments = 0;
+    while (std::getline(fin, line)) {
+        if (line.empty() || line[0] == '#')
+            continue;
+
+        num_experiments += (line[0] != '@');
+        lines.emplace_back(std::move(line));
```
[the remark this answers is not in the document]
karasikov: Can you add commends and describe what @ and lines represent?

karasikov (changes requested):

metagraph/src/annotation/binary_matrix/row_diff/row_diff.hpp
hmusta: If row ids corresponding to dummy sink k-mers are passed, then `row_diff_successor` crashes. For the other annotation representations, these indices just return empty annotations, so I wanted to have that same behaviour here.

metagraph/src/cli/assemble.cpp
```diff
@@ -93,15 +227,62 @@ int assemble(Config *config) {
...
+        call_masked_graphs(*anno_graph, config, lines,
+            [&](const graph::MaskedDeBruijnGraph &graph, const std::string &header) {
+                seq_io::FastaWriter writer(config->outfbase, header,
+                                           config->enumerate_out_sequences,
+                                           num_threads > 1, /* async write */
+                                           "a" /* append mode */);
+
+                if (config->unitigs || config->min_tip_size > 1) {
+                    graph.call_unitigs([&](const auto &unitig, auto&&) {
+                                           std::lock_guard<std::mutex> lock(write_mutex);
```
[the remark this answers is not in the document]
hmusta: Yes. async mode simply means that writes to disk happen in a separate thread so pushes to the buffer can continue to happen concurrently. The lock here is needed since pushes are not thread safe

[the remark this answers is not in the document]
karasikov: Ah, got it. It's a different kind of parallelism. Maybe calling `async_writer` instead of the simple `async` would help to avoid the confusion.

metagraph/src/annotation/binary_matrix/row_diff/row_diff.hpp
karasikov: Hm, it was designed so that you never call it for not existing/dummy nodes.

Why do you need to do that?

hmusta: There are two cases where this can come up
1) slice_columns (to be renamed call_columns) for now iterates through each row index and checks it. This one is easy to fix since we can check for dummy sinks while iterating.
2) call_outgoing_kmers and adjacent_outgoing_nodes in DBGSuccinct returns dummy sinks, so if these methods are used to populate a row index vector, then it'll include dummy sinks. I don't remember if there's a reason why we kept it like this.

karasikov: Both of these would be incredibly inefficient to call with the current implementation.

I suggest better throw error on those calls and not change this implementation.

hmusta: The `call_outgoing_kmers` approach is used in alignment for traversal (after the alignment is done, all traversed nodes can be queried as a batch). In basically all cases where this can occur, checks can be put in other parts of the code to prevent it from propagating here.

In the end, not fixing this just leaves a conceptual issue: a row index corresponding to a dummy k-mer is still a technically valid input, but it causes this function to crash.

metagraph/src/cli/assemble.cpp
```diff
@@ -149,6 +149,9 @@ void call_masked_graphs(const AnnotatedDBG &anno_graph,
         auto line_split = utils::split_string(line, "\t", false);
 
         if (line[0] == '@') {
+            // lines of the form '@\tIN1,IN2,IN3\tOUT1,OUT2,OUT3' designate
```
karasikov, line 152: What if comma is part of a label name?

What about this format:
```
@ Experiment 1: ...
IN: LABEL1
IN: LABEL2
...
OUT: ...
...
@ Experiment 2: ...
# other comments
IN: ...
```

metagraph/src/annotation/binary_matrix/row_diff/row_diff.hpp
karasikov: Ok, but you never want to align to dummy nodes. So you should check and skip them. Then you can extract annotations for the batch of actual real nodes.

> In the end, not fixing this just leaves a conceptual issue: a row index corresponding to a dummy k-mer is still a technically valid input, but it causes this function to crash.

The dummy nodes are valid for BOSS, but not on the higher abstraction levels (SuccinctDBG/DeBruijnGraph), where they are as invalid as any negative index. So having UB or throwing an error for that looks ok to me, there are no conceptual issues.

karasikov (changes requested):

metagraph/src/cli/assemble.cpp
hmusta: I think the best way to avoid all of these issues is to pass in a JSON file to define the experiments. This way, there's no ambiguity in parsing and we can just use the library to read it.

karasikov: But that is less human-readable/writable.

hmusta: It can be just as readable. For example
```json
{
    "shared_labels": [
        {
            "start_index": 0,
            "in": ["IN1","IN2","IN3"],
            "out": ["OUT1","OUT2"]
        }
    ],
    "experiments": [
        {
            "name": "fuzzy",
            "in": ["myIN1","myIN2","myIN3"],
            "out": ["myOUT1","myOUT2"],
            "in_min_fraction": 0.6,
            "out_max_fraction": 0.5,
            "other_max_fraction": 0.0,
            "unitig_in_min_fraction": 1.0,
            "unitig_out_max_fraction": 0.0
        },
        {
            "name": "or",
            "in": ["myIN1","myIN2"],
            "out": ["myOUT1","myOUT2"],
            "in_min_fraction": 0.5,
            "out_max_fraction": 0.0,
            "other_max_fraction": 0.0,
            "unitig_in_min_fraction": 1.0,
            "unitig_out_max_fraction": 0.0
        }
    ]
}
```

metagraph/integration_tests/test_assemble.py
```diff
@@ -195,3 +200,103 @@ def test_round_robin_graph_size_via_gfa(self, gfa_test):
...
+        construct_command = '{exe} build {mask_dummy} -p {num_threads} \
+                --graph {repr} -k {k} -o {outfile} {input}'.format(
+            exe=METAGRAPH,
+            num_threads=NUM_THREADS,
+            mask_dummy='--mask-dummy' if cls.mask_dummy else '',
+            repr=cls.graph_repr,
+            k=k,
+            outfile=cls.tempdir.name + '/graph',
+            input=TEST_DATA_DIR + '/metasub_fake_data.fa'
+        )
```
karasikov, line 238: use f strings

metagraph/src/annotation/binary_matrix/column_sparse/column_major.cpp
```diff
@@ -76,6 +76,17 @@ ColumnMajor::slice_rows(const std::vector<Row> &row_ids) const {
...
+void ColumnMajor::call_columns(const std::vector<Column> &columns,
+                               const std::function<void(Column, bitmap&&)> &callback,
+                               size_t num_threads) const {
+    #pragma omp parallel for num_threads(num_threads) schedule(dynamic)
+    for (size_t j = 0; j < columns.size(); ++j) {
+        const bitmap *c = (*columns_)[columns[j]].get();
+        callback(j, bitmap_generator([c](const auto &call_one) { c->call_ones(call_one); },
+                                     c->size(), c->num_set_bits()));
+    }
+}
```
karasikov, line 88: The method is const, but you call an rvalue, hence you must either copy or create a wrapper on every call.
Making it const reference `const bitmap &` will be better: you can still copy or wrap it in the caller, but in addition to this, you can call the column directly, as would be useful in this case:
```suggestion
void ColumnMajor::call_columns(const std::vector<Column> &columns,
                               const std::function<void(Column, const bitmap &)> &callback,
                               size_t num_threads) const {
    #pragma omp parallel for num_threads(num_threads) schedule(dynamic)
    for (size_t j = 0; j < columns.size(); ++j) {
        callback(j, *(*columns_)[columns[j]]);
    }
}
```

metagraph/src/annotation/binary_matrix/row_diff/row_diff.hpp
```diff
@@ -127,10 +127,16 @@ std::vector<BinaryMatrix::Row> RowDiff<BaseMatrix>::get_column(Column column) co
     assert(anchor_.size() == diffs_.num_rows() && "anchors must be loaded");
     assert(!fork_succ_.size() || fork_succ_.size() == graph_->get_boss().get_last().size());
 
-    // TODO: implement a more efficient algorithm
+    const auto &boss = graph_->get_boss();
```
karasikov, line 130:
```suggestion
    const BOSS &boss = graph_->get_boss();
```

metagraph/src/cli/assemble.cpp
karasikov: Yeah, readability looks ok but it looks way harder to write.

metagraph/src/cli/assemble.cpp
```diff
@@ -17,64 +21,205 @@ using mtg::common::logger;
...
+                 line_split[0]);
+    }
+}
+
+std::pair<std::vector<std::string>, unsigned int>
+parse_diff_file(const std::string &fname) {
+    std::ifstream fin(fname);
+    if (!fin.good()) {
+        throw std::iostream::failure("Failed to read label mask file");
+        exit(1);
```
karasikov, line 207: exit(1) will never be called. Remove?

karasikov (changes requested):

metagraph/src/cli/assemble.cpp
karasikov: But maybe it's fine, if you also add an example file to the repo, so there is something to start with.

metagraph/integration_tests/test_assemble.py
hmusta: There are a lot of places in the integration tests where f strings can be used instead of `format`, so let's save that for a later refactor

metagraph/src/graph/annotated_graph_algorithm.cpp
```diff
@@ -1,264 +1,495 @@
...
+std::pair<sdsl::int_vector<>, std::unique_ptr<bitmap>>
+construct_diff_label_count_vector(const AnnotatedDBG &anno_graph,
+                                  const std::vector<Label> &labels_in,
+                                  const std::vector<Label> &labels_out,
+                                  size_t num_threads);
+
+
+template <class GetKeptIntervals>
+void update_masked_graph_by_unitig(MaskedDeBruijnGraph &masked_graph,
+                                   const GetKeptIntervals &get_kept_intervals,
```
karasikov, line 48: This is confusing, I had to check out the code and do some search to find the type of `GetKeptIntervals`.
Please add a command explaining this.
Also, maybe pick a better name.
Why is it "kept" and not "masked"?
Why "interval"? Is this an interval of nodes?

metagraph/src/graph/annotated_graph_algorithm.cpp
```diff
@@ -1,264 +1,495 @@
...
+                          bool add_complement,
+                          bool make_boss,
+                          size_t num_threads);
+
+
+MaskedDeBruijnGraph mask_nodes_by_label(const AnnotatedDBG &anno_graph,
+                                        const std::vector<Label> &labels_in,
+                                        const std::vector<Label> &labels_out,
+                                        const std::vector<Label> &labels_in_post,
+                                        const std::vector<Label> &labels_out_post,
```
karasikov, line 69: add a comment to explain what "_post" means

metagraph/src/graph/annotated_graph_algorithm.cpp
```diff
@@ -1,264 +1,495 @@
...
+        // construct a new subgraph
+        logger->trace("Constructing BOSS from labeled subgraph");
+        std::vector<std::pair<std::string, sdsl::int_vector<>>> contigs;
+        std::mutex add_mutex;
+        BOSSConstructor constructor(
+            graph_ptr->get_k() - 1,
+            masked_canonical,
+            0 /* count width */, "" /* suffix */, 1 /* num_threads */,
+            masked_graph->num_nodes() * 32
+        );
```
karasikov, line 255: Why do you need to rebuild a boss graph?
What is the workflow? (add a comment above to describe what this function `make_initial_masked_graph` is doing)
Can't this be done separately in the `build` mode?

karasikov (changes requested):

metagraph/src/graph/annotated_graph_algorithm.cpp
hmusta: I reworked the process so it no longer needs a new BOSS graph to be built.

metagraph/src/graph/annotated_graph_algorithm.cpp
hmusta: I added a bunch of comments to describe these.

metagraph/src/graph/annotated_graph_algorithm.cpp
hmusta: I renamed `_post` to `_round2` and provided an explanation in the header for what this exactly means.

metagraph/src/cli/config/config.cpp
```diff
@@ -1022,13 +1018,9 @@ void Config::print_usage(const std::string &prog_name, IdentityType identity) {
...
             fprintf(stderr, "\n");
-            fprintf(stderr, "\t-a --annotator [STR] \t\t\tannotator to load []\n");
-            fprintf(stderr, "\t   --label-mask-in [STR] \t\tlabel to include in masked graph\n");
-            fprintf(stderr, "\t   --label-mask-out [STR] \t\tlabel to exclude from masked graph\n");
-            fprintf(stderr, "\t   --label-mask-in-fraction [FLOAT] \tminimum fraction of mask-in labels among the set of masked labels [1.0]\n");
-            fprintf(stderr, "\t   --label-mask-out-fraction [FLOAT] \tmaximum fraction of mask-out labels among the set of masked labels [0.0]\n");
-            fprintf(stderr, "\t   --label-other-fraction [FLOAT] \tmaximum fraction of other labels allowed [1.0]\n");
-            fprintf(stderr, "\t   --filter-by-kmer \t\t\tmask out graph k-mers individually [off]\n");
+            fprintf(stderr, "\t-a --annotator [STR] \t\tannotator to load []\n");
+            fprintf(stderr, "\t   --label-mask-file [STR] \tJSON file describing labels to mask in and out and their relative fractions []\n");
```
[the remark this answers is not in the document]
hmusta: I set it to `--diff-assembly-rules` to not confuse those who just want to do simple sequence extraction

metagraph/integration_tests/test_assemble.py
hmusta: I looked into this, f-strings only work for variable substitution and don't allow you to compute values within the braces, so I think it's cleaner the way it is now.

README.md
```diff
@@ -109,12 +109,12 @@ Requires `M*V/8 + Size(BRWT)` bytes of RAM, where `M` is the number of rows in t
...
                         --unitigs \
                         -a <GRAPH_DIR>/annotation.column.annodbg \
-                        --label-mask-in LABEL_1 \
-                        --label-mask-in LABEL_2 \
-                        --label-mask-out LABEL_3 \
+                        --label-mask-file diff_assembly_experiment_file.json \
                         -o diff_assembled.fa
 ```
 
+See `metagraph/tests/data/example.diff.json` for a sample file.
```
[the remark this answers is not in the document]
hmusta: I pushed a second, simpler sample config.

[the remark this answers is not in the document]
hmusta: I still want to keep the more complex one to more thoroughly describe the supported features.

metagraph/tests/annotation/test_matrix_helpers.cpp
```diff
@@ -209,6 +209,49 @@ void test_matrix(const TypeParam &matrix, const BitVectorPtrArray &columns) {
         }
     }
 
+    // check call_columns
+    for (size_t m : { size_t(0),
+                      size_t(matrix.num_columns() / 2),
+                      size_t(matrix.num_columns()) }) {
+        std::vector<uint64_t> indices(m);
+        std::iota(indices.begin(), indices.end(), 0);
```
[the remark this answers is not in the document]
hmusta: It wasn't that much of an increase.
Before: `170 tests from 10 test suites ran. (35979 ms total)`
After: `170 tests from 10 test suites ran. (39394 ms total)`

metagraph/tests/data/example.diff.json
```diff
@@ -0,0 +1,64 @@
...
+                        "A k-mer is an in-k-mer if it has at least in_min_fraction",
+                        "(default value 1.0) in-labels.",
+                        "A k-mer is an out-k-mer if it has more than out_max_fraction",
+                        "(default value 0.0) out-labels.",
+                    ],
+                    "name": "metasub_by_kmer",
+                    "in_min_fraction": 1.0,
+                    "out_max_fraction": 0.1,
+                    "in": ["kl_sample;metasub_name=nan;city=kuala_lumpur;latitude=3.11318;longitude=101.68186;surface_material=metal",
+                           "zh_sample;metasub_name=nan;city=zurich;latitude=nan;longitude=nan;surface_material=nan;station=nan;num_reads=1303795.0"],
```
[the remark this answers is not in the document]
hmusta: I made a new copy of the input file with simplified headers.

metagraph/src/graph/annotated_graph_algorithm.cpp
```diff
@@ -15,249 +12,417 @@ namespace graph {
...
+            for (const auto &[label, sig] : anno_graph.get_top_label_signatures(contig, num_labels)) {
+                if (labels_in.count(label) || labels_out.count(label))
+                    continue;
+
+                bool found_in_round2 = labels_in_round2.count(label);
+                bool found_out_round2 = labels_out_round2.count(label);
+                if (!found_in_round2 && !found_out_round2) {
+                    if (check_other) {
+                        mask_or(other_mask, sig, path);
+                    }
```
karasikov, line 146:
```suggestion
                    if (check_other)
                        mask_or(other_mask, sig, path);
```

metagraph/src/graph/annotated_graph_algorithm.cpp
```diff
@@ -15,249 +12,417 @@ namespace graph {
...
+            std::vector<DeBruijnGraph::node_index> rc_path = path;
+            reverse_complement_seq_path(*graph_ptr, rc_seq, rc_path);
+
+            auto it = rc_path.rbegin();
+            for (size_t i = 0; i < path.size(); ++i, ++it) {
+                if (*it) {
+                    uint64_t in_count = atomic_fetch(counts, path[i] * 2, vector_backup_mutex, MEM_ORDER);
+                    uint64_t out_count = atomic_fetch(counts, path[i] * 2 + 1, vector_backup_mutex, MEM_ORDER);
+                    atomic_fetch_and_add(counts, *it * 2, in_count, vector_backup_mutex, MEM_ORDER);
+                    atomic_fetch_and_add(counts, *it * 2 + 1, out_count, vector_backup_mutex, MEM_ORDER);
```
karasikov, line 288: Why do you need to pass a mutex? Isn't it always allowed to do atomic_fetch and add with any memory order?

karasikov (changes requested):

metagraph/src/graph/annotated_graph_algorithm.cpp
hmusta: Atomic operations on unaligned blocks of size > 64 are not allowed. So if the int_vector width is too large, this defaults to doing all operations with a lock.

metagraph/src/graph/annotated_graph_algorithm.cpp
```diff
@@ -15,249 +12,417 @@ namespace graph {
...
+ * Return an int_vector<>, bit_vector pair, each of length anno_graph.get_graph().max_index().
+ * For an index i, the int_vector will contain a packed integer representing the
+ * number of labels in labels_in and labels_out which the k-mer of index i is
+ * annotated with. The least significant half of each integer represents the count
+ * from labels_in, while the most significant half represents the count from
+ * labels_out. The width of the int_vector<> is computed to be wide enough to contain
+ * counts up to num_labels.
+ * The returned bit_vector is a binarization of the int_vector
+ */
+std::pair<sdsl::int_vector<>, sdsl::bit_vector>
```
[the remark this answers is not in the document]
hmusta: The description is a leftover from a previous iteration where the int_vector was the same length as the bit_vector, but double-width (now, it's double the length of the bit_vector).

So I guess instead we can say that the bit_vector is a mask indicating all k-mers which are annotated with an in- or out-label

metagraph/integration_tests/test_assemble.py
```diff
@@ -195,3 +200,132 @@ def test_round_robin_graph_size_via_gfa(self, gfa_test):
...
+        assemble_command = '{exe} assemble -p {num_threads} \
+                -a {annotation} -o {outfile} \
+                --diff-assembly-rules {mask} {graph}'.format(
+            exe=METAGRAPH,
+            num_threads=NUM_THREADS,
+            outfile=self.tempdir.name + '/diff_contigs',
+            graph=self.tempdir.name + '/graph' + graph_file_extension[self.graph_repr],
+            annotation=self.tempdir.name + '/annotation' + anno_file_extension[self.anno_repr],
+            mask=TEST_DATA_DIR + '/example.diff.json'
+        )
```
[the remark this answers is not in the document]
hmusta: f-strings don't work as well here since the variable values are not precomputed. If I did it that way, I'd have to define a bunch of these variables in the lines above, so it wouldn't look as nice

metagraph/tests/data/example_simple.diff.json
```diff
@@ -0,0 +1,23 @@
...
+            "experiments": [
+                {
+                    "__comment__": [
+                        "This is an example of the k-mer-based mode.",
+                        "A k-mer is an in-k-mer if it has at least in_min_fraction",
+                        "(default value 1.0) in-labels.",
+                        "A k-mer is an out-k-mer if it has more than out_max_fraction",
+                        "(default value 0.0) out-labels.",
+                    ],
+                    "name": "metasub_by_kmer",
```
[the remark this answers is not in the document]
hmusta: The corresponding integration test uses the fake metasub sample as its input.

metagraph/src/graph/annotated_graph_algorithm.cpp
karasikov: how can it be larger than 64 for int_vector?

hmusta: All of the operations need to at least be aligned to byte blocks. So, if for example, your width is 63, then you need to have atomic access to an unaligned block of >64-bits since the starting point has to be byte-aligned.

metagraph/src/cli/config/config.cpp
```diff
@@ -549,6 +539,12 @@ Config::Config(int argc, char *argv[]) {
         print_usage_and_exit = true;
     }
 
+    if ((identity == ASSEMBLE || identity == TRANSFORM)
+            && (infbase_annotators.size() && label_mask_file.empty())) {
+        std::cerr << "Error: annotator passed, but no label mask file provided" << std::endl;
```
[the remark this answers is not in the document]
hmusta: yes, I forgot to change this error message

metagraph/integration_tests/test_assemble.py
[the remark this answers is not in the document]
karasikov: what do you mean? You don't need to precompute anything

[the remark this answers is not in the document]
karasikov:
```suggestion
        assemble_command = f'{METAGRAPH} assemble -p {NUM_THREADS} \
                -a {self.tempdir.name}/annotation{anno_file_extension[self.anno_repr]} \
                -o {self.tempdir.name}/diff_contigs \
                --diff-assembly-rules {TEST_DATA_DIR}/example.diff.json \
                {self.tempdir.name}/graph{graph_file_extension[self.graph_repr]}'
```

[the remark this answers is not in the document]
hmusta: ok, I tried it out like this just now and it worked. I looked back at what I was doing before and for some reason, I tried stuff like this and it was complaining...
```python
{self.tempdir.name + "diff_contigs"}
```

but yes, I should have done it as suggested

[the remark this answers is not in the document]
karasikov: this thing works as well, you just need to escape the quotes correctly

metagraph/src/graph/annotated_graph_algorithm.cpp
karasikov: got it

metagraph/src/cli/assemble.cpp
```diff
@@ -17,66 +23,141 @@ using mtg::common::logger;
...
+                const auto [it, inserted] = background_labels.emplace(out_label.asString());
+                if (!inserted || foreground_labels.count(*it)
+                        || shared_foreground_labels.count(*it)
+                        || shared_background_labels.count(*it)) {
+                    has_overlap = true;
+                }
+            }
 
+            if (has_overlap)
+                logger->warn("Label sets overlap");
```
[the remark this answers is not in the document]
hmusta: It's semantically strange to say that a given label should be both included and excluded from an experiment (and the implementation makes the assumption that these sets are disjoint).

But I guess it would be useful to print them so the user can know where to look in their config file.

metagraph/src/cli/assemble.cpp
```diff
@@ -93,15 +174,47 @@ int assemble(Config *config) {
 
     logger->trace("Graph loaded in {} sec", timer.elapsed());
 
-    std::unique_ptr<graph::AnnotatedDBG> anno_graph;
     if (config->infbase_annotators.size()) {
-        anno_graph = initialize_annotated_dbg(graph, *config);
+        config->infbase = files.at(0);
```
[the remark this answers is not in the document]
hmusta: This check is already handled in `config.cpp`, I modified it there to print an error message.

README.md
```diff
@@ -153,12 +153,12 @@ Requires `M*V/8 + Size(BRWT)` bytes of RAM, where `M` is the number of rows in t
...
                         --unitigs \
                         -a <GRAPH_DIR>/annotation.column.annodbg \
-                        --label-mask-in LABEL_1 \
-                        --label-mask-in LABEL_2 \
-                        --label-mask-out LABEL_3 \
+                        --diff-assembly-rules diff_assembly_rules.json \
                         -o diff_assembled.fa
 ```
 
+See [`metagraph/tests/data/example.diff.json`](metagraph/tests/data/example.diff.json) and [`metagraph/tests/data/example_simple.diff.json`](metagraph/tests/data/example_simple.diff.json) for sample files.
```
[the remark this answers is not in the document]
hmusta: ok, I wrote up a first draft in the docs.

metagraph/src/cli/assemble.cpp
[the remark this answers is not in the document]
karasikov: Then it would be good to add an assert.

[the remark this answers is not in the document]
karasikov:
```suggestion
        assert(files.size() == 1u);
        config->infbase = files.at(0);
```

metagraph/src/graph/annotated_graph_algorithm.cpp
```diff
@@ -15,249 +12,416 @@ namespace graph {
...
-    const bool atomic = num_threads > 1;
+/**
+ * Return an int_vector<>, bit_vector pair, of lengths anno_graph.get_graph().max_index() * 2
+ * and anno_graph.get_graph().max_index(), respectively.
+ * For an index i, the int_vector at indices 2*i and 2*i + 1 represent the
+ * number of labels in labels_in and labels_out which the k-mer of index i is
+ * annotated with, respectively. The width of the int_vector<> is computed to be
+ * wide enough to contain counts up to num_labels.
+ * The returned bit_vector is a k-mer mask indicating those k-mers which are annotated
+ * with at least one in-label or out-label.
```
karasikov, line 30:
```suggestion
 * The returned bit_vector is a k-mer mask indicating those k-mers annotated
 * with at least one in-label or out-label.
```

metagraph/src/graph/annotated_graph_algorithm.cpp
```diff
@@ -15,249 +12,416 @@ namespace graph {
...
+    binmat.call_columns(label_out_codes,
+        [&](auto, const bitmap &rows) {
+            rows.call_ones([&](auto r) {
+                node_index i = AnnotatedDBG::anno_to_graph_index(r);
+                set_bit(indicator.data(), i, parallel, MO_RELAXED);
+                atomic_fetch_and_add(counts, i * 2 + 1, 1, vector_backup_mutex, MO_RELAXED);
+            });
+        },
+        num_threads
+    );
```
karasikov, line 353: Would be good to merge these two to reduce everything to a single call of call_columns. For hierarchically compressed representation this can greatly reduce the run time.

1. build a total list of columns: std::vector<Column> and indicators sdsl::vector<int> -- in/out/both
2. query all columns and update the counts according to the indicators (in/out/both)

metagraph/src/graph/annotated_graph_algorithm.cpp
```diff
@@ -15,249 +12,416 @@ namespace graph {
...
+
+void update_masked_graph_by_node(MaskedDeBruijnGraph &masked_graph,
+                                 const KeepNode &keep_node,
+                                 size_t num_threads) {
+    bool parallel = num_threads > 1;
+    std::atomic<size_t> kept_nodes(0);
+    size_t total_nodes = masked_graph.num_nodes();
+
+    const auto &in_mask = dynamic_cast<const bitmap_vector&>(masked_graph.get_mask()).data();
+    sdsl::bit_vector mask(in_mask);
```
karasikov, line 410:
```suggestion
    sdsl::bit_vector mask = in_mask;
```

karasikov: Use copy assignment operators to make it clear that no conversion is happening and this just creates a copy.

metagraph/src/graph/annotated_graph_algorithm.cpp
```diff
@@ -15,249 +12,416 @@ namespace graph {
...
+                                  size_t num_threads);
+
+// Regions of a graph mask which should be kept (i.e., masked in)
+typedef std::vector<std::pair<size_t, size_t>> Intervals;
+
+// Returns a vector of kept regions given a unitig and its corresponding path
+typedef std::function<Intervals(const std::string&, const std::vector<node_index>&)> GetKeptIntervals;
+
+// Returns true if a node is kept (i.e., masked in)
+typedef std::function<bool(node_index)> KeepNode;
```
karasikov, line 46: this only makes thing confusing. Better write `std::function<bool(node_index)>` -- it's short and descriptive.

metagraph/src/graph/annotated_graph_algorithm.cpp
```diff
@@ -15,249 +12,416 @@ namespace graph {
...
+// Assemble unitigs from the masked graph, then use get_kept_intervals to generate
+// regions which should be masked in. Update the graph mask accordingly.
+void update_masked_graph_by_unitig(MaskedDeBruijnGraph &masked_graph,
+                                   const GetKeptIntervals &get_kept_intervals,
+                                   size_t num_threads);
+
+// Iterate through all nodes in the masked graph, then use keep_node to decide
+// which should be masked in. Update the graph mask accordingly.
+void update_masked_graph_by_node(MaskedDeBruijnGraph &masked_graph,
+                                 const KeepNode &keep_node,
```
karasikov, line 57:
```suggestion
                                 const std::function<bool(node_index)> &keep_node,
```

metagraph/src/graph/annotated_graph_algorithm.cpp
```diff
@@ -15,249 +12,416 @@ namespace graph {
...
+
+        logger->trace("Filtering by node");
+        update_masked_graph_by_node(*masked_graph, [&](node_index node) {
+            uint64_t in_count = count_vector.first[node * 2];
+            uint64_t out_count = count_vector.first[node * 2 + 1];
+            uint64_t sum = in_count + out_count;
+            return in_count >= config.label_mask_in_kmer_fraction * sum
+                && out_count <= config.label_mask_out_kmer_fraction * sum;
+        }, num_threads);
+
```
karasikov, line 183: Why do you write all these callbacks? I see no point in them here. They bring no benefits.
Why not do a pass over `count_vector` in a loop and initialize everything directly?

karasikov: So you basically don't need function `update_masked_graph_by_node`.

karasikov (changes requested):

metagraph/src/graph/annotated_graph_algorithm.cpp
hmusta: good point, I implemented it this way. Since in and out are disjoint, I just did this using a std::vector<bool>

metagraph/src/graph/annotated_graph_algorithm.cpp
```diff
@@ -173,13 +164,28 @@ mask_nodes_by_label(const AnnotatedDBG &anno_graph,
         }
 
         logger->trace("Filtering by node");
-        update_masked_graph_by_node(*masked_graph, [&](node_index node) {
+        size_t total_nodes = masked_graph->num_nodes();
+        const auto &in_mask = dynamic_cast<const bitmap_vector&>(masked_graph->get_mask()).data();
+        sdsl::bit_vector mask = in_mask;
+
+        // TODO: make this part multithreaded
```
karasikov, line 171: this should be trivial

metagraph/src/graph/annotated_graph_algorithm.cpp
```diff
@@ -322,37 +322,30 @@ construct_diff_label_count_vector(const AnnotatedDBG &anno_graph,
...
-                set_bit(indicator.data(), i, parallel, MO_RELAXED);
-                atomic_fetch_and_add(counts, i * 2, 1, vector_backup_mutex, MO_RELAXED);
-            });
-        },
-        num_threads
-    );
 
-    binmat.call_columns(label_out_codes,
-        [&](auto, const bitmap &rows) {
+    binmat.call_columns(label_codes,
```
karasikov, line 342: As I remember, you do allow having the same label in "IN" and "OUT". So I don't see how this will work properly in that case. Or, am I wrong?

karasikov (changes requested):

metagraph/src/graph/annotated_graph_algorithm.cpp
karasikov: but can be done later, as you prefer

metagraph/src/graph/annotated_graph_algorithm.cpp
hmusta: I output a warning if it finds duplicates, but I think given this last patch, we should make it write an error and just skip that experiment.

karasikov: Printing a warning is not suitable in this case. If the behavior for a case is unpredictable or wrong, you must not allow that scenario to happen. That is, print an error and exit.
Otherwise, fix this block of code to behave properly.
Theoretically, I see no issues with having a non-zero overlap between IN and OUT. I suppose, there indeed might be practical cases where this will be required.

karasikov: For instance, "find sequences present in at least one of the samples input, but not in all of them".

hmusta: That's a good point, actually. I guess it won't change the code too much to support having overlapping labels.

metagraph/src/graph/annotated_graph_algorithm.cpp
```diff
@@ -15,249 +12,397 @@ namespace graph {
...
+            label_codes.push_back(code);
+    }
+    for (const std::string &label_out : labels_out) {
+        uint64_t code = label_encoder.encode(label_out);
+        auto [it, inserted] = code_to_indicator.emplace(code, 2);
+        if (inserted) {
+            label_codes.push_back(code);
+        } else {
+            it.value() = 3;
+        }
```
karasikov, line 338:
```suggestion
        uint64_t code = label_encoder.encode(label_out);
        code_to_indicator[code] |= 2;
```

metagraph/src/graph/annotated_graph_algorithm.cpp
```diff
@@ -15,249 +12,397 @@ namespace graph {
...
+    const auto &binmat = anno_graph.get_annotation().get_matrix();
+
+    tsl::hopscotch_map<uint64_t, uint8_t> code_to_indicator;
+    std::vector<uint64_t> label_codes;
+    label_codes.reserve(labels_in.size() + labels_out.size());
+    for (const std::string &label_in : labels_in) {
+        uint64_t code = label_encoder.encode(label_in);
+        auto [it, inserted] = code_to_indicator.emplace(code, 1);
+        if (inserted)
+            label_codes.push_back(code);
```
karasikov, line 329:
```suggestion
        code_to_indicator[code] = 1;
```

metagraph/src/graph/annotated_graph_algorithm.cpp
karasikov: and then initialize `label_codes` from all pairs in `code_to_indicator`

karasikov (approved):
Can you add some integration tests for the overlapping labels, which would have failed without the fix?

PR #227, opened 2020-10-28
Added atomic_adds, atomic increment of label counts

hmusta:
We should test to see if this speeds things up.

### DIFF
--- a/metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
+++ b/metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
@@ -83,8 +83,25 @@ void ColumnCompressed<Label>::add_label_counts(const std::vector<Index> &indices
     assert(indices.size() == counts.size());
 
     const auto &columns = get_matrix().data();
+    {
+        std::lock_guard<std::mutex> lock(count_mutex_);
+        if (relation_counts_.size() != columns.size()) {
+            relation_counts_.resize(columns.size());
+            relation_counts_resize_mutex_vector_.resize(columns.size());
+        }
+    }
+
     for (const auto &label : labels) {
         const auto j = label_encoder_.insert_and_encode(label);
+
+        // lock mutex for this label
+        char *mu = &relation_counts_resize_mutex_vector_[j];
+        while (!__atomic_test_and_set(mu, __ATOMIC_SEQ_CST)) {}
+        if (!relation_counts_[j].size()) {
+            relation_counts_[j] = aligned_int_vector(columns[j]->num_set_bits(), 0, kCountBits, 16);
+        }
+        // unlock mutex
+        __atomic_clear(mu, __ATOMIC_SEQ_CST);
 
         for (size_t i = 0; i < indices.size(); ++i) {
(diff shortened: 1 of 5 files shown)


hmusta:
After running this and `dev` on a graph made from 20 random uncleaning samples, it looks like there is room for improving the parallelism of count annotation. Both of the runs below were with 30 threads and were only timed on the count annotation portion.

before:
```
User time (seconds): 31259.14
        System time (seconds): 7255.55
        Percent of CPU this job got: 385%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 2:46:33
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 14448976
```
        
after:
```
User time (seconds): 41121.22
        System time (seconds): 7411.13
        Percent of CPU this job got: 1418%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 57:00.32
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 19453840
```

karasikov:
> After running this and `dev` on a graph made from 20 random uncleaning samples, it looks like there is room for improving the parallelism of count annotation. Both of the runs below were with 30 threads and were only timed on the count annotation portion.
> 
> before:
> 
> ```
> User time (seconds): 31259.14
>         System time (seconds): 7255.55
>         Percent of CPU this job got: 385%
>         Elapsed (wall clock) time (h:mm:ss or m:ss): 2:46:33
>         Average shared text size (kbytes): 0
>         Average unshared data size (kbytes): 0
>         Average stack size (kbytes): 0
>         Average total size (kbytes): 0
>         Maximum resident set size (kbytes): 14448976
> ```
> 
> after:
> 
> ```
> User time (seconds): 41121.22
>         System time (seconds): 7411.13
>         Percent of CPU this job got: 1418%
>         Elapsed (wall clock) time (h:mm:ss or m:ss): 57:00.32
>         Average shared text size (kbytes): 0
>         Average unshared data size (kbytes): 0
>         Average stack size (kbytes): 0
>         Average total size (kbytes): 0
>         Maximum resident set size (kbytes): 19453840
> ```

Was this without `--separately`?

Can you try it with?

karasikov:
The more complex we make the code (as in this PR), the harder it is to make sure there are no bugs, and when there is one, it's nearly impossible to find it. Especially if everything is atomic and hence there are never race conditions, so you can't use tools like thread sanitizer.

So I am not a fan of optimizations like this where the complexity is increased but the benefit is not clear.
We always annotate with `--separately` (in fact, maybe it makes sense to make that a default), and this PR doesn't make it any faster in that mode.

karasikov:
We can merge updates in `vector_algorithm.hpp` though, but you'll need someone else to verify that code and ideally write unit tests as well.

hmusta:
I think for now we can keep this as a draft since those methods in `vector_algorithm` aren't used anywhere else

PR #229, opened 2020-11-02
Eliminate SortedSetDisk when building row diffs

danieldanciu:
- simply use a file merger, most of the functionality of SSD (ChunkedWaitQueue, de-duping) is not used anyway
 - I added the option to disable intermediate merging in SSD, even though in the end we are not using it; I think it's useful to have it.

### DIFF
--- a/metagraph/src/common/file_merger.hpp
+++ b/metagraph/src/common/file_merger.hpp
@@ -32,15 +32,21 @@ class MergeHeap {
 
   public:
     inline void emplace(T el, uint32_t idx) {
-        els.push_back({ el, idx });
-        std::push_heap(els.begin(), els.end(), cmp_);
+        els.emplace_back(el, idx);
+        std::push_heap(els.begin(), els.end(),
+                       [&](const value_type &a, const value_type &b) {
+                           return compare_(a.first, b.first);
+                       });
     }
 
     inline const value_type& top() const { return els.front(); }
 
     inline value_type pop() {
-        value_type result = els.front();
-        std::pop_heap(els.begin(), els.end(), cmp_);
+        std::pop_heap(els.begin(), els.end(),
+                      [&](const value_type &a, const value_type &b) {
+                          return compare_(a.first, b.first);
+                      });
+        value_type result = els.back();
         els.pop_back();
         return result;
     }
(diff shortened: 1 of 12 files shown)


metagraph/src/common/file_merger.hpp
```diff
@@ -0,0 +1,137 @@
...
+        return result;
+    }
+
+    inline bool empty() const { return els.empty(); }
+
+  private:
+    // elements stored as a min-heap
+    std::vector<value_type> els;
+    Compare compare_ = Compare();
+};
```
karasikov, line 60: Why do we need yet another merger? Why not use the ones from Elias Fano mergers?

metagraph/src/common/sorted_sets/sorted_set_disk_base.hpp
```diff
@@ -182,18 +185,15 @@ class SortedSetDiskBase {
     ChunkedWaitQueue<T> merge_queue_;
 
     /**
-     * Thread pool for doing the "level 1" merging, i.e. merging #MERGE_L1_COUNT chunk
+     * Thread pool for doing the "level 1" merging, i.e. merging #merge_count_ chunk
      * files at a time, while new files are still being added to the data structure.
      */
-    ThreadPool async_merge_l1_ = ThreadPool(1, 100);
+    ThreadPool async_merge_l1_;;
```
karasikov, line 191:
```suggestion
    ThreadPool async_merge_l1_;
```

karasikov (approved):

metagraph/src/common/file_merger.hpp
karasikov, line 60: Did you benchmark this thing? Is this faster than the same thing from `elias_fano_file_merger.hpp`?

danieldanciu: I wanted a simpler merger that doesn't compress/decompress data, as in this case disk space is not an issue. But we can totally use the EF encoders, and the corresponding merger, too.

danieldanciu: I benchmarked it back in the time when I introduced the Elias-Fano encoder. I remember seeing no significant performance penalty with EF, using the Google Cloud Network disks. For SSDs I don't think EF can beat dumping the data directly, so I preferred this version. I can add a TODO: to do a proper benchmarking once the paper is out and time is not an issue.

karasikov: There is a `MergeHeap` there with exactly the same interface. The difference is only in internals.

So I think we should benchmark both of them and select the fastest one.

PR #233, opened 2020-11-12
Conversion of RowDiff to Multi-BRWT

karasikov:
Fixes and improvements

### DIFF
--- a/metagraph/src/annotation/row_diff_builder.cpp
+++ b/metagraph/src/annotation/row_diff_builder.cpp
@@ -235,11 +235,11 @@ void traverse_anno_chunked(
         #pragma omp parallel for num_threads(num_threads) schedule(dynamic)
         for (size_t l_idx = 0; l_idx < col_annotations.size(); ++l_idx) {
             for (size_t j = 0; j < col_annotations[l_idx].num_labels(); ++j) {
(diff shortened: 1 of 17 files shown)


metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -214,32 +214,40 @@ void traverse_anno_chunked(
...
-                ++pred_chunk_idx[i + 1];
-                pred_chunk.push_back(*pred_it);
-                ++pred_it;
+                ++pred_chunk_idx[i];
                 ++pred_boundary_it;
             }
             ++pred_boundary_it;
         }
 
+        // read all predecessors for the block
```
hmusta, line 234: Maybe add some runtime checks to make sure that the buffers are at the ends of the files? This could be a good way to check for corruption.

hmusta (approved):

metagraph/src/annotation/row_diff_builder.cpp
karasikov: This file is read partially (just a chunk).
Indeed, it would be good to add checks for correctness, but it doesn't seem straightforward.
Maybe we'll need to come up with a universal solution for such cases. Otherwise, it would require writing too many hacky checks, which is not good.

PR #235, opened 2020-11-18
Various minor fixes

hmusta:
1) Added `BOSS::tighten_range` to clean up some of the BOSS internal functions
2) Discovered a legitimate case where an alignment can lead to an assertion failure, but without breaking the rest of the alignment. Fixing it properly requires some structural changes, so the assertion has been replaced with a warning.
3) Added some more asserts to the seeder

Details about 2)
Basically, if a deletion was extended, but then the optimal operation at that position is no longer a deletion, the score of the subsequent position becomes incorrect since it used the gap extension instead of the opening penalty.

### DIFF
--- a/metagraph/src/graph/representation/canonical_dbg.cpp
+++ b/metagraph/src/graph/representation/canonical_dbg.cpp
@@ -9,6 +9,7 @@ namespace graph {
 
 using mtg::common::logger;
 
+
 // If the graph has even order and is primary, then only is_palindrome_cache_
 // is required. rev_comp_cache_ is only required if the underlying graph is not
 // primary.
@@ -132,6 +133,7 @@ void CanonicalDBG
         call_incoming_kmers(node - offset_, [&](node_index next, char c) {
             ::reverse_complement(&c, &c + 1);
             callback(reverse_complement(next), c);
+            assert(traverse(node, c) == reverse_complement(next));
         });
         return;
     }
@@ -150,7 +152,9 @@ void CanonicalDBG
         std::vector<node_index> children(alph_size);
 
         graph_.call_outgoing_kmers(node, [&](node_index next, char c) {
-            children[alph_map_[c]] = next;
+            if (c != boss::BOSS::kSentinel)
+                children[alph_map_[c]] = next;
+
             --alph_size;
         });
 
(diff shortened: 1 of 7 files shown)


metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -211,7 +218,9 @@ ::call_incoming_kmers(node_index node, const IncomingEdgeCallback &callback) con
         std::vector<node_index> parents(alph_size);
 
         graph_.call_incoming_kmers(node, [&](node_index prev, char c) {
-            parents[alph_map_[c]] = prev;
+            if (c != boss::BOSS::kSentinel)
+                parents[alph_map_[c]] = prev;
+
             --alph_size;
```
karasikov, line 224: What does this alph_size refer to? Better introduce another variable and pick an appropriate name for it.

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -228,7 +237,7 @@ ::call_incoming_kmers(node_index node, const IncomingEdgeCallback &callback) con
                 ::reverse_complement(&c, &c + 1);
 
                 rev_seq.back() = c;
-                prev = map_sequence_to_nodes(graph_, rev_seq)[0];
+                prev = graph_.kmer_to_node(rev_seq);
```
karasikov, line 240: Why did it call `map_sequence_to_nodes` before?

Did we need to map it to canonical k-mers, so `kmer_to_node` did not work?

karasikov (approved):

metagraph/src/graph/representation/canonical_dbg.cpp
hmusta: I think it was just an oversight on my part. `map_sequence_to_nodes` doesn't map to canonical k-mers anyway, so these two are functionally equivalent.

metagraph/src/graph/representation/canonical_dbg.cpp
hmusta: I renamed it to `max_num_edges_left`. It's basically used to check whether we should bother to check the incoming nodes of the reverse complement.

PR #236, opened 2020-11-21
Ensure that exceptions thrown by worker threads are captured

hmusta:


metagraph/src/common/threads/threading.hpp
```diff
@@ -60,20 +60,39 @@ class ThreadPool {
...
+        auto task_exception = std::make_shared<std::exception_ptr>(nullptr);
         auto task = std::make_shared<std::packaged_task<return_type()>>(
-            std::bind(std::forward<F>(f), std::forward<Args>(args)...)
+            [f=std::move(wrap),task_exception]() mutable -> return_type {
+                try {
+                    return f();
+                } catch (...) {
+                    // catch and store the exception
+                    *task_exception = std::current_exception();
+                    std::rethrow_exception(*task_exception);
```
karasikov, line 72: I think we need to remove this one. Otherwise it will throw 2 exceptions: one here and another one later on line 82.

karasikov (changes requested):

metagraph/src/common/threads/threading.hpp
hmusta: There was another issue where even if only one exception was being thrown, it was not getting caught by the main thread. I modified this so that exceptions are now thrown in the main thread.

hmusta: But now there's an issue where exceptions are only handled while the thread pool is joining...

metagraph/src/common/threads/threading.hpp
```diff
@@ -57,11 +57,24 @@ class ThreadPool {
...
-            std::bind(std::forward<F>(f), std::forward<Args>(args)...)
+            [this,fc=std::bind(std::forward<F>(f), std::forward<Args>(args)...)]() mutable -> return_type {
+                try {
+                    return fc();
+                } catch (...) {
+                    // catch and store the exception
+                    if (!exception_thrown_.exchange(true))
+                        exception_ = std::current_exception();
+
+                    return return_type();
```
hmusta, line 75: this won't work if `return_type` has no default constructor

metagraph/src/common/threads/threading.hpp
karasikov: Maybe it's fine that it's not caught in the main thread? Why is that a problem?

hmusta: Because if you have a `try...catch` block in the main thread, it won't catch if there's an exception thrown by a worker thread.

hmusta: With the current version, it almost works as expected. The only exception is if `get` is called on all futures and `join` is not called on the `ThreadPool`. In this case, the exception is thrown in the `ThreadPool` destructor, which is undefined behaviour and on my system leads to the exception not being caught.

karasikov: Maybe that's ok that it throws in a different thread? So if you want to catch an exception, you'd need to wrap the task into a try section yourself.

Otherwise, it will wait until join, which is not ideal...

hmusta: There are some uses for having it this way. If we're using the thread pool to process a bunch of independent columns, this this strategy will allow for the ones that ran successfully to still proceed to completion.

The obvious drawback is that if all of the results are needed for a subsequent step, then this approach does a lot of wasteful computation before throwing.

hmusta: I'll have to try again to see if it still works if the user puts a try block in their task, I think it didn't work when I tried it last time.

karasikov: But if a user wants to allow it to continue with one of the tasks failing, he should obviously wrap the task into a try .. catch section.

metagraph/tests/test_utils.cpp
```diff
@@ -469,9 +469,16 @@ TEST(ThreadPool, MultiThreadException) {
...
+            std::exception_ptr exception = nullptr;
             for (size_t t = 0; t < 1000; ++t) {
                 result.emplace_back(pool.enqueue([&](size_t i) {
-                    throw std::runtime_error("error");
+                    try {
+                        throw std::runtime_error("error in test");
+                    } catch (...) {
+                        if (!exception_count.fetch_add(1))
+                            exception = std::current_exception();
+                    }
```
karasikov, line 481: Hm... this test seems useless. It doesn't actually test the thread pool.

This code will never throw because you catch all exceptions and increment the counter.

So it the result of the test will never depend on the implementation of the thread pool and how exceptions are handled there.

I will try to propose a different test...

karasikov: It seems the only thing it makes sense to test here is that it throws in worker threads.

I test it by checking an implication of this -- that the main thread dies.

karasikov (approved):

hmusta:
Should I merge this?

karasikov:
> Should I merge this?

Yes, please, if you're ok with the solution.

PR #237, opened 2020-11-23
Don't open more than N files at once in RowDiff construction

karasikov:


metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -443,34 +444,90 @@ void convert_batch_to_row_diff(const std::string &pred_succ_fprefix,
...
         for (size_t j = 0; j < label_encoders[l_idx].size(); ++j) {
-            auto call_ones = [&](const std::function<void(uint64_t)>& call) {
+            auto call_ones = [&](const std::function<void(uint64_t)> &call) {
                 std::vector<std::string> filenames;
                 for (uint32_t chunk = 0; chunk < num_chunks[l_idx][j]; ++chunk) {
                     filenames.push_back(tmp_file(l_idx, j, chunk));
                 }
+
+                // if there are too many chunks, merge them into larger ones
+                while (filenames.size() > files_open_per_thread) {
```
danieldanciu, line 467: I would prefer all this merge logic to be in `common::merge_files` rather than here. Perhaps `merge_files` can take an extra parameter, `max_open_files` which limits the total number of files it's allowed to open when merging. Or just copy this section of code (module removing the first file) into a new method `merge_files_bounded`.

danieldanciu (approved):

metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -443,34 +444,90 @@ void convert_batch_to_row_diff(const std::string &pred_succ_fprefix,
...
+
+                        common::merge_files<uint64_t>(to_merge, [&](uint64_t i) {
+                            buf.push_back(i);
+                            if (buf.size() == buf.capacity()) {
+                                f.write(reinterpret_cast<const char *>(buf.data()),
+                                        buf.size() * sizeof(decltype(buf.front())));
+                                buf.resize(0);
+                            }
+                        });
+                        if (buf.size()) {
```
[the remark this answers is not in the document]
karasikov: yeah, maybe

metagraph/src/annotation/row_diff_builder.cpp
karasikov: yeah, I thought about that, but then thought we'd also need to pass a path for temp files, so it would change the interfaces there and looks like a bigger refactoring.

PR #238, opened 2020-11-24
Fixed potential out-of-bounds access when splitting edge queue

hmusta:
`EdgeQueue::size()` is computed as the sum of the sizes of its two vectors. In the `split_half` method, `h` is adjusted based on `size()`, even though it's in an intermediate state. So, there is the potential for `h` to be negative.


karasikov (approved):

hmusta:
For posterity: this fixed a bug introduced in c770202c707e193d09ae7e6b2f26a4ab0ad5abaf

PR #239, opened 2020-11-24
Don't suppress death tests in Release compile mode

karasikov:


metagraph/tests/test_helpers.hpp
```diff
@@ -12,6 +12,12 @@ namespace {
 
 using namespace mtg;
 
+// run debug death tests only for debug to check assertions
+#ifdef NDEBUG
+#undef ASSERT_DEBUG_DEATH
+#define ASSERT_DEBUG_DEATH(...) (void)0
+#endif
```
hmusta, line 19: doesn't `ASSERT_DEBUG_DEATH` already do this? as in, do nothing if `NDEBUG`?

hmusta (changes requested):

metagraph/tests/test_helpers.hpp
karasikov: That's what I thought first, but actually, it still runs the statement inside. It just doesn't expect the process to die.
Quite strange behavior, to be honest, not sure what was their original intention...

So in this case, if we execute the statement, we may get a segfault. That's why we have to switch it off for non-debug.

hmusta: hmm, then maybe a better approach will be to only have this on ASSERT_DEATHs which take up a ton of time (like the bit_vector tests)

karasikov: What do you mean?

The current scheme is:
`ASSERT_DEATH(statement, ...)` -> run the statement for any build and check that the process dies
`ASSERT_DEBUG_DEATH(statement, ...)` -> same as ASSERT_DEATH, except for the builds with NDEBUG turned on, when it doesn't even run the statement to avoid possible segfaults.

Plus, `-DNO_DEATH_TESTS` in cmake turns off all death tests if a user doesn't want them.

karasikov: `ASSERT_DEATH(statement, ...)` is essential, for instance, to make sure that the process dies if one of the workers in the thread pool throws an exception which is then not caught. And that this happens even in Release, where `NDEBUG` is defined.

karasikov: The issue with running time for bit_vector tests should be considered separately.
It's not related to the aim of this PR.

hmusta: ah, I see, so this also covers the case where an ASSERT_DEATH might fail because a failed assertion may not cause the program to halt. ok, makes sense

PR #240, opened 2020-11-25
New compressed graph state

karasikov:
old `stat` -> deprecated
new `stat` = old `small` (same format)
new `small`: a new compressed representation (usually, `sd_vector` for `last` and `wt_huff<rrr63>` for `W`)

New `small` has about 2x smaller memory footprint than `stat`

### DIFF
--- a/metagraph/src/common/vectors/wavelet_tree.hpp
+++ b/metagraph/src/common/vectors/wavelet_tree.hpp
@@ -218,11 +218,9 @@ class wavelet_tree_sdsl : public wavelet_tree {
 };
 
 
-typedef wavelet_tree_sdsl_fast<> wavelet_tree_stat;
+typedef wavelet_tree_sdsl<> wavelet_tree_stat;
 
-typedef wavelet_tree_sdsl<> wavelet_tree_small;
-
-typedef wavelet_tree_sdsl<sdsl::wt_huff<sdsl::rrr_vector<63>>> wavelet_tree_compr;
+typedef wavelet_tree_sdsl<sdsl::wt_huff<sdsl::rrr_vector<63>>> wavelet_tree_small;
 
 typedef partite_vector<> wavelet_tree_fast;
 
--- a/metagraph/src/common/vectors/wavelet_tree.hpp
+++ b/metagraph/src/common/vectors/wavelet_tree.hpp
@@ -207,7 +207,7 @@ class wavelet_tree_sdsl : public wavelet_tree {
     bool load(std::istream &in);
     void serialize(std::ostream &out) const;
 
-    void clear() { wwt_ = t_wt_sdsl(); }
+    void clear() { wwt_ = t_wt_sdsl(); count_.assign(count_.size(), 0); }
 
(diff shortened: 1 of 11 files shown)


metagraph/src/common/vectors/wavelet_tree.hpp
```diff
@@ -207,7 +207,7 @@ class wavelet_tree_sdsl : public wavelet_tree {
     bool load(std::istream &in);
     void serialize(std::ostream &out) const;
 
-    void clear() { wwt_ = t_wt_sdsl(); }
+    void clear() { wwt_ = t_wt_sdsl(); count_.assign(count_.size(), 0); }
```
hmusta, line 210: so this was broken before this? do we not have any unit tests which test `clear`?

metagraph/src/graph/representation/succinct/boss.hpp
```diff
@@ -283,10 +283,10 @@ class BOSS {
      * Representation states of the BOSS table.
      *
      * SMALL is the smallest
-     * STAT provides a good space/time tradeoff
+     * STAT provides the best space/time trade-off
```
hmusta, line 286: why not say "the best (at the time of writing)" to make the statement less strong? :)

metagraph/src/graph/representation/succinct/boss.hpp
```diff
@@ -283,10 +283,10 @@ class BOSS {
...
      * FAST is the fastest but large
```
hmusta, line 287: For better clarity, maybe it would be useful to also describe which encoding is used for each of L and W in these states

hmusta (changes requested):

metagraph/src/graph/representation/succinct/boss.hpp
hmusta: same in `dbg_succinct.hpp`

metagraph/src/graph/representation/succinct/boss.hpp
karasikov: Strong is fine here, it's not a theorem, it's not even defined what this tarde-off is.

If at some point some better representations will be developed, this state should be updated and we will switch to those better representations.

metagraph/src/common/vectors/wavelet_tree.hpp
karasikov: It wasn't designed to be re-used after you call `clear()`, so there were no issues.
I just thought it's easy to make it re-usable and added this cheap change.
So now a vector can be re-used after clear().

But I have no use-cases for this right now. When we do, we should write unit tests.
For now, I'd still assume that clear() may invalidate the vector.

karasikov: So it's not a "bugfix", rather geeneralization/improvement which may be helpful in the future.

metagraph/src/graph/representation/succinct/boss.hpp
karasikov: done

hmusta (approved):

PR #242, opened 2020-11-26
DBGSuccinctRange wrapper to speed up and simplify sub-k seed-and-extend

hmusta:
This adds a wrapper DBGSuccinctRange which adds virtual nodes corresponding to dummy source and sink k-mers not present in the underlying DBGSuccinct. This way, seeds of length k can be implemented without testing all possible nodes.

This deprecates SuffixSeeder, and also provides a conceptually simpler approach to sub-k seeding on primary graphs wrapped around CanonicalDBG.

TODO: add unit tests to more directly test out the traversal functions in DBGSuccinctRange


hmusta:
Other than the missing unit tests, this should be ready now

metagraph/src/graph/representation/succinct/dbg_succinct_range.cpp
```diff
@@ -0,0 +1,864 @@
...
+    const auto &boss_graph = dbg_succ_.get_boss();
+    assert(begin <= end);
+    assert(boss_graph.get_k() >= static_cast<size_t>(end - begin));
+
+    if (std::find(begin, end, boss_graph.alph_size) != end)
+        return 0;
+
+    auto [first, last, seq_it] = boss_graph.index_range(begin, end);
+
+    const boss::BOSS::TAlphabet *min_match_pos = require_exact_length ? end : begin + 1;
```
hmusta, line 446: for some reason, setting `begin + min_match_length_` instead of `begin + 1` almost doubles the run time

PR #243, opened 2020-12-04
Faster merge of row reduction vectors

karasikov:
Faster merge of row reduction vectors for anchor optimization in RowDiff.
Also uses less disk space.

### DIFF
--- a/metagraph/src/annotation/row_diff_builder.cpp
+++ b/metagraph/src/annotation/row_diff_builder.cpp
@@ -631,7 +631,7 @@ void optimize_anchors_in_row_diff(const std::string &graph_fname,
     static_assert(sizeof(uint64_t) * 8 > ROW_REDUCTION_WIDTH);
     const uint64_t MINUS_BIT = 1llu << (row_reduction.front().width() - 1);
 
-    ProgressBar progress_bar(anchors.size(), "Optimize anchors", std::cerr,
+    ProgressBar progress_bar(anchors.size(), "Assign extra anchors", std::cerr,
                              !common::get_verbose());
     for (uint64_t i = 0; i < anchors.size(); i += BLOCK_SIZE) {
         // adjust if the last block
--- a/metagraph/src/annotation/row_diff_builder.cpp
+++ b/metagraph/src/annotation/row_diff_builder.cpp
@@ -588,7 +588,7 @@ void optimize_anchors_in_row_diff(const std::string &graph_fname,
         auto path = p.path();
         if (utils::ends_with(path, row_reduction_extension)) {
             logger->info("Found row reduction vector {}", path);
-            row_reduction.push_back(sdsl::int_vector_buffer(path, std::ios::in, 1024 * 1024));
+            row_reduction.emplace_back(path, std::ios::in, 1024 * 1024);
 
             if (row_reduction.back().size() != row_reduction.front().size()) {
                 logger->error("Row reduction vectors have different sizes");
--- a/metagraph/src/annotation/row_diff_builder.cpp
+++ b/metagraph/src/annotation/row_diff_builder.cpp
@@ -11,7 +11,7 @@
 #include "common/vectors/bit_vector_sd.hpp"
 #include "graph/annotated_dbg.hpp"
 
-constexpr uint64_t BLOCK_SIZE = 1 << 25;
+constexpr uint64_t BLOCK_SIZE = 1 << 27;
 constexpr uint64_t ROW_REDUCTION_WIDTH = 32;
 constexpr uint32_t MAX_NUM_FILES_OPEN = 2000;
 
@@ -608,8 +608,6 @@ void optimize_anchors_in_row_diff(const std::string &graph_fname,
         exit(1);
     }
 
-    const uint8_t width = row_reduction.front().width();
-
     sdsl::bit_vector anchors;
     uint64_t num_anchors_old;
     {
@@ -630,6 +628,9 @@ void optimize_anchors_in_row_diff(const std::string &graph_fname,
     }
 
     std::vector<uint64_t> buf;
+    static_assert(sizeof(uint64_t) * 8 > ROW_REDUCTION_WIDTH);
+    const uint64_t MINUS_BIT = 1llu << (row_reduction.front().width() - 1);
+
     ProgressBar progress_bar(anchors.size(), "Optimize anchors", std::cerr,
                              !common::get_verbose());
     for (uint64_t i = 0; i < anchors.size(); i += BLOCK_SIZE) {
@@ -647,7 +648,7 @@ void optimize_anchors_in_row_diff(const std::string &graph_fname,
 
         for (uint64_t t = 0; t < buf.size(); ++t) {
             // check if the reduction is negative
-            if (buf[t] >> (width - 1))
+            if (buf[t] & MINUS_BIT)
                 anchors[i + t] = true;
         }
 
--- a/metagraph/src/annotation/row_diff_builder.cpp
+++ b/metagraph/src/annotation/row_diff_builder.cpp
@@ -11,7 +11,7 @@
 #include "common/vectors/bit_vector_sd.hpp"
 #include "graph/annotated_dbg.hpp"
 
-constexpr uint64_t BLOCK_SIZE = 1 << 27;
+constexpr uint64_t BLOCK_SIZE = 1 << 25;
 constexpr uint64_t ROW_REDUCTION_WIDTH = 32;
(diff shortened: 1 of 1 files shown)


metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -583,76 +583,76 @@ void optimize_anchors_in_row_diff(const std::string &graph_fname,
...
+    }
 
-        if (filenames.size() % 2)
-            filenames_new.back() = filenames.back();
-
-        logger->trace("Merged {} row reduction vectors into {}",
-                      filenames.size(), filenames_new.size());
+    std::vector<uint64_t> buf;
+    static_assert(sizeof(uint64_t) * 8 > ROW_REDUCTION_WIDTH);
+    const uint64_t MINUS_BIT = 1llu << (row_reduction.front().width() - 1);
```
hmusta, line 632: This won't always work, since you can still have an overflow and have this bit be 0. For example, if width = 2 and the value is 8, the check below will fail.

```suggestion
    const uint64_t OVERFLOW_MASK = 0xFFFFFFFFFFFFFFFF << row_reduction.front().width();
```

metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -583,76 +583,76 @@ void optimize_anchors_in_row_diff(const std::string &graph_fname,
...
-        std::ifstream f(original_anchors_fname, ios::binary);
-        old_anchors.load(f);
-    }
-    uint64_t num_anchors_old = old_anchors.num_set_bits();
+        __atomic_thread_fence(__ATOMIC_ACQUIRE);
 
-    sdsl::int_vector_buffer row_reduction(filenames[0], std::ios::in, 1024 * 1024);
+        for (uint64_t t = 0; t < buf.size(); ++t) {
+            // check if the reduction is negative
+            if (buf[t] & MINUS_BIT)
```
hmusta, line 651:
```suggestion
            if (buf[t] & OVERFLOW_MASK)
```

hmusta (changes requested):

metagraph/src/annotation/row_diff_builder.cpp
hmusta: Unless I misunderstood something and this is checking something else.

metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -583,76 +583,76 @@ void optimize_anchors_in_row_diff(const std::string &graph_fname,
...
+            if (row_reduction.back().size() != row_reduction.front().size()) {
+                logger->error("Row reduction vectors have different sizes");
+                exit(1);
+            }
+            if (row_reduction.back().width() != row_reduction.front().width()) {
+                logger->error("Row reduction vectors have different width: {} != {}",
+                              row_reduction.back().width(),
+                              row_reduction.front().width());
+                exit(1);
+            }
```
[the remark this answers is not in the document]
karasikov: these are reduction vectors (one per batch), we never have many of them. Usually -- 1-10.

metagraph/src/annotation/row_diff_builder.cpp
karasikov: I think it's good, yours won't work for negative arithmetics. Could you double check please?

hmusta: alternately, one could sign-extend the values read from `row_reduction` to get `uint64_t`, then do the addition, and finally check the most significant bit to see if it's set

hmusta: so the summation line would look something like
```c++
static_cast<uint64_t>(static_cast<int64_t>(static_cast<int32_t>(row_reduction[j][i + t])))
```
but this looks ugly...

karasikov: Is there any problem with the current code?

karasikov: Ah, you suggest to cast it to signed/unsigned, and then one doesn't have to bother about the internals.

A downside is that it would work only with 32 bits, while the current thing is meant to work with arbitrary width.

Though maybe it's fine to hardcode this... Should we do that?

hmusta: so a minimal example of the problem I'm talking about looks like this
```c++
#include <cstdint>
#include <limits>
#include <iostream>

int main() {
    uint8_t width = 32;
    uint64_t a = 1llu << (width - 1);
    uint64_t b = a + a;

    uint64_t mask = 1llu << (width - 1);
    std::cout << (b & mask) << "\n";
    return 0;
}
```
which outputs 0

hmusta: but fixing this for arbitrary widths becomes a bit complicated because we can't use casts to do the conversion

metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -583,76 +583,76 @@ void optimize_anchors_in_row_diff(const std::string &graph_fname,
...
     }
 
-    if (filenames.size() > 1u)
-        logger->trace("Merging row reduction vectors");
-
-    while (filenames.size() > 1u) {
-        std::vector<std::string> filenames_new((filenames.size() + 1) / 2);
+    sdsl::bit_vector anchors;
+    uint64_t num_anchors_old;
+    {
```
danieldanciu, line 613: Is this block intentional? It's already delimited by newlines, doesn't seem to bring much in readability

metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -583,76 +583,76 @@ void optimize_anchors_in_row_diff(const std::string &graph_fname,
...
+                logger->error("Row reduction vectors have different width: {} != {}",
+                              row_reduction.back().width(),
+                              row_reduction.front().width());
+                exit(1);
+            }
         }
     }
 
-    if (!filenames.size()) {
+    if (!row_reduction.size()) {
```
danieldanciu, line 606: I still find `if(row_reduction.empty())` more natural (and type safe), as it doesn't use the old C bool-int conversion.

danieldanciu (approved):
This looks great. Glad to see the annoying merging of row reductions gone.

metagraph/src/annotation/row_diff_builder.cpp
danieldanciu: What is the problem with the current code? Isn't this value bounded by the no of columns? We are many orders of magnitude below overflowing an int32, there is no way we can overflow an int64.

hmusta: It won't happen given the current sizes, but if we ever change the width to something smaller, we might run into this issue.

karasikov: Harun, your example shows the overflow issue, not the incorrectness of arithmetics.

The overflow is unavoidable since we use a finite number of bits for numbres.

The thing we must make correct though is the emulation of negative numbers and their addition.

Like, `A + B - A = B` for all |A|, |B| < 2^31.

This means, for instance, that the mask has to be `1llu << (row_reduction.front().width() - 1);` and not `const uint64_t OVERFLOW_MASK = 0xFFFFFFFFFFFFFFFF << row_reduction.front().width();`.
Because of the following:
```c++
delta[0] = 0;  // 0x00000000
delta[0] -= 1;  // 0xFFFFFFFF
uint64_t x = delta[0]; // 0x00000000FFFFFFFF  (because of truncation with 32 bits)
x += 1; // 0x0000000100000000  (because of truncation with 32 bits)
x & OVERFLOW_MASK != 0 // so it looks like x<0, which is wrong result, because the actual result should be 0.
```

metagraph/src/annotation/row_diff_builder.cpp
karasikov: It's to automatically remove all temp variables like `ifstream` and `old_anchors` which will not be needed below.

danieldanciu: Maybe add a comment at the beginning of the block to justify its existence. 

Did you notice any difference when not using the block? I suspect the compiler will destroy those variables for you anyway when they are not needed if it can apply the "as if" rule.

karasikov: It's just for readability.

The function is very large, so I want to avoid creating and keeping variables that are never needed below.
Otherwise, I would have to keep everything in my head, because never know if that variable will or will not be used later.
This thing clearly shows -- these are local temporary objects that are not used later.

Also, I can use that variable name below without messing with those that are created above.

PR #245, opened 2020-12-09
Support for GCC 8

hmusta:
List of changes:
1) Fixed a few lines which only work in GCC-9+
2) Add GCC-8 to the list of compilers in the CI workflow

### DIFF
--- a/metagraph/src/graph/alignment/aligner_extender_methods.cpp
+++ b/metagraph/src/graph/alignment/aligner_extender_methods.cpp
@@ -237,10 +237,12 @@ inline void update_del_scores(const DBGAlignerConfig &config,
 
 #ifdef __AVX2__
 
+// Drop-in replacement for _mm_loadu_si64
 inline __m128i mm_loadu_si64(const void *mem_addr) {
     return _mm_loadl_epi64((const __m128i_u*)mem_addr);
 }
 
+// Drop-in replacement for _mm_storeu_si64
 inline void mm_storeu_si64(void *mem_addr, __m128i a) {
     _mm_storel_epi64((__m128i_u*)mem_addr, a);
 }
--- a/metagraph/src/graph/alignment/aligner_extender_methods.cpp
+++ b/metagraph/src/graph/alignment/aligner_extender_methods.cpp
@@ -237,10 +237,18 @@ inline void update_del_scores(const DBGAlignerConfig &config,
 
 #ifdef __AVX2__
 
+inline __m128i mm_loadu_si64(const void *mem_addr) {
+    return _mm_loadl_epi64((const __m128i_u*)mem_addr);
+}
+
+inline void mm_storeu_si64(void *mem_addr, __m128i a) {
+    _mm_storel_epi64((__m128i_u*)mem_addr, a);
+}
+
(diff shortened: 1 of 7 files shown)


metagraph/src/seq_io/sequence_io.hpp
```diff
@@ -202,8 +203,33 @@ class FastaParser::iterator : public std::iterator<std::input_iterator_tag,
...
+    }
+
+    void deinit_stream() {
+        if (read_stream_) {
+            gzFile input_p = read_stream_->f->f;
+            kseq_destroy(read_stream_);
+            gzclose(input_p);
+        }
+        read_stream_ = NULL;
+    }
```
hmusta, line 232: this was done to prevent gcc from complaining about kseq_init and kseq_destroy being unused in sequence_io.hpp

TODO: find a better way to do this

karasikov: would something like `(void) kseq_destroy;` solve the problem?

karasikov:
Please add gcc-8 to CI

metagraph/src/seq_io/sequence_io.hpp
hmusta: Looking through `kseq.h`, I found that using `KSEQ_DECLARE` instead of `KSEQ_INIT` fixed the issue

hmusta: The issue now is that `sequence_io.cpp` has a sequence of macros taken from kseq.h to fill in missing declarations

hmusta:
So it looks like this libmaus2 change is causing more problems than expected. How about we merge up to 7554c7f and save the rest for later?

metagraph/src/graph/alignment/aligner_extender_methods.cpp
```diff
@@ -237,10 +237,20 @@ inline void update_del_scores(const DBGAlignerConfig &config,
...
 
+// Drop-in replacement for _mm_loadu_si64
+inline __m128i mm_loadu_si64(const void *mem_addr) {
+    return _mm_loadl_epi64((const __m128i_u*)mem_addr);
+}
+
+// Drop-in replacement for _mm_storeu_si64
+inline void mm_storeu_si64(void *mem_addr, __m128i a) {
+    _mm_storel_epi64((__m128i_u*)mem_addr, a);
+}
```
karasikov, line 248: why not use `_mm_loadu_si64` when it's defined?

karasikov (approved):

metagraph/src/graph/alignment/aligner_extender_methods.cpp
hmusta: I think it's actually better to do it this way. According to the Intel Intrinsics guide, `_mm_loadu_si64` has much higher latency than `_mm_loadl_epi64`. The code I have above is actually how gcc implements `_mm_loadu_si64`.

Plus, implementing a check for `_mm_loadu_si64`s existence would be more annoying since we'd have to do it with cmake.

hmusta: I don't think it's worth further complicating the CMakeLists.txt for just this source file.

PR #249, opened 2020-12-14
Fixes for amino acid DBGs

hmusta:


metagraph/tests/graph/all/test_dbg_traverse.cpp
```diff
@@ -14,6 +14,12 @@ namespace {
 using namespace mtg;
 using namespace mtg::test;
 
+#if _PROTEIN_GRAPH
+const size_t maxK = 12;
+#else
+const size_t maxK = 20;
+#endif
```
karasikov, line 21: MAX_K

metagraph/src/cli/config/config.cpp
```diff
@@ -774,7 +774,11 @@ void Config::print_usage(const std::string &prog_name, IdentityType identity) {
             fprintf(stderr, "\t   --graph [STR] \tgraph representation: succinct / bitmap / hash / hashstr / hashfast [succinct]\n");
             fprintf(stderr, "\t   --count-kmers \tcount k-mers and build weighted graph [off]\n");
             fprintf(stderr, "\t   --count-width \tnumber of bits used to represent k-mer abundance [8]\n");
+#if _PROTEIN_GRAPH
+            fprintf(stderr, "\t   --index-ranges [INT]\tindex all node ranges in BOSS for suffixes of given length [3]\n");
+#else
             fprintf(stderr, "\t   --index-ranges [INT]\tindex all node ranges in BOSS for suffixes of given length [10]\n");
+#endif
```
karasikov, line 781: Better generalize this to `log(2^20)/log(|Sigma|)`, so for DNA with |Sigma| = 4, this will be exactly 10, and 4 for Proteins.

karasikov:
@hmusta please take a look

metagraph/src/graph/representation/bitmap/dbg_bitmap.cpp
```diff
@@ -16,14 +15,14 @@ DBGBitmap::DBGBitmap(size_t k, bool canonical_mode)
       : k_(k),
         canonical_mode_(canonical_mode),
         seq_encoder_(),
-        kmers_(std::pow(static_cast<long double>(seq_encoder_.alphabet.size()), k_) + 1, true),
+        kmers_((1llu << (k * seq_encoder_.bits_per_char)) + 1, true),
```
hmusta, line 18: we should have an assert here to ensure that `seq_encoder_.alphabet.size() == 2`

karasikov: Why? Alphabet size can be arbitrary.

hmusta: Never mind, for some reason I was thinking that `KmerExtractor2Bit` was only implemented for DNA alphabets

karasikov: Yeah, the name is confusing. We should change it later

PR #250, opened 2020-12-16
Remove libmaus2 dependence

hmusta:


metagraph/src/common/serialization.hpp
```diff
@@ -21,6 +31,13 @@ template <typename T>
 bool load_number_vector(std::istream &in, std::vector<T> *vector);
 
 
+void serialize_string(std::ostream &out, const std::string_view str);
```
karasikov, line 34: if it's const, then better pass a const reference, there are no reasons not to.

metagraph/src/common/serialization.cpp
```diff
@@ -329,18 +517,19 @@ bool load_number_string_map(std::istream &in, Map *map) {
         values.reserve(size);
 
         for (size_t i = 0; i < size; ++i) {
-            values.push_back(StringSerialisation::deserialiseString(in));
+            values.emplace_back();
+            load_string(in, &values.back());
```
karasikov, line 521: `load_string` returns `bool`, so you have to check if it's not false.

metagraph/src/common/serialization.cpp
```diff
@@ -350,10 +539,11 @@ bool load_number_string_map(std::istream &in, Map *map) {
...
 
         try {
-            size_t const num = NumberSerialisation::deserialiseNumber(in);
+            size_t const num = load_number(in);
             for (size_t i = 0; i < num; ++i) {
-                auto key = NumberSerialisation::deserialiseNumber32(in);
-                std::string value = StringSerialisation::deserialiseString(in);
+                auto key = load_number32(in);
+                std::string value;
+                load_string(in, &value);
```
karasikov, line 546: check the return value or make it throw exceptions instead of returning `bool`

karasikov: same in other similar cases...

metagraph/src/common/serialization.cpp
```diff
@@ -106,6 +181,117 @@ template bool load_number_vector<uint8_t>(std::istream &in, std::vector<uint8_t>
 template bool load_number_vector<bool>(std::istream &in, std::vector<bool> *);
 
 
+void encode_utf8(uint32_t num, std::ostream &out) {
```
karasikov, line 184: Could you add that statement "adapted from libmaus2 for backward compatibility" here as well?

karasikov (approved):

metagraph/src/common/serialization.cpp
hmusta: I ended up only keeping `decode_utf8` from libmaus2, everything else has been implemented differently.

PR #251, opened 2020-12-17
keep the topological order of k-mers in batch query

karasikov:
RowDiff<> is implemented in the most efficient way to query a batch of rows.
However, if there are multiple query batches (as in construction of large query graphs), we should additionally make sure the neighboring k-mers go to the same batches and not randomly, as it was happening before.

This PR implements this and makes the query performance in practice almost the same as with a single batch of unlimited size (the fastest possible case), while still using multiple batches to limit the RAM overhead during query.


metagraph/src/cli/query.cpp
```diff
@@ -263,29 +263,27 @@ void call_hull_sequences(const DeBruijnGraph &full_dbg,
...
         // shortcut construction for Rainbow<> annotation
-        std::vector<uint64_t> row_indexes;
-        row_indexes.reserve(index_in_full.size());
-        for (uint64_t i : index_in_full) {
-            if (i != npos) {
-                row_indexes.push_back(i);
-            } else {
-                row_indexes.push_back(0);
-            }
+        std::vector<bool> empty_rows(num_rows, true);
```
hmusta, line 280: Why not just initialize `row_indices` to `full_annotation.num_objects()` and use that to indicate empty rows?

hmusta (approved):

metagraph/src/cli/query.cpp
karasikov: `get_rows` requires that all indexes in `row_indices` are valid.

PR #252, opened 2020-12-17
performance optimizations in RowDiff transform

karasikov:


danieldanciu (approved):

metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -166,10 +166,56 @@ using CallOnes = std::function<void(const bit_vector &source_col,
...
+        {
+            // read predecessor offsets
+            pred_chunk_idx.resize(block_size + 1);
+            pred_chunk_idx[0] = 0;
+            for (uint64_t i = 1; i <= block_size; ++i) {
+                // find where the last predecessor for the node ends
+                pred_chunk_idx[i] = pred_chunk_idx[i - 1];
+                while (*pred_boundary_it == 0) {
+                    ++pred_chunk_idx[i];
+                    ++pred_boundary_it;
```
hmusta, line 202: what happens here if the file pointed to by `pred_boundary_it` is truncated? does calling `operator++` throw an exception if this tries to go beyond the end?

metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -363,9 +404,10 @@ void convert_batch_to_row_diff(const std::string &pred_succ_fprefix,
 
     // total number of set bits in the original rows
     std::vector<uint32_t> row_nbits_block;
+    std::vector<uint32_t> row_nbits_block_other;
```
hmusta, line 407: add a comment to describe what this is supposed to be storing any how it's related to `row_nbits_block`

hmusta (approved):

metagraph/src/annotation/row_diff_builder.cpp
karasikov: It will pad with zeros, so this will be an infinite loop. Looks quite safe.

karasikov: I added an explicit check.

PR #254, opened 2020-12-18
Split chunks before merging and make everything parallel

karasikov:
1. First, the original chunks are split into blocks (currently, only by F and W), in parallel
2. Then, each block for a pair (F, W) is built by merging its respective chunks
3. If reverse complement k-mers have to be added, they are generated for each block (F, W), in parallel and the blocks are updated
4. The rest goes as before.

Also minor cleanups

### DIFF
--- a/metagraph/src/common/sorted_sets/sorted_set_disk_base.cpp
+++ b/metagraph/src/common/sorted_sets/sorted_set_disk_base.cpp
@@ -122,15 +122,16 @@ void SortedSetDiskBase<T>::dump_to_file(bool is_done) {
         encoder.add(v);
     }
     total_chunk_size_bytes_ += encoder.finish();
+    chunk_count_++;
+
     data_.resize(0);
+
     if (is_done) {
         async_merge_l1_.remove_waiting_tasks();
     } else if (total_chunk_size_bytes_ > max_disk_space_bytes_) {
         async_merge_l1_.remove_waiting_tasks();
-        async_merge_l1_.join();
         std::string all_merged_file = merged_all_name(chunk_file_prefix_, merged_all_count_);
         // increment chunk_count, so that get_file_names() returns correct values
-        chunk_count_++;
         merge_all(all_merged_file, get_file_names());
 
         total_chunk_size_bytes_ = std::filesystem::file_size(all_merged_file);
@@ -141,12 +142,10 @@ void SortedSetDiskBase<T>::dump_to_file(bool is_done) {
         merged_all_count_++;
         chunk_count_ = 0;
         l1_chunk_count_ = 0;
-        return;
-    } else if (merge_count_ > 0 && (chunk_count_ + 1) % merge_count_ == 0) {
-        async_merge_l1_.enqueue(merge_l1, chunk_file_prefix_, chunk_count_, merge_count_,
-                                &l1_chunk_count_, &total_chunk_size_bytes_);
+    } else if (merge_count_ > 0 && chunk_count_ % merge_count_ == 0) {
+        async_merge_l1_.enqueue(merge_l1, chunk_file_prefix_, chunk_count_ - merge_count_,
+                                merge_count_, &l1_chunk_count_, &total_chunk_size_bytes_);
     }
-    chunk_count_++;
 }
 
 template <typename T>
@@ -187,17 +186,17 @@ void SortedSetDiskBase<T>::try_reserve(size_t size, size_t min_size) {
 
 template <typename T>
 void SortedSetDiskBase<T>::merge_l1(const std::string &chunk_file_prefix,
-                                    uint32_t chunk_count,
+                                    uint32_t chunk,
                                     size_t merge_count,
                                     std::atomic<uint32_t> *l1_chunk_count,
                                     std::atomic<size_t> *total_size) {
     const std::string &merged_l1_file_name
-            = SortedSetDiskBase<T>::merged_l1_name(chunk_file_prefix,
-                                                   chunk_count / merge_count);
-    std::vector<std::string> to_merge(merge_count);
+            = merged_l1_name(chunk_file_prefix, chunk / merge_count);
+
+    std::vector<std::string> to_merge;
     for (uint32_t i = 0; i < merge_count; ++i) {
-        to_merge[i] = chunk_file_prefix + std::to_string(chunk_count - i);
-        *total_size -= static_cast<int64_t>(std::filesystem::file_size(to_merge[i]));
+        to_merge.push_back(chunk_file_prefix + std::to_string(chunk + i));
+        *total_size -= static_cast<int64_t>(std::filesystem::file_size(to_merge.back()));
     }
     logger->trace("Starting merging last {} chunks into {}", merge_count,
                   merged_l1_file_name);
@@ -207,7 +206,7 @@ void SortedSetDiskBase<T>::merge_l1(const std::string &chunk_file_prefix,
     merge_files(to_merge, on_new_item);
     encoder.finish();
 
-    (*l1_chunk_count)++;
+    *l1_chunk_count += 1;
     logger->trace("Merging last {} chunks into {} done", merge_count,
                   merged_l1_file_name);
     *total_size += std::filesystem::file_size(merged_l1_file_name);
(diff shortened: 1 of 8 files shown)


karasikov:
*It didn't want to compile on linux for some reason (could be a bug in gcc, or something), before I made minor changes to the KMerBOSS class (95596ff).

But all commits are meant to work, so reviewing one by one should be easier.

metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
```diff
@@ -339,41 +355,36 @@ using Encoder = common::EliasFanoEncoderBuffered<T>;
...
-/**
- * Splits #kmers by W (kmer[0]) and F (kmer[k]) into |ALPHABET\{$}|^2 chunks.
- * @tparam T_REAL k-mers over the alphabet without the sentinel character (e.g. ACGT).
- * @return names of the files with the partitioned k-mers
- */
 template <typename T_REAL>
-std::vector<std::string> split(size_t k,
-                               const std::filesystem::path &dir,
-                               const ChunkedWaitQueue<T_REAL> &kmers) {
+std::vector<std::string> split(size_t k, const std::string &chunk_fname) {
```
danieldanciu, line 359: I would rather update than delete the comment? This code is already quite complex and hard to follow. Helping the reader with a brief explanation of what the method does seems like a reasonable thing to do.

danieldanciu (approved):
Did you also test this on Leomed and got a speed improvement? I can see how this will be faster on a machine with a fast disk, but it's worth checking if all the extra disk I/O that this code adds does not affect Leomed, if you haven't done so already.
Otherwise, the changes look great.

metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
```diff
@@ -169,7 +169,7 @@ void add_dummy_source_kmers(size_t k,
 
 template <typename T>
 inline T rev_comp(size_t k, T kmer, const std::vector<TAlphabet> &complement_code) {
-    if constexpr (utils::is_pair_v<T>) {
+    if constexpr(utils::is_pair_v<T>) {
```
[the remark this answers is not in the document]
karasikov: We wrote the clang-format settings not to change the code style but to make it possible to convert code written in a completely different style to something closer to the style of the project.

Unfortunately, due to its limitations, clang-format can't edit according to the original code style.

If you know how to make it not add a space between constexpr in `if`, please change it.

metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
```diff
@@ -33,7 +33,7 @@ using utils::get_first;
 using utils::get_first_type_t;
 using TAlphabet = KmerExtractorBOSS::TAlphabet;
 
-constexpr size_t ENCODER_BUFFER_SIZE = 100'000;
+const size_t ENCODER_BUFFER_SIZE = 100'000;
```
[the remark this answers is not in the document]
karasikov: It's just a usual global const variable, and we use `const ...` for them.

Actually, I believe the compiler comprehends them as constexpr anyway. Besides, this parameter doesn't have to be constexpr, since the compilation doesn't depend on it.

karasikov:
> Did you also test this on Leomed and got a speed improvement? I can see how this will be faster on a machine with a fast disk, but it's worth checking if all the extra disk I/O that this code adds does not affect Leomed, if you haven't done so already.
> Otherwise, the changes look great.

I've just run tests on a large plants canonical graph. The speedup is about 3x on the cluster, and will be only larger for bigger graphs.

karasikov:
@hmusta have you had a chance to take a look?

metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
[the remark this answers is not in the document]
danieldanciu: That was totally not my point. My point is that *you* come up with some arbitrary rules, such as that there should be no space between constexpr and the paren, and then we all have to remember this rule that cannot be enforced by clang-format.

metagraph/src/common/sorted_sets/sorted_set_disk_base.cpp
```diff
@@ -22,7 +22,7 @@ SortedSetDiskBase<T>::SortedSetDiskBase(size_t num_threads,
       merge_count_(merge_count),
       chunk_file_prefix_(tmp_dir/"chunk_"),
       merge_queue_(std::min(reserved_num_elements, QUEUE_EL_COUNT)),
-      async_merge_l1_(merge_count_ == 0 ? 0 : 1, 100) {
+      async_merge_l1_(merge_count_ == 0 ? 0 : 3, 100) {
```
hmusta, line 25: why not put a check here to see if `get_num_threads() > 1`?

metagraph/src/common/sorted_sets/sorted_set_disk_base.cpp
```diff
@@ -207,7 +204,7 @@ void SortedSetDiskBase<T>::merge_l1(const std::string &chunk_file_prefix,
     merge_files(to_merge, on_new_item);
     encoder.finish();
 
-    (*l1_chunk_count)++;
+    *l1_chunk_count += 1;
```
hmusta, line 207: any particular reason why? don't these two use the same memory order anyway?

metagraph/src/common/sorted_sets/sorted_set_disk_base.cpp
```diff
@@ -122,15 +120,16 @@ void SortedSetDiskBase<T>::dump_to_file(bool is_done) {
         encoder.add(v);
     }
     total_chunk_size_bytes_ += encoder.finish();
+    chunk_count_++;
+
     data_.resize(0);
+
     if (is_done) {
         async_merge_l1_.remove_waiting_tasks();
```
hmusta, line 128: this is a bit confusing, which tasks do we expect to still be waiting if `is_done` is true?

metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
```diff
@@ -221,6 +221,20 @@ void add_reverse_complements(size_t k, size_t num_threads, Vector<T> *kmers) {
...
+    static_assert(sizeof(T_TO) == sizeof(T_FROM));
+    return reinterpret_cast<ChunkedWaitQueue<T_TO>&>(container);
+}
+
+template <typename T_TO, typename T_FROM>
+Vector<T_TO>& reinterpret_container(Vector<T_FROM> &container) {
+    static_assert(sizeof(T_TO) == sizeof(T_FROM));
+    return reinterpret_cast<Vector<T_TO>&>(container);
+}
+
```
hmusta, line 236: I know these are not in the public interface, but maybe add a static assert to these to make sure that `T_TO` and `T_FROM` are not `bool` in case this code ends up being used in other places.

hmusta (changes requested):

metagraph/src/common/sorted_sets/sorted_set_disk_base.cpp
karasikov: I thought this looks simpler, no other reasons

metagraph/src/common/sorted_sets/sorted_set_disk_base.cpp
karasikov: The "merge_l1" tasks. If we are about to merge all the chunks anyway, there is little point to do this l1 pre-merge blocking the ultimate merge. This is a minor optimization.

metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
karasikov: done

metagraph/src/common/sorted_sets/sorted_set_disk_base.cpp
karasikov: This will be very slow if done strictly with one thread. It's better to have these workers that mostly do slow IO in the background.

metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
karasikov: done

hmusta (approved):

PR #257, opened 2021-01-03
don't write row reduction vectors in anchor optimization stage

karasikov:
saves disk space and time by not computing and not writing row reduction vectors, which are not needed anyway if the first stage was fully finished


metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -402,6 +405,9 @@ void convert_batch_to_row_diff(const std::string &pred_succ_fprefix,
                 }
             },
             [&](uint64_t block_begin) {
+                if (!compute_row_reduction)
```
danieldanciu, line 408: Isn't it more efficient to just pass an empty lambda if `compute_row_reduction` is false?

danieldanciu (approved):

PR #262, opened 2021-01-07
Restrict alignment extension to labeled subgraphs

hmusta:
1. Store the labels of the seed when an alignment extension starts
2. While the set of labels is maintained, proceed with extension as usual.
3. Given an edge A -> B with label sets LA and LB, if LA is not a subset of LB, then create a new extender with label set LA \intersect LB and replace the label set of the original extender with LA - LB. Continue extending using both extenders.

TODO: unit tests and benchmarking


hmusta:
This makes #150 obsolete.

metagraph/src/graph/alignment/aligner_labeled.hpp
```diff
@@ -0,0 +1,219 @@
...
+    for (const auto &edge : base_edges) {
+        const auto &[out_node, c] = edge;
+        AnnotatedDBG::row_index out_row = anno_graph_.graph_to_anno_index(out_node);
+
+        if (target_columns_.size() == 1) {
+            if (mat.get(out_row, target_columns_[0]))
+                edges.push_back(edge);
+
+        } else {
+            auto row = mat.get_row(out_row);
```
hmusta, line 181: This kind of query against the binary matrix is done with the assumption that the annotator already does some kind of result prefetching (like with RowDiff).

TODO: maybe implement a prefetching technique to work for the other BinaryMatrix types

hmusta: ah....it looks like this is only done within a `get_rows` call and not across multiple calls, so this should be implemented here to speed things up

metagraph/src/graph/alignment/aligner_labeled.hpp
```diff
@@ -0,0 +1,241 @@
...
+        return edges;
+    }
+
+    // decode all rows
+    std::vector<AnnotatedDBG::row_index> base_rows;
+    base_rows.reserve(base_edges.size());
+    for (const auto &[out_node, c] : base_edges) {
+        base_rows.push_back(anno_graph_.graph_to_anno_index(out_node));
+    }
+    auto rows = mat.get_rows(base_rows);
```
hmusta, line 178: TODO: cache this

metagraph/src/graph/alignment/aligner_labeled.hpp
```diff
@@ -0,0 +1,275 @@
...
+                const auto &dbg_succ = *rd->graph();
+                const auto &boss = dbg_succ.get_boss();
+                if (boss.get_last(dbg_succ.kmer_to_boss_index(base_edges[i].first))) {
+                    // apply diff to target_columns_
+                    auto diff_row = rd->get_diff(anno_graph_.graph_to_anno_index(node));
+                    std::set_difference(target_columns_.begin(), target_columns_.end(),
+                                        diff_row.begin(), diff_row.end(),
+                                        std::back_inserter(rows[i]));
+                } else {
+                    rows[i] = mat.get_row(base_rows[i]);
```
karasikov, line 203: always avoid `get_row`. It's inefficient in most of the cases. Always do pre-batching and then call `get_rows`.

karasikov (changes requested):

metagraph/src/graph/alignment/aligner_labeled.hpp
hmusta: Doing a batched query here doesn't actually speed things up that much. Each batch would only be size 1-2 on average.

hmusta:
> This call can make the entire thing slower.
> So this optimization is questionable but much more complex than `get_rows`.
> I think it's better to call `get_rows`, if you're to choose between these two.
> 
> Or, for proper optimization, write a special function for querying annotations of pre-fetched contiguous paths.

I think I'll take the approach of writing a special function for this. When benchmarking with the MetaSub read set against the BIGSI 750 graph, this optimization led to a 20% run time improvement.

hmusta:
> > This call can make the entire thing slower.
> > So this optimization is questionable but much more complex than `get_rows`.
> > I think it's better to call `get_rows`, if you're to choose between these two.
> > Or, for proper optimization, write a special function for querying annotations of pre-fetched contiguous paths.
> 
> I think I'll take the approach of writing a special function for this. When benchmarking with the MetaSub read set against the BIGSI 750 graph, this optimization led to a 20% run time improvement.

The approach of aligning along a path first, then querying contiguous paths is a bit problematic. If it turns out that the path you took did not match your label, then there would have to be some kind of backtracking approach to undo the changes to the dynamic programming table.

karasikov:
> > This call can make the entire thing slower.
> > So this optimization is questionable but much more complex than `get_rows`.
> > I think it's better to call `get_rows`, if you're to choose between these two.
> > Or, for proper optimization, write a special function for querying annotations of pre-fetched contiguous paths.
> 
> I think I'll take the approach of writing a special function for this. When benchmarking with the MetaSub read set against the BIGSI 750 graph, this optimization led to a 20% run time improvement.

Try on the full graph or a larger graph. The 750-subset has the smallest annotation. For larger annotations the situation will be different.
Roughly, the annotation query time grows linearly with annotation size, while for graph it grows sublinearly and much slower.
So here you trade redundant graph query operations for annotation query, which is not beneficial for graphs with large annotation.

PR #263, opened 2021-01-08
Merge F-W chunks and build BOSS blocks separately in parallel

karasikov:
* speed up the final merge by merging real k-mers in parallel beforehand
* generate dummy source k-mers in parallel
* construct final BOSS chunks in parallel

This should make the last stage of graph construction 4 times faster

### DIFF
--- a/metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
+++ b/metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
@@ -927,6 +927,7 @@ BOSS::Chunk* build_boss(const std::vector<std::string> &real_names,
     }
 
     // concatenate all chunks
+    logger->trace("Concatenating {} chunks into a single BOSS table", chunks.size() + 1);
     for (size_t F = 0; F < real_names.size(); ++F) {
         chunk->extend(*chunks[F]);
         delete chunks[F];
--- a/metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
+++ b/metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
@@ -405,11 +405,11 @@ void skip_same_suffix(const KMER &el, Decoder &decoder, size_t suf) {
 
 /**
  * Generates non-redundant dummy-1 source k-mers and dummy sink kmers from #kmers.
- * @return a triplet containing the names of the original k-mer blocks, the dummy-1 source
+ * @return a triplet containing the names of the original k-mer block, the dummy-1 source
  * k-mer blocks and the dummy sink k-mers
  */
 template <typename T_REAL, typename T>
-std::tuple<std::vector<std::string>, std::vector<std::string>, std::string>
+std::tuple<std::string, std::vector<std::string>, std::string>
 generate_dummy_1_kmers(size_t k,
                        size_t num_threads,
                        const std::filesystem::path &dir,
@@ -422,13 +422,19 @@ generate_dummy_1_kmers(size_t k,
 
     const uint8_t alphabet_size = KmerExtractor2Bit().alphabet.size();
 
+    using T_INT = get_int_t<T>;
+    using T_INT_REAL = get_int_t<T_REAL>;
+    std::vector<Encoder<T_INT>> real_chunks;
     std::vector<Encoder<KMER_INT>> dummy_l1_chunks;
     std::vector<Encoder<KMER_INT>> dummy_sink_chunks;
+    std::vector<std::string> real_split_by_F_names(alphabet_size);
     std::vector<std::string> dummy_l1_names(alphabet_size);
     std::vector<std::string> dummy_sink_names(alphabet_size);
     for (TAlphabet i = 0; i < alphabet_size; ++i) {
+        real_split_by_F_names[i] = dir/("real_split_by_F_" + std::to_string(i));
         dummy_l1_names[i] = dir/("dummy_source_1_" + std::to_string(i));
         dummy_sink_names[i] = dir/("dummy_sink_" + std::to_string(i));
+        real_chunks.emplace_back(real_split_by_F_names[i], ENCODER_BUFFER_SIZE);
         dummy_l1_chunks.emplace_back(dummy_l1_names[i], ENCODER_BUFFER_SIZE);
         dummy_sink_chunks.emplace_back(dummy_sink_names[i], ENCODER_BUFFER_SIZE);
     }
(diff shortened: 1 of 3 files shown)


metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
```diff
@@ -404,12 +409,17 @@ void skip_same_suffix(const KMER &el, Decoder &decoder, size_t suf) {
 }
 
 /**
- * Generates non-redundant dummy-1 source k-mers and dummy sink kmers from #kmers.
- * @return a triplet containing the names of the original k-mer blocks, the dummy-1 source
- * k-mer blocks and the dummy sink k-mers
+ * Generates ordered blocks of k-mers for the BOSS table.
+ * Adds non-redundant dummy-1 source k-mers and dummy sink k-mers.
```
hmusta, line 413: I'm assuming "dummy-1 source k-mers" means every dummy k-mer except $$$$$$..., right?

metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
```diff
@@ -489,39 +525,25 @@ generate_dummy_1_kmers(size_t k,
         }
     }
 
+    // remove all unmerged chunks
+    common::remove_chunks(real_F_W);
+
     uint64_t num_sink = 0;
     uint64_t num_source = 0;
     for (TAlphabet i = 0; i < alphabet_size; ++i) {
+        real_chunks[i].finish();
```
hmusta, line 534: The ordering here is a bit confusing to me. Since `finish()` is not called until after `remove_chunks`, isn't there a chance that a chunk may be removed before it's merged?

metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
```diff
@@ -673,114 +710,225 @@ void recover_dummy_nodes_disk(KmerCollector &kmer_collector,
...
             }
-            num_kmers++;
-        };
-        common::merge_files(dummy_names, write_dummy, false);
 
-        std::for_each(dummy_next_chunks.begin(), dummy_next_chunks.end(),
-                      [](auto &v) { v.finish(); });
-        dummy_names = std::move(dummy_next_names);
+            KMER prev_kmer(0);
+            const std::function<void(const KMER_INT &)> &write_dummy = [&](const KMER_INT &v) {
```
hmusta, line 765:
```suggestion
            const auto &write_dummy = [&](const KMER_INT &v) {
```
Is the explicit type declaration needed here? I think it's clear from context.

metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
```diff
@@ -673,114 +710,225 @@ void recover_dummy_nodes_disk(KmerCollector &kmer_collector,
...
+            std::vector<std::string> F_chunk_names(names.begin() + F * alphabet_size,
+                                                   names.begin() + (F + 1) * alphabet_size);
+            common::merge_files(F_chunk_names, write_dummy, true);
+
+            dummy_chunk.finish();
+            std::for_each(dummy_next_chunks.begin(), dummy_next_chunks.end(),
+                          [](auto &v) { v.finish(); });
+
+            #pragma omp atomic
+            num_kmers += dummy_chunk.size();
```
hmusta, line 787: Does declaring `num_kmers` to be `std::atomic<uint64_t>` make this slower?

metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
```diff
@@ -673,114 +710,225 @@ void recover_dummy_nodes_disk(KmerCollector &kmer_collector,
...
+                                             k, both_strands_mode, bits_per_count, swap_dir);
+    logger->trace("Chunk for ..$. done");
+    // construct all other chunks in parallel
+    #pragma omp parallel for ordered num_threads(num_threads) schedule(dynamic)
+    for (size_t F = 0; F < real_names.size(); ++F) {
+        std::vector<std::string> dummy_names { dummy_sink_names[F] };
+        for (size_t i = 0; i < dummy_source_names.size(); ++i) {
+            dummy_names.push_back(dummy_source_names[i][F + 1]);
+        }
+        BOSS::Chunk *next = build_boss_chunk<T>(false, real_names[F], dummy_names,
```
hmusta, line 923: at some point, we should have a TODO to make these return smart pointers instead

hmusta, line 923:
```suggestion
        BOSS::Chunk *next = build_boss_chunk<T>(false /* is_first_chunk*/ ,
                                                real_names[F], dummy_names,
```

hmusta (changes requested):

metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
karasikov: no, only with a single sentinel in the beginning: $AAAAAAA

metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
karasikov: all chunks have been read by this point, so we can safely remove them. `finish()` is called to flush the output buffers, when no more reading is done at this point.

metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
karasikov: It's required to make the next function infer the type correctly.

metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
karasikov: I didn't try, but I'm sure either way is fine here

metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
hmusta: ah, right..never mind

metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
```diff
@@ -673,114 +710,225 @@ void recover_dummy_nodes_disk(KmerCollector &kmer_collector,
...
             }
         }
-        while (!decoder.empty()) {
-            kmers_out->push(decoder.pop());
+        while (next) {
+            kmers_out->push(next.value());
+            next = decoder.next();
         }
         while (!decoder_dummy.empty()) {
             kmers_out->push(decoder_dummy.pop());
```
hmusta, line 852: This is optional, but do you think this can be made simpler with `std::merge` if the right interface is there?

metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
```diff
@@ -673,114 +710,225 @@ void recover_dummy_nodes_disk(KmerCollector &kmer_collector,
...
+                        const std::vector<std::string> &dummy_sink_names,
+                        std::filesystem::path dir,
+                        size_t k,
+                        bool both_strands_mode,
+                        uint8_t bits_per_count,
+                        std::filesystem::path swap_dir,
+                        size_t num_threads) {
+    assert(real_names.size() == dummy_sink_names.size());
+    assert(real_names.size() + 1 == dummy_source_names.front().size());
+    assert(dummy_source_names.size() == k);
```
hmusta, line 876: Maybe also add an assert to ensure that `dummy_source_names` is already sorted?

hmusta (approved):

metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
karasikov: Yeah, would be good, but it requires iterators and decoders have a different interface

PR #267, opened 2021-01-13
annotate with disk swap

karasikov:
pass `--disk-swap ...` to reduce the RAM usage when annotating very large columns


metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.hpp
```diff
@@ -27,7 +27,9 @@ class ColumnCompressed : public MultiLabelEncoded<Label> {
     using VLabels = typename MultiLabelEncoded<Label>::VLabels;
 
     ColumnCompressed(uint64_t num_rows = 0,
-                     size_t num_columns_cached = 1);
+                     size_t num_columns_cached = 1,
+                     const std::string &swap_dir = "",
+                     uint64_t buffer_size = 1024 * 1024 * 1024 / sizeof(uint64_t));
```
hmusta, line 32: Even though it's clear from the default value that `buffer_size` is in GiB, maybe rename this to `buffer_size_gb` or modify the code to use bytes as units. The latter is more standard and could avoid confusion later.

metagraph/src/common/vectors/bitmap_builder.hpp
```diff
@@ -75,4 +78,35 @@ class bitmap_builder_set : public bitmap_builder {
...
 
 
+class bitmap_builder_set_disk : public bitmap_builder {
+  public:
+    bitmap_builder_set_disk(uint64_t size,
+                            size_t num_threads,
+                            uint64_t buffer_size,
+                            const std::string &swap_dir,
+                            size_t max_disk_space_bytes = -1,
+                            size_t merge_count = 4)
```
hmusta, line 88: It's not clear what `merge_count` is supposed to be. Add a comment

metagraph/src/common/vectors/bitmap_builder.cpp
```diff
@@ -0,0 +1,37 @@
...
+bitmap_builder_set_disk::InitializationData
+bitmap_builder_set_disk::get_initialization_data() const {
+    if (merged_)
+        throw std::runtime_error("ERROR: Trying to flush the bitmap twice");
+
+    mtg::common::EliasFanoEncoderBuffered<uint64_t> encoder(tmp_dir_/"merged",
+                                                            kStreamBufferSize);
+    auto &merged = const_cast<bitmap_builder_set_disk*>(this)->set_bit_positions_.data();
+    uint64_t num_set_bits = 0;
+    for (auto &it = merged.begin(); it != merged.end(); ++it) {
```
hmusta, line 18: This `const_cast` looks kind of ugly and is unnecessary since you're not using move iterators. Maybe add a `const_iterator` to this class to avoid it?

metagraph/src/common/vectors/bitmap_builder.cpp
```diff
@@ -0,0 +1,37 @@
...
+        assert(*it < size_ && "Indexes cannot be greater than bitmap's size");
+        encoder.add(*it);
+        num_set_bits++;
+    }
+    encoder.finish();
+
+    const_cast<bitmap_builder_set_disk*>(this)->set_bit_positions_.clear();
+
+    auto call_indexes = [&](auto callback) {
+        mtg::common::EliasFanoDecoder<uint64_t> decoder(tmp_dir_/"merged", true);
```
hmusta, line 28: Maybe register a new tmp directory just in case `bitmap_builder_set_disk` is destroyed before the callback is called.

hmusta (changes requested):

metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.hpp
karasikov: Good point, it's really unclear what the units are...

metagraph/src/common/vectors/bitmap_builder.hpp
karasikov: I'll just make it a constant. I doubt anyone will change this parameter.

metagraph/src/common/vectors/bitmap_builder.cpp
karasikov: ChunkedWaitQueue hasn't got a const iterator. That's why we need to cast.

hmusta: why not add one? it's only a few lines of boilerplate code

karasikov: Because it actually does change the state of the queue, hence it's not const.
One needs to move the merging logic into the iterator in order to fix this, and that's more work.

metagraph/src/common/vectors/bitmap_builder.hpp
```diff
@@ -75,4 +78,35 @@ class bitmap_builder_set : public bitmap_builder {
...
+
+    ~bitmap_builder_set_disk() { std::filesystem::remove_all(tmp_dir_); }
+
+    virtual void add_one(uint64_t pos) { set_bit_positions_.insert(&pos, &pos + 1); }
+    virtual void add_ones(const uint64_t *begin, const uint64_t *end) {
+        set_bit_positions_.insert(begin, end);
+    }
+
+    // can only be called once
+    virtual InitializationData get_initialization_data() const;
```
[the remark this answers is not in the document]
karasikov: It was const for all other builders, and I wanted to stick to the same interface.
But yeah, maybe it's just better to switch all to non-const, because of this one.

metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.hpp
karasikov: I added a comment

metagraph/src/common/vectors/bitmap_builder.cpp
karasikov: No, actually, it was assumed that bitmap_builder can be destroyed only after the callback is called.
This doesn't restrict the usage in general but makes things easier.

The same assumption is applied for the other bitmap_builder without disk swap.

PR #272, opened 2021-01-21
remove temp directory after it's not needed, before exit()

karasikov:
A small change, makes it remove temp directories right after they are not needed and not wait until the program terminates.
So it doesn't show a long list of `Cleaning up temporary directory ...` in the end.

### DIFF
--- a/metagraph/src/annotation/row_diff_builder.cpp
+++ b/metagraph/src/annotation/row_diff_builder.cpp
@@ -302,7 +302,6 @@ void convert_batch_to_row_diff(const std::string &pred_succ_fprefix,
 
     const std::filesystem::path tmp_path = utils::create_temp_dir(
             std::filesystem::path(dest_dir).remove_filename(), "col");
-    logger->trace("Using temporary directory {}", tmp_path);
 
     // stores the row indices that were set because of differences to incoming/outgoing
     // edges, for each of the sources, per chunk. set_rows_fwd is already sorted
@@ -543,8 +542,8 @@ void convert_batch_to_row_diff(const std::string &pred_succ_fprefix,
 
         logger->trace("Serialized {}", fpath);
     }
-    logger->trace("Removing temp directory: {}", tmp_path);
-    std::filesystem::remove_all(tmp_path);
+
+    utils::remove_temp_dir(tmp_path);
 
     if (!compute_row_reduction)
         return;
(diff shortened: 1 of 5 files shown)


metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -302,7 +302,6 @@ void convert_batch_to_row_diff(const std::string &pred_succ_fprefix,
 
     const std::filesystem::path tmp_path = utils::create_temp_dir(
             std::filesystem::path(dest_dir).remove_filename(), "col");
-    logger->trace("Using temporary directory {}", tmp_path);
```
hmusta, line 305: Instead of removing this, why not wrap it around an `#ifndef NDEBUG ... #endif` block instead?

metagraph/src/common/utils/file_utils.cpp
```diff
@@ -19,24 +19,27 @@ namespace utils {
...
-            std::filesystem::remove_all(tmp_dir);
-        } catch (...) {
-            logger->error("Failed cleaning up temporary directory {}", tmp_dir);
-        }
+void cleanup_temp_dir_nolock(const std::filesystem::path &tmp_dir) {
+    logger->trace("Cleaning up temporary directory {}", tmp_dir);
+    try {
+        std::filesystem::remove_all(tmp_dir);
+    } catch (...) {
+        logger->error("Failed to clean up temporary directory {}", tmp_dir);
```
hmusta, line 35: Won't this lead to exit code 0 when there's a filesystem error? This isn't ideal...

hmusta (changes requested):

metagraph/src/annotation/row_diff_builder.cpp
karasikov: `utils::create_temp_dir` already prints a line in the log that it created a new temp dir, so I figured this second message is redundant.

metagraph/src/common/utils/file_utils.cpp
hmusta: Maybe within the `catch` block you can store the exception in a `std::exception_ptr` and rethrow it after all cleanup is done.

karasikov: Hm, I don't want to throw exceptions here.
It's called either in a signal callback or as a normal single cleanup call.
In both cases, if it failed to remove the temp dir for some reason, it may still finish the rest without issues.
So, just an error message should be enough for now.

I was more concerned about the mutex and locks, which is not fully thread-safe at the moment, in case of finishing a failed run.
But also, don't know how to make it completely safe. Not sure if acquiring a mutex in a signal callback is a good idea.

PR #275, opened 2021-01-22
Alignment fixes

hmusta:


metagraph/src/graph/alignment/aligner_helper.cpp
```diff
@@ -273,8 +273,11 @@ Alignment<NodeType>::Alignment(const DPTable<NodeType> &dp_table,
 
                 // assert(column->second.prev_nodes.at(i + 1) == 0xFF);
                 // assert(column->second.scores.at(i + 1) >= score_track);
-                assert(column->second.ops.at(i - shift) != Cigar::INSERTION);
-                score_track -= column->second.ops.at(i - shift) == Cigar::DELETION
+                if (column->second.ops.at(i - shift) == Cigar::DELETION)
+                    logger->error("INSERTION after DELETION: {}", query_view);
```
karasikov, line 277: Maybe better `warn`?

karasikov (approved):

metagraph/src/graph/alignment/aligner_helper.cpp
hmusta: In DEBUG mode this leads to an assertion failure. I wanted to add the error message here to make sure it's detected even in RELEASE mode.

karasikov: then write assert

hmusta: The assert was already there before, it's on the next line.

hmusta: This time it got changed since `DELETION` and `INSERTION` are being swapped

karasikov: ah, got it. Yeah, it's better to keep them together for clarity and maybe inside a `#ifndef NDEBUG` block

karasikov: Looking at all this now -- since you run the check in release mode anyway, better do it always and throw exception, for both release and debug. Or logger->error + exit(1), without assert and `#ifdef NDEBUG`

PR #276, opened 2021-01-22
move constructors for graph chunks

karasikov:
Refactoring in graph chunks: use move constructors instead of passing pointers

### DIFF
--- a/metagraph/src/graph/representation/succinct/boss_chunk.cpp
+++ b/metagraph/src/graph/representation/succinct/boss_chunk.cpp
@@ -138,7 +138,8 @@ BOSS::Chunk::Chunk(uint64_t alph_size, size_t k, bool canonical,
 }
 
 BOSS::Chunk::~Chunk() {
-    utils::remove_temp_dir(dir_);
+    if (!dir_.empty())
+        utils::remove_temp_dir(dir_);
 }
 
 template <typename Array>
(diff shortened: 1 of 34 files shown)


metagraph/src/graph/representation/succinct/boss_chunk.cpp
```diff
@@ -138,7 +138,8 @@ BOSS::Chunk::Chunk(uint64_t alph_size, size_t k, bool canonical,
 }
 
 BOSS::Chunk::~Chunk() {
-    utils::remove_temp_dir(dir_);
+    if (!dir_.empty())
+        utils::remove_temp_dir(dir_);
```
hmusta, line 142: So I'm guessing the logic here is that if the tmp directory is corrupted somehow you want to keep it? Maybe clarify this with a comment.

metagraph/src/graph/representation/succinct/boss_merge.cpp
```diff
@@ -179,27 +179,23 @@ BOSS::Chunk* merge_blocks_to_chunk(const std::vector<const BOSS*> &graphs,
     if (verbose)
         print_bin_stats(bins);
 
-    std::vector<BOSS::Chunk*> blocks(bins.front().size() - 1, NULL);
```
hmusta, line 182: I guess you weren't using this vector before anyway? The only reference to it that I see is getting its size later.

hmusta (approved):

metagraph/src/graph/representation/succinct/boss_chunk.cpp
karasikov: Not exactly. Now we can have uninitialized chunks, where dir_ is empty, and thus can't call blindly call `utils::remove_temp_dir` which can only be called for directories created by `utils::create_temp_dir`.

metagraph/src/graph/representation/succinct/boss_merge.cpp
karasikov: Yeah, this was redundant, I noticed this when was changing these lines.

metagraph/src/graph/representation/succinct/boss_chunk.cpp
karasikov: I added a comment

PR #277, opened 2021-01-22
Improve traversal operations for CanonicalDBG wrapped DBGSuccinct. Only allow for use on primary graphs

hmusta:


metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -125,6 +125,58 @@ void CanonicalDBG::map_to_nodes(std::string_view sequence,
     }
 }
 
+void CanonicalDBG
+::get_kmers_from_suffix(node_index node, std::vector<node_index> &children) const {
```
karasikov, line 129:
```suggestion
void CanonicalDBG::get_kmers_from_suffix(node_index node,
                                         std::vector<node_index> &children) const {
```

karasikov: describe how it modifies `children` in a comment

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -125,6 +125,58 @@ void CanonicalDBG::map_to_nodes(std::string_view sequence,
...
+            get_k() - 1
+        );
+
+    } else {
+        for (size_t i = 0; i < alphabet.size(); ++i) {
+            if (children[i] != DeBruijnGraph::npos)
+                continue;
+
+            char c = alphabet[i];
+            ::reverse_complement(&c, &c + 1);
```
karasikov, line 166: Add another function
```suggestion
            char c = ::reverse_complement(alphabet[i]);
```

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -125,6 +125,58 @@ void CanonicalDBG::map_to_nodes(std::string_view sequence,
...
+                char c = boss.decode(boss.get_minus_k_value(edge, get_k() - 2).first);
+                if (c == boss::BOSS::kSentinel)
+                    return;
+
+                ::reverse_complement(&c, &c + 1);
+                size_t i = boss.encode(c);
+
+                if (children[i] == DeBruijnGraph::npos) {
+                    if (!primary_)
+                        rev_comp_cache_.Put(next, next + offset_);
```
karasikov, line 152: Maybe add a comment why you put these to the cache and how you expect it to make things faster.
Like, you expect these nodes to be called soon after this function, or something?

If there is no reason, it's better not to change the cache.

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -125,6 +125,58 @@ void CanonicalDBG::map_to_nodes(std::string_view sequence,
     }
 }
 
+void CanonicalDBG
+::get_kmers_from_suffix(node_index node, std::vector<node_index> &children) const {
+    const auto &alphabet = graph_.alphabet();
+    node_index next;
```
karasikov, line 131: remove it from here
```suggestion
```

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -125,6 +125,58 @@ void CanonicalDBG::map_to_nodes(std::string_view sequence,
...
+    } else {
+        for (size_t i = 0; i < alphabet.size(); ++i) {
+            if (children[i] != DeBruijnGraph::npos)
+                continue;
+
+            char c = alphabet[i];
+            ::reverse_complement(&c, &c + 1);
+
+            rev_seq[0] = c;
+            next = graph_.kmer_to_node(rev_seq);
```
karasikov, line 169:
```suggestion
            node_index next = graph_.kmer_to_node(rev_seq);
```

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -125,6 +125,58 @@ void CanonicalDBG::map_to_nodes(std::string_view sequence,
...
+    node_index next;
+    std::string rev_seq = get_node_sequence(node).substr(1) + std::string(1, '\0');
+    ::reverse_complement(rev_seq.begin(), rev_seq.end());
+    assert(rev_seq[0] == '\0');
+
+    const auto *dbg_succ = dynamic_cast<const DBGSuccinct*>(&graph_);
+    if (dbg_succ) {
+        const auto &boss = dbg_succ->get_boss();
+        dbg_succ->call_nodes_with_suffix_matching_longest_prefix(
+            std::string_view(&rev_seq[1], get_k() - 1),
```
karasikov, line 140: write a comment about what `rev_seq` represents

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -125,6 +125,58 @@ void CanonicalDBG::map_to_nodes(std::string_view sequence,
...
+            [&](node_index next, uint64_t /* match length */) {
+                auto edge = dbg_succ->kmer_to_boss_index(next);
+                char c = boss.decode(boss.get_minus_k_value(edge, get_k() - 2).first);
+                if (c == boss::BOSS::kSentinel)
+                    return;
+
+                ::reverse_complement(&c, &c + 1);
```
karasikov, line 147: c = ::reverse_complement(c);

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -191,6 +223,64 @@ ::call_outgoing_kmers(node_index node, const OutgoingEdgeCallback &callback) con
...
+
+                    if (parents[i] == DeBruijnGraph::npos) {
+                        if (!primary_)
+                            rev_comp_cache_.Put(prev, prev + offset_);
+
+                        parents[i] = prev + offset_;
+                    }
+                }
+
+            } while (--edge && !boss.get_last(edge));
```
karasikov, line 260: no new line
```suggestion
                }
            } while (--edge && !boss.get_last(edge));
```

karasikov, line 260: I think there should be a routine like `call_outgoing` in BOSS for this do-while loop.

karasikov (changes requested):

metagraph/src/graph/representation/canonical_dbg.cpp
hmusta: There isn't, so I added one

metagraph/src/graph/representation/canonical_dbg.cpp
hmusta: There's already a cache for storing the children and parents of each given node, so this is redundant

metagraph/tests/graph/test_canonical_dbg.cpp
```diff
@@ -574,96 +574,6 @@ TYPED_TEST(CanonicalDBGTest, CallUnitigs) {
...
-                    }
-                );
-
-                EXPECT_EQ(*stable_graph, *reconstructed_stable_graph);
-            }
-        }
-    }
-}
-
-TYPED_TEST(CanonicalDBGTest, CallUnitigsRegularCanonical) {
```
hmusta, line 621: TODO: replace this with a death test

metagraph/src/graph/representation/canonical_dbg.hpp
```diff
@@ -172,17 +156,19 @@ class CanonicalDBG : public DeBruijnGraph {
...
 
     node_index reverse_complement(node_index node) const;
+
+    // find all parent nodes of node which are only represented in their reverse
+    // complement orientation in the underlying graph
+    void append_parent_nodes_using_node_rev_comp(node_index node, std::vector<node_index> &parents) const;
+
+    // find all child nodes of node which are only represented in their reverse
+    // complement orientation in the underlying graph
+    void append_child_nodes_using_node_rev_comp(node_index node, std::vector<node_index> &children) const;
```
karasikov, line 171: `which are only represented ...` is redundant. THe graph is primary, hence they can't be represented in both orientations in the underlying graph.

Maybe better renmae these methods to `append_next_rc_nodes`, and `append_prev_rc_nodes`?

metagraph/src/graph/representation/canonical_dbg.hpp
```diff
@@ -22,43 +22,29 @@ class CanonicalDBG : public DeBruijnGraph {
...
      * Constructs a CanonicalDBG
      * @param graph a pointer to the graph
-     * @param primary indicates whether the underlying graph is primary or not
-     * (i.e., only one of a k-mer and its reverse complement are in the graph)
      * @param cache_size the number of graph traversal call results to be cached
      */
-    CanonicalDBG(std::shared_ptr<DeBruijnGraph> graph,
-                 bool primary = false,
-                 size_t cache_size = 100'000);
+    CanonicalDBG(std::shared_ptr<DeBruijnGraph> graph, size_t cache_size = 100'000);
```
karasikov, line 47: it would be good to benchmarks the speed for different cache size at some point.

Maybe it never makes much sense to make the cache size larger than 100k, for example, and then we can simply set that as a global constant.

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -380,36 +441,18 @@ DeBruijnGraph::node_index CanonicalDBG::reverse_complement(node_index node) cons
     if (node > offset_) {
         // we know that this node is definitely not present in the base graph
 
-        if (!primary_) {
-            rev_comp_cache_.Put(node - offset_, node);
-        } else if (!(graph_.get_k() & 1)) {
+        if (!(graph_.get_k() & 1))
```
karasikov, line 444:
```suggestion
        if (graph_.get_k() % 2 == 0)
```

karasikov: Looks more natural to me

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -191,13 +212,75 @@ ::call_outgoing_kmers(node_index node, const OutgoingEdgeCallback &callback) con
...
+}
+
 void CanonicalDBG
 ::call_incoming_kmers(node_index node, const IncomingEdgeCallback &callback) const {
     assert(node);
     assert(node <= offset_ * 2);
     if (node > offset_) {
         call_outgoing_kmers(node - offset_, [&](node_index prev, char c) {
-            ::reverse_complement(&c, &c + 1);
+            c = ::reverse_complement(c);
```
karasikov, line 283: Maybe we should call it `complement` instead of `reverse_complement`, and then you wouldn't have to add `::` in front.
```suggestion
            c = complement(c);
```

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -191,13 +212,75 @@ ::call_outgoing_kmers(node_index node, const OutgoingEdgeCallback &callback) con
...
+    //        lshift    rc
+    // AGCCAT -> *AGCCA -> TGGCT*
+    std::string rev_seq = std::string(1, '\0') + get_node_sequence(node).substr(0, get_k() - 1);
+    ::reverse_complement(rev_seq.begin(), rev_seq.end());
+    assert(rev_seq.back() == '\0');
+
+    // for each n, check for TGGCTn. If found, define and store the index for
+    // rc(n)AGCCA as index(TGGCTn) + offset_
+    const auto *dbg_succ = dynamic_cast<const DBGSuccinct*>(&graph_);
+    if (dbg_succ) {
```
karasikov, line 239:
```suggestion
    if (const auto *dbg_succ = dynamic_cast<const DBGSuccinct*>(&graph_)) {
```

metagraph/src/graph/representation/canonical_dbg.cpp
karasikov: And even better, compute it once and store in bool `k_odd_`. It won't make it any faster, but will make it easier to read.

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -125,13 +110,69 @@ void CanonicalDBG::map_to_nodes(std::string_view sequence,
...
+            [&](node_index next, uint64_t /* match length */) {
+                auto edge = dbg_succ->kmer_to_boss_index(next);
+                char c = boss.decode(boss.get_minus_k_value(edge, get_k() - 2).first);
+                if (c == boss::BOSS::kSentinel)
+                    return;
+
+                size_t i = boss.encode(::reverse_complement(c));
+
+                if (children[i] == DeBruijnGraph::npos)
+                    children[i] = next + offset_;
```
karasikov, line 151: KmerExtractorBOSS has methods `complement` and `reverse_complement`.
```suggestion
                boss::BOSS::TAlphabet c = boss.get_minus_k_value(edge, get_k() - 2).first;
                if (c == boss::BOSS::kSentinelCode)
                    return;

                c = KmerExtractorBOSS::complement(c);

                if (children[c] == DeBruijnGraph::npos)
                    children[c] = next + offset_;
```

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -125,13 +110,69 @@ void CanonicalDBG::map_to_nodes(std::string_view sequence,
...
+            },
+            get_k() - 1
+        );
+
+    } else {
+        for (size_t i = 0; i < alphabet.size(); ++i) {
+            if (children[i] != DeBruijnGraph::npos)
```
karasikov, line 158: use `c` for characters from the alphabet, not `i`.

karasikov (changes requested):

metagraph/src/graph/representation/succinct/boss.cpp
```diff
@@ -2970,6 +2970,13 @@ void BOSS::call_unitigs(Call<std::string&&, std::vector<edge_index>&&> callback,
     }, num_threads, true, kmers_in_single_form, subgraph_mask, true);
 }
 
+void BOSS::call_outgoing(edge_index edge, const Call<edge_index> &callback) const {
+    edge = succ_last(edge);
+    do {
+        callback(edge);
+    } while (--edge && !get_last(edge));
+}
```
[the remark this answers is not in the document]
hmusta: done

metagraph/src/graph/representation/canonical_dbg.hpp
hmusta: How about `find all parent nodes of node in the CanonicalDBG which are represented in the reverse complement orientation in the underlying primary graph`?

metagraph/src/graph/representation/canonical_dbg.hpp
hmusta: Instead of keeping this fixed, I set it up such that `--batch-size` also sets the size of the cache. I still have to test this out on larger graphs, but it seems to work well for the microbe 750 subset.

metagraph/src/cli/align.cpp
```diff
@@ -459,6 +454,7 @@ int align_to_graph(Config *config) {
 
         auto it = fasta_parser.begin();
         auto end = fasta_parser.end();
+
         while (it != end) {
             uint64_t num_bytes_read = 0;
             // A generator that can be called multiple times until all sequences
```
hmusta, line 460: update comment

metagraph/src/common/seq_tools/reverse_complement.hpp
```diff
@@ -28,38 +28,38 @@
 
 #include <htslib/kseq.h>
 
+static const unsigned char kCompTab[] = {
```
karasikov, line 31:
```suggestion
const unsigned char kCompTab[] = {
```

karasikov, line 31: `kCompTab ` looks like it's a scalar constant.

Let's call it COMPL_TAB?

metagraph/src/graph/representation/canonical_dbg.hpp
karasikov: Maybe having the cache size of 1 or 10 MB is enough and increasing it never makes things faster.

In this case, we could hard code it to that value.

karasikov (approved):

metagraph/src/graph/representation/canonical_dbg.hpp
hmusta: I'm playing around with different values now, so far on a 5MB input, it seems like 100KB and 1MB are optimal. I'll see if that's still the case with a larger file.

hmusta: There are diminishing returns past 500KB, but that could just be because of my particular input

metagraph/src/common/seq_tools/reverse_complement.hpp
hmusta: done

metagraph/src/graph/representation/canonical_dbg.hpp
hmusta: I created a batching scheme where if the given batch size is too big (the entire query fits in the batch), then it splits the batch into `get_num_threads()` mini batches

PR #278, opened 2021-01-22
Refactor and simplify aligner method classes

hmusta:


metagraph/src/graph/alignment/aligner_methods.hpp
```diff
@@ -110,62 +105,67 @@ class ExactSeeder : public ExactMapSeeder<NodeType> {
...
 
     ExactSeeder(const DeBruijnGraph &graph, const DBGAlignerConfig &config)
-          : ExactMapSeeder<NodeType>(graph, config) {}
+          : graph_(graph), config_(config) { assert(config_.check_config_scores()); }
 
-    void call_seeds(std::function<void(Seed&&)> callback) const;
+    void call_seeds(std::function<void(Seed&&)> callback) const override;
+
+    const DeBruijnGraph& get_graph() const override { return graph_; }
+    const DBGAlignerConfig& get_config() const override { return config_; }
```
karasikov, line 113: same, you probably don't need these getters...

karasikov (changes requested):

metagraph/tests/graph/test_aligner.cpp
```diff
@@ -1451,6 +1451,13 @@ TEST(DBGAlignerTest, align_suffix_seed_snp_min_seed_length) {
         check_extend(graph, aligner.get_config(), paths, query);
     }
     {
+        size_t k = 15;
+        std::string reference = "TTTCGAGGCCAAAG" "CTTTCGAGGCCAA";
+        std::string query =                 "AC" "CTTTCGAGGCCAA";
```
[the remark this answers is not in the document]
hmusta: The point is to emphasize the position at which the two strings differ. Another way to do it would be to put a comment right above with an arrow indicating that position.

hmusta:
I'm gonna rework this a bit. I want to restructure these methods so the main alignment core takes a query generator rather than a single query as input.

metagraph/src/graph/alignment/dbg_aligner.hpp
```diff
@@ -139,96 +148,94 @@ class DBGAligner : public SeedAndExtendAligner<Seeder, Extender> {
 };
 
 
-template <class Seeder, class Extender>
-inline void SeedAndExtendAligner<Seeder, Extender>
+template <class ThisSeeder, class ThisExtender>
+inline void SeedAndExtendAligner<ThisSeeder, ThisExtender>
```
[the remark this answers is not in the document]
hmusta: I did this to prevent these from shadowing the abstract parent class `Seeder`. I fixed this by renaming the parent classes to `ISeeder` and `IExtender`

metagraph/src/graph/alignment/aligner_seeder_methods.cpp
```diff
@@ -288,7 +177,7 @@ void MEMSeeder<NodeType>::call_seeds(std::function<void(Seed&&)> callback) const
         score_t match_score = partial_sum[end_it - query.data()]
             - partial_sum[begin_it - query.data()];
```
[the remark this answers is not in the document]
hmusta: I cleaned this up in a few more places

metagraph/src/graph/alignment/aligner_seeder_methods.cpp
```diff
@@ -92,151 +80,44 @@ SuffixSeeder<NodeType>::SuffixSeeder(const DeBruijnGraph &graph,
...
+                    DBGAlignerConfig::score_t match_score = config.match_score(seed_seq);
+
+                    if (match_score > config.min_cell_score) {
+                        Seed seed(seed_seq, { alt_node }, match_score, i,
+                                  orientation, k - seed_length);
+
+                        assert(seed.is_valid(graph, &config));
+                        callback(std::move(seed));
+                    }
+                },
```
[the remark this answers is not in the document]
hmusta: done

metagraph/src/graph/alignment/aligner_seeder_methods.cpp
```diff
@@ -92,151 +80,44 @@ SuffixSeeder<NodeType>::SuffixSeeder(const DeBruijnGraph &graph,
...
-    const auto &graph = dynamic_cast<const DBGSuccinct&>(get_graph());
+void SuffixSeeder<NodeType>::call_seeds(std::function<void(Seed&&)> callback) const {
+    base_seeder_->call_seeds(callback);
+
+    const auto query = get_query();
     const auto &config = get_config();
-    auto &query_nodes = get_query_nodes();
-    auto &offsets = get_offsets();
-    alt_query_nodes.clear();
+    const auto &query_nodes = get_query_nodes();
```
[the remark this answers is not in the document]
hmusta: done

metagraph/src/graph/alignment/aligner_seeder_methods.cpp
```diff
@@ -92,151 +80,44 @@ SuffixSeeder<NodeType>::SuffixSeeder(const DeBruijnGraph &graph,
...
+    const auto &graph = dynamic_cast<const DBGSuccinct&>(get_graph());
 
-    base_seeder_->initialize(query, orientation);
     size_t k = graph.get_k();
-
-    // if node suffixes of length < k were matched, then query_nodes.size() may
-    // be greater than query.size() - k + 1
-    if (get_num_matching_kmers()
+    if (this->get_num_matching_kmers()
```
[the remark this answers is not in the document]
hmusta: I changed how this behaves, so now it counts matching nucleotides instead of k-mers. So, the comment isn't relevant anymore.

metagraph/src/graph/alignment/aligner_methods.hpp
```diff
@@ -110,62 +105,67 @@ class ExactSeeder : public ExactMapSeeder<NodeType> {
...
-                                      graph.max_index() + 1)) {}
+    SuffixSeeder(const DeBruijnGraph &graph, const DBGAlignerConfig &config);
+
+    void initialize(std::string_view query, bool orientation) {
+        base_seeder_->initialize(query, orientation);
+    }
+
+    void call_seeds(std::function<void(Seed&&)> callback) const;
+
+    const DeBruijnGraph& get_graph() const { return base_seeder_->get_graph(); }
```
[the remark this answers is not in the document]
hmusta: done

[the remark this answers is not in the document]
hmusta: I made all of these methods protected.

metagraph/src/graph/alignment/aligner_methods.hpp
hmusta: I made these protected instead

metagraph/src/graph/alignment/aligner_methods.hpp
```diff
@@ -184,6 +184,7 @@ class Extender {
 
     virtual void initialize(const DBGAlignment &path) = 0;
     virtual void initialize_query(const std::string_view query) = 0;
+    virtual const std::string_view get_query() const = 0;
 
     virtual const DeBruijnGraph& get_graph() const = 0;
     virtual const DBGAlignerConfig& get_config() const = 0;
```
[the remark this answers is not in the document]
hmusta: done

metagraph/src/graph/alignment/aligner_methods.hpp
```diff
@@ -220,16 +223,17 @@ class DefaultColumnExtender : public Extender<NodeType> {
...
-    virtual const DBGAlignerConfig& get_config() const override { return config_; }
+    const DBGAlignerConfig& get_config() const override { return config_; }
 
-    virtual const DPTable<NodeType>& get_dp_table() const { return dp_table; }
-    virtual const ColumnQueue& get_column_queue() const{ return columns_to_update; }
+    const DPTable<NodeType>& get_dp_table() const { return dp_table; }
+    const ColumnQueue& get_column_queue() const { return columns_to_update; }
 
   protected:
     virtual void set_graph(const DeBruijnGraph &graph) override { graph_ = &graph; }
```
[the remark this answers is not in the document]
hmusta: My plan a while back was to have the graph be swapped out for a masked graph in certain cases, but I think this just complicates everything. I'll shelve this idea for now.

metagraph/tests/graph/test_aligner.cpp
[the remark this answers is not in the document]
hmusta: I've annotated most of the sequences to mark the positions where edits occur.

metagraph/src/graph/alignment/aligner_methods.hpp
```diff
@@ -13,203 +13,187 @@
...
-    const std::vector<uint8_t>& get_offsets() const { return offsets_; }
-    const std::vector<score_t>& get_partial_sums() const { return partial_sum_; }
-    bool get_orientation() const override { return orientation_; }
+    const std::vector<DBGAlignerConfig::score_t>& get_partial_sums() const {
+        return partial_sum_;
+    }
 
-    virtual void initialize(std::string_view query, bool orientation) override;
+    const DeBruijnGraph& get_graph() const { return graph_; }
+    const DBGAlignerConfig& get_config() const { return config_; }
```
karasikov, line 56: Simply move `graph_` and `config_`to protected so you can access them directly and remove these methods.

metagraph/src/graph/alignment/aligner_methods.hpp
```diff
@@ -13,203 +13,187 @@
...
 
-    void initialize(std::string_view query, bool orientation);
+template <typename NodeType = typename DeBruijnGraph::node_index>
+class ExactSeeder : public ExactMapSeeder<NodeType> {
+  public:
+    typedef NodeType node_index;
+    typedef typename ISeeder<NodeType>::Seed Seed;
 
-    void call_seeds(std::function<void(Seed&&)> callback) const;
+    template <typename ... Args>
```
karasikov, line 112:
```suggestion
    template <typename... Args>
```

metagraph/src/graph/alignment/aligner_methods.hpp
```diff
@@ -219,20 +203,12 @@ class DefaultColumnExtender : public Extender<NodeType> {
...
-
-    virtual const DBGAlignerConfig& get_config() const override { return config_; }
-
-    virtual const DPTable<NodeType>& get_dp_table() const { return dp_table; }
-    virtual const ColumnQueue& get_column_queue() const{ return columns_to_update; }
+    const DPTable<NodeType>& get_dp_table() const { return dp_table; }
 
   protected:
-    virtual void set_graph(const DeBruijnGraph &graph) override { graph_ = &graph; }
+    const DeBruijnGraph& get_graph() const { return graph_; }
```
karasikov, line 209: same, remove this.

metagraph/src/graph/alignment/aligner_seeder_methods.cpp
```diff
@@ -36,269 +50,157 @@ void ExactMapSeeder<NodeType>::initialize(std::string_view query, bool orientati
...
-    const auto &query_nodes = this->get_query_nodes();
-
-    query_node_flags_.assign(query_nodes.size(), 0);
+        if (match_score > config.min_cell_score) {
+            Seed seed(seed_seq,
+                      { alt_node },
+                      match_score,
+                      i,
+                      orientation,
+                      k - seed_length);
```
karasikov, line 115: It's not a declaration, so you can have multiple parameters per line. As you prefer.

metagraph/src/graph/alignment/aligner_seeder_methods.cpp
```diff
@@ -36,269 +50,157 @@ void ExactMapSeeder<NodeType>::initialize(std::string_view query, bool orientati
...
+            assert(seed.is_valid(dbg_succ_, &config));
+            callback(std::move(seed));
+        }
+    };
+
+    for (size_t i = 0; i + config.min_seed_length <= query.size(); ++i) {
+        if (i >= query_nodes.size() || !query_nodes[i]) {
+            size_t max_seed_length = std::min(config.max_seed_length,
+                std::min(k, query.size() - i)
+            );
```
karasikov, line 126: if you have many statements like this, it would make sense to write a new min in utils.

`utils::min(first, second, third, ...);`

metagraph/src/graph/alignment/aligner_seeder_methods.cpp
```diff
@@ -36,269 +50,157 @@ void ExactMapSeeder<NodeType>::initialize(std::string_view query, bool orientati
...
+    }
+
     // find start of MEM
-    auto it = query_node_flags_.begin();
-    while ((it = std::find_if(it, query_node_flags_.end(),
+    auto it = query_node_flags.begin();
+    while ((it = std::find_if(it, query_node_flags.end(),
                               [](uint8_t flags) { return flags & 2; }))
-            < query_node_flags_.end()) {
+            < query_node_flags.end()) {
```
karasikov, line 168:
```suggestion
            != query_node_flags.end()) {
```

metagraph/src/graph/alignment/dbg_aligner.cpp
```diff
@@ -0,0 +1,25 @@
+#include "dbg_aligner.hpp"
+
+namespace mtg {
+namespace graph {
+namespace align {
+
+IDBGAligner::DBGQueryAlignment IDBGAligner::align(const std::string_view query,
+                                                  bool orientation) const {
```
karasikov, line 8: what does `orientation` mean? Does it specify that you pass the forward orientation, or it asks to align with reverse complement?

Try to choose a more descriptive name.

metagraph/src/graph/alignment/dbg_aligner.hpp
```diff
@@ -23,17 +23,26 @@ class IDBGAligner {
     typedef Alignment<node_index> DBGAlignment;
     typedef QueryAlignment<node_index> DBGQueryAlignment;
     typedef typename DBGAlignment::score_t score_t;
+    typedef std::function<void(std::string_view /* header */,
+                               std::string_view /* seq */,
+                               bool /* orientation */)> QueryCallback;
```
karasikov, line 28: ha, perfect, you already have this guy and don't have to redefine (see my comment above).

karasikov (changes requested):

metagraph/src/graph/alignment/dbg_aligner.hpp
```diff
@@ -59,46 +59,40 @@ class SeedAndExtendAligner : public IDBGAligner {
...
-    typedef const std::function<void(const std::function<void(ISeeder<node_index>&&,
-                                                              IExtender<node_index>&&)>&)> AlignmentCoreGenerator;
     typedef const std::function<void(const std::function<void(DBGAlignment&&)>&,
                                      const std::function<score_t(const DBGAlignment&)>&)> AlignmentGenerator;
 
     // Generate seeds, then extend them
     void align_core(const std::string_view query,
-                    const AlignmentCoreGenerator &seed_generator,
+                    ISeeder<node_index>&& seeder,
+                    IExtender<node_index>&& extend,
```
karasikov, line 68: extend -> extender ?

metagraph/src/graph/alignment/dbg_aligner.cpp
hmusta: I've renamed this to `query_orientation` do indicate that it's stating the orientation of the query (relative to the input)

metagraph/src/graph/alignment/aligner_seeder_methods.cpp
hmusta: done

metagraph/src/graph/alignment/aligner_seeder_methods.cpp
hmusta: ok

metagraph/src/graph/alignment/aligner_methods.hpp
hmusta: done

metagraph/src/graph/alignment/aligner_methods.hpp
hmusta: done

metagraph/src/cli/align.cpp
```diff
@@ -397,68 +405,92 @@ int align_to_graph(Config *config) {
...
-                                                      query,
-                                                      *graph,
-                                                      secondary,
-                                                      header)
-                                                  ) << "\n";
+            thread_pool.enqueue([&](auto batch) {
+                aligner->align_batch(
+                    [&](const auto &query_callback) {
+                        for (const auto &[seq, header] : batch) {
+                            query_callback(header, seq, false /* orientation */);
```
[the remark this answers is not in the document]
hmusta: done

metagraph/src/cli/align.cpp
```diff
@@ -397,68 +405,92 @@ int align_to_graph(Config *config) {
...
+                        }
+                    },
+                    [&](std::string_view header, auto&& paths) {
+                        const std::string_view query = paths.get_query();
+                        std::lock_guard<std::mutex> lock(print_mutex);
+                        if (!config->output_json) {
+                            *out << header << "\t" << query;
+                            if (paths.empty()) {
+                                *out << "\t*\t*\t" << config->alignment_min_path_score
+                                     << "\t*\t*\t*";
```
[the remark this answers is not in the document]
hmusta: done

metagraph/src/cli/align.cpp
```diff
@@ -397,68 +405,92 @@ int align_to_graph(Config *config) {
...
+                            }
+
+                            if (paths.empty()) {
+                                *out << Json::writeString(builder,
+                                                          DBGAligner<>::DBGAlignment().to_json(
+                                                              query,
+                                                              *graph,
+                                                              secondary,
+                                                              header)
+                                                          ) << "\n";
```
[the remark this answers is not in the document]
hmusta: done

metagraph/src/cli/align.cpp
```diff
@@ -397,68 +405,92 @@ int align_to_graph(Config *config) {
...
+            for ( ; it != end && num_bytes_read <= batch_size; ++it) {
+                seq_batch.emplace_back(
+                    it->seq.s,
+                    config->fasta_anno_comment_delim != Config::UNINITIALIZED_STR
+                        && it->comment.l
+                       ? utils::join_strings({ it->name.s, it->comment.s },
+                                             config->fasta_anno_comment_delim,
+                                             true)
+                       : std::string(it->name.s)
+                );
```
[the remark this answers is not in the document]
hmusta: done

metagraph/src/graph/alignment/aligner_extender_methods.cpp
```diff
@@ -654,14 +652,16 @@ ::fork_extension(NodeType node,
     double num_bytes = static_cast<double>(dp_table.num_bytes()) / 1024 / 1024;
     bool added = false;
 
-    if (dynamic_cast<const DBGSuccinct*>(graph_)) {
-        const auto &dbg_succ = *dynamic_cast<const DBGSuccinct*>(graph_);
-        const auto &boss = dbg_succ.get_boss();
-        graph_->adjacent_outgoing_nodes(node, [&](auto next_node) {
+    const auto *dbg_succ = dynamic_cast<const DBGSuccinct*>(&graph_);
+    if (dbg_succ) {
```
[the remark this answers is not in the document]
hmusta: done

metagraph/src/cli/align.cpp
```diff
@@ -39,7 +39,7 @@ DBGAlignerConfig initialize_aligner_config(size_t k, const Config &config) {
     aligner_config.min_cell_score = config.alignment_min_cell_score;
     aligner_config.min_path_score = config.alignment_min_path_score;
     aligner_config.xdrop = config.alignment_xdrop;
-    aligner_config.exact_kmer_match_fraction = config.discovery_fraction;
+    aligner_config.exact_match_fraction = config.discovery_fraction;
```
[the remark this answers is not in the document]
hmusta: Threshold isn't that much more informative either. Maybe `exact_match_lower_bound`?

metagraph/src/graph/alignment/aligner_helper.cpp
```diff
@@ -703,33 +703,33 @@ Json::Value Alignment<NodeType>::path_json(size_t node_size,
...
+                                         const std::string_view read_name,
+                                         const std::string_view label) const {
     assert(is_valid(graph));
 
     // encode alignment
     Json::Value alignment;
 
-    alignment["name"] = read_name;
-    alignment["sequence"] = query;
+    alignment["name"] = read_name.data() ? std::string(read_name) : "";
```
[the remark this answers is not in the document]
hmusta: no, `std::string(NULL)` throws an exception
```
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_M_construct null not valid
Aborted (core dumped)
```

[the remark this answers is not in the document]
karasikov: yeah, but you wouldn't pass std::string(NULL), you would pass std::string(std::string_view{}), right? Or that throws as well?

metagraph/src/graph/alignment/aligner_helper.hpp
```diff
@@ -418,9 +423,11 @@ class QueryAlignment {
   public:
     typedef Alignment<NodeType> value_type;
 
-    QueryAlignment(const std::string_view query);
-    QueryAlignment(const QueryAlignment &other);
-    QueryAlignment(QueryAlignment&& other) noexcept;
+    QueryAlignment(const std::string_view query, bool orientation = false);
```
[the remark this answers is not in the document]
hmusta: Renamed to `is_reverse_complement`

metagraph/src/graph/alignment/aligner_seeder_methods.cpp
hmusta: Turns out `std::min` works with initializer lists...

metagraph/src/graph/alignment/dbg_aligner.cpp
hmusta: I've again renamed it to `is_reverse_complement`. Hopefully that's clear enough

metagraph/src/graph/alignment/aligner_helper.hpp
```diff
@@ -473,8 +480,8 @@ class QueryAlignment {
     // vector may be incorrect, so this corrects them
     void fix_pointers(const std::string &query, const std::string &query_rc);
 
-    const std::string query_;
-    const std::string query_rc_;
+    std::string query_;
+    std::string query_rc_;
     std::vector<value_type> alignments_;
```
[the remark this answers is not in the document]
hmusta: The `Alignment` objects stored in `QueryAlignment` need to store some kind of reference to the query strings to be able to access the string. If indices are stored, then one would have to also store either a pointer to a `QueryAlignment` object (making it less versatile), or a pointer to the string (which is essentially the same as what's already done).

In both cases, when a `QueryAlignment` is moved, some kind of pointer has to be updated in each of the `Alignment`s

metagraph/src/cli/config/config.cpp
```diff
@@ -861,8 +861,9 @@ void Config::print_usage(const std::string &prog_name, IdentityType identity) {
             fprintf(stderr, "\t   --count-kmers \t\tfor each sequence, report the number of k-mers discovered in graph [off]\n");
             fprintf(stderr, "\n");
             fprintf(stderr, "\t   --query-presence \t\ttest sequences for presence, report as 0 or 1 [off]\n");
-            fprintf(stderr, "\t   --discovery-fraction [FLOAT] fraction of k-mers required to count sequence [0.7]\n");
+            fprintf(stderr, "\t   --discovery-fraction [FLOAT] fraction of matching nucleotides required to align sequence [0.7]\n");
```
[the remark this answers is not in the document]
hmusta: I added `--align-min-exact-match-threshold`

metagraph/src/cli/align.cpp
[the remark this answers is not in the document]
hmusta: I added `min_exact_match_threshold`

metagraph/src/graph/alignment/aligner_helper.cpp
```diff
@@ -703,33 +703,33 @@ Json::Value Alignment<NodeType>::path_json(size_t node_size,
...
-        path["name"] = label;
+    if (label.data())
+        path["name"] = std::string(label);
 
     return path;
 }
 
 template <typename NodeType>
-Json::Value Alignment<NodeType>::to_json(const std::string &query,
+Json::Value Alignment<NodeType>::to_json(const std::string_view query,
```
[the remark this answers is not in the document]
hmusta: It's a hint that `query` will never be changed to point somewhere else. However, we're not very religious about doing this hint everywhere in the code, so I just changed them to non-const

metagraph/integration_tests/test_align.py
```diff
@@ -52,7 +52,7 @@ def test_simple_align_all_graphs(self, representation):
         self.assertEqual('nodes (k): 16438', params_str[1])
         self.assertEqual('canonical mode: no', params_str[2])
 
-        stats_command = '{exe} align -i {graph} --align-vertical-bandwidth 1000000 --discovery-fraction 0.0 {reads}'.format(
+        stats_command = '{exe} align -i {graph} --align-vertical-bandwidth 1000000 --align-min-exact-match-threshold 0.0 {reads}'.format(
```
karasikov, line 55: "min" and "threshold" are synonyms here. Use just one of them.

--align-min-exact-match
or
--align-exact-match-threshold

karasikov (changes requested):

metagraph/integration_tests/test_align.py
hmusta: I used `align-min-exact-match`

metagraph/src/cli/align.cpp
```diff
@@ -397,68 +407,95 @@ int align_to_graph(Config *config) {
...
+            std::vector<std::pair<std::string, std::string>> seq_batch;
+            num_bytes_read = 0;
+            for ( ; it != end && num_bytes_read <= batch_size; ++it) {
+                std::string header
+                    = config->fasta_anno_comment_delim != Config::UNINITIALIZED_STR
+                        && it->comment.l
+                            ? utils::join_strings({ it->name.s, it->comment.s },
+                                                  config->fasta_anno_comment_delim,
+                                                  true)
+                            : std::string(it->name.s);
```
karasikov, line 435:
```suggestion
                std::string header
                    = config->fasta_anno_comment_delim != Config::UNINITIALIZED_STR
                            ? utils::join_strings({ it->name.s, it->comment.s },
                                                  config->fasta_anno_comment_delim,
                                                  true)
                            : std::string(it->name.s);
```

karasikov: this should be equivalent

metagraph/src/cli/align.cpp
```diff
@@ -397,68 +407,95 @@ int align_to_graph(Config *config) {
...
+
+                            if (paths.empty()) {
+                                Json::Value json_line = DBGAligner<>::DBGAlignment().to_json(
+                                    query, *graph, secondary, header
+                                );
+                                sout += fmt::format(
+                                    "{}\n", Json::writeString(builder, json_line)
+                                );
+                            }
+                        }
```
karasikov, line 491: This could be a separate function.

metagraph/src/graph/alignment/aligner_aggregator.hpp
```diff
@@ -16,7 +16,7 @@ class AlignmentAggregator {
     typedef Alignment<NodeType> DBGAlignment;
     typedef typename DBGAlignment::score_t score_t;
 
-    AlignmentAggregator(const std::string_view query, const std::string_view rc_query,
+    AlignmentAggregator(std::string_view query, std::string_view rc_query,
```
karasikov, line 19:
```suggestion
    AlignmentAggregator(std::string_view query,
                        std::string_view rc_query,
```

metagraph/src/graph/alignment/aligner_extender_methods.cpp
```diff
@@ -32,9 +33,7 @@ ::initialize_query(const std::string_view query) {
     assert(config_.match_score(query) == partial_sums_.front());
     assert(config_.get_row(query.back())[query.back()] == partial_sums_.back());
 
-    profile_score.clear();
-    profile_op.clear();
```
karasikov, line 36: hm, was this not used? Or now you assume you create a new DefaultColumnExtender for each alignment?

Maybe add `assert(profile_score.size() == 0);`?

metagraph/src/graph/alignment/aligner_helper.cpp
```diff
@@ -44,7 +44,8 @@ bool DPTable<NodeType>::add_seed(const Alignment<NodeType> &seed,
         table_init.gap_scores[start_pos] = std::max(
             last_op == Cigar::DELETION
                 ? table_init.scores[start_pos]
-                : table_init.scores[start_pos] - last_char_score + config.gap_opening_penalty + config.gap_opening_penalty,
+                : table_init.scores[start_pos] - last_char_score
+                    + config.gap_opening_penalty + config.gap_opening_penalty,
```
karasikov, line 48:
```suggestion
                : table_init.scores[start_pos] - last_char_score
                                               + config.gap_opening_penalty
                                               + config.gap_opening_penalty,
```
way easier to read

karasikov: is this on purpose you add `config.gap_opening_penalty` twice?

metagraph/src/graph/alignment/aligner_helper.cpp
```diff
@@ -410,8 +411,8 @@ void Alignment<NodeType>::trim_offset() {
             continue;
         }
 
-        size_t jump = std::min(std::min(offset_, size_t(it->second)),
-                               static_cast<size_t>(nodes_.end() - jt));
+        size_t jump = std::min({ offset_, size_t(it->second),
+                               static_cast<size_t>(nodes_.end() - jt) });
```
karasikov, line 415:
```suggestion
        size_t jump = std::min({ offset_, static_cast<size_t>(it->second),
                                          static_cast<size_t>(nodes_.end() - jt) });
```

metagraph/src/graph/alignment/dbg_aligner.hpp
```diff
@@ -393,7 +401,7 @@ ::align_aggregate(DBGQueryAlignment &paths,
     );
 
     path_queue.call_alignments([&](auto&& alignment) {
-        assert(alignment.is_valid(graph_, &config_));
+        assert(alignment.is_valid(get_graph(), &get_config()));
```
karasikov, line 404: why get_graph()? just call graph_ directly. Same for config_.

metagraph/src/graph/alignment/aligner_helper.hpp
karasikov, line 485: Introducing this `value_type` makes it only harder to track what is stored. Remove it.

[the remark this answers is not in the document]
karasikov: Can't you do that in copy/move constructors in `Alignment<NodeType>`? Then you wouldn't need to define copy/move constructors here and this `fix_pointers` method.

karasikov (changes requested):

metagraph/src/graph/alignment/aligner_extender_methods.cpp
hmusta: Before, I used to create an `Extender` first, then `initialize` it with a query. I've now combined those two so the query is passed to the constructor. Since it's in the constructor, we know that they're empty already.

metagraph/src/graph/alignment/aligner_helper.cpp
hmusta: I'll add a comment to describe this. This array is used to store the best score achieved if the last operation was a deletion. So, I'm initializing this to the value that you would get if you performed an insertion and a deletion instead of a character match.

metagraph/src/graph/alignment/dbg_aligner.hpp
hmusta: This method was moved to the abstract `SeedAndExtendAligner` class, which doesn't store a graph or a config.

metagraph/src/graph/alignment/aligner_helper.hpp
hmusta: I don't remember for sure, but I think `value_type` has to be defined for structured binding to work.

metagraph/src/cli/align.cpp
hmusta: done

metagraph/src/graph/alignment/aligner_helper.hpp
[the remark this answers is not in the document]
hmusta: I solved this by instead replacing `std::string` with `std::shared_ptr<std::string>`. This way, as long as at least one `QueryAlignment` is alive, the `Alignment` objects won't be invalid.

metagraph/src/cli/align.cpp
hmusta: I tried this and it throws an exception because `it->comment.s == NULL`

metagraph/src/graph/alignment/aligner_helper.hpp
karasikov: If you added it to make some other feature like structured binding, you don't have to use it. It just makes more tangled. What is the point?

karasikov: I mean, you can introduce it, if it's necessary, but don't use it. Write the original type in this line.

metagraph/src/graph/alignment/dbg_aligner.hpp
karasikov: but it does store a reference to it, right?
Or you can always make it so if you can implement these getters.

hmusta: when I said "doesn't store a graph or a config" I meant that it doesn't store pointers, it just as `get_graph` and `get_config` as pure virtual methods

metagraph/src/graph/alignment/aligner_helper.hpp
hmusta: I tried it out now and it looks like it's not needed, having a `const_iterator` is good enough

metagraph/src/graph/alignment/dbg_aligner.hpp
karasikov: Then its children must store them, so you can move those pointers/references to this class.
What is the point to declare these virtual getters?

metagraph/src/cli/align.cpp
karasikov: sounds like a bug that needs to be fixed

karasikov: ah, no, no bug

metagraph/src/graph/alignment/dbg_aligner.hpp
hmusta: I don't know why it's not showing up for me in "Files changed", but the point is that with this structure, I can now move most of the alignment code out of the header, since most of the alignment functions don't depend `DBGAligner`'s three template classes.

hmusta: Though I could move the references to `ISeedAndExtendAligner` and make them protected, then anything that inherits from it can use them.

hmusta:
> are you sure you need these?

they're not absolutely required, no

metagraph/src/graph/alignment/dbg_aligner.hpp
```diff
@@ -24,377 +20,153 @@ class IDBGAligner {
...
+    virtual void align_batch(const QueryGenerator &generate_query,
+                             const AlignmentCallback &callback) const = 0;
+    DBGQueryAlignment align(std::string_view query,
+                            bool is_reverse_complement = false) const;
 };
 
-template <class Seeder = ExactSeeder<>,
-          class Extender = DefaultColumnExtender<>>
-class SeedAndExtendAligner : public IDBGAligner {
+template <class AlignmentCompare = std::less<IDBGAligner::DBGAlignment>>
```
karasikov, line 38: this template is not used anywhere

metagraph/src/graph/alignment/dbg_aligner.hpp
```diff
@@ -24,377 +20,153 @@ class IDBGAligner {
...
+                    const AlignmentGenerator &alignment_generator) const;
 
+  protected:
+    const DeBruijnGraph &graph_;
+    DBGAlignerConfig config_;
+};
 
 template <class Seeder = ExactSeeder<>,
           class Extender = DefaultColumnExtender<>,
           class AlignmentCompare = std::less<Alignment<>>>
```
karasikov, line 108: I would suggest removing this template parameter `AlignmentCompare`.
It's only passed to the priority queue in AlignmentAggregator, and why would you be interested in getting something other than the best alignments?

karasikov (changes requested):

metagraph/src/graph/alignment/dbg_aligner.hpp
```diff
@@ -24,377 +20,153 @@ class IDBGAligner {
...
-    typedef Alignment<node_index> DBGAlignment;
-    typedef QueryAlignment<node_index> DBGQueryAlignment;
-    typedef typename DBGAlignment::score_t score_t;
+    typedef ISeedAndExtendAligner<AlignmentCompare> BaseAligner;
+    typedef IDBGAligner::node_index node_index;
+    typedef IDBGAligner::DBGAlignment DBGAlignment;
+    typedef IDBGAligner::DBGQueryAlignment DBGQueryAlignment;
+    typedef IDBGAligner::score_t score_t;
+    typedef IDBGAligner::QueryGenerator QueryGenerator;
+    typedef IDBGAligner::AlignmentCallback AlignmentCallback;
```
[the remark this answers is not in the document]
hmusta: For now it's like this, but #279 introduces another derived class called `LabeledDBGAligner` which inherits from `ISeedAndExtendAligner`, so that's why I structured it this way.

metagraph/src/graph/alignment/dbg_aligner.hpp
```diff
@@ -24,377 +20,153 @@ class IDBGAligner {
...
           class Extender = DefaultColumnExtender<>,
           class AlignmentCompare = std::less<Alignment<>>>
-class DBGAligner : public SeedAndExtendAligner<Seeder, Extender> {
+class DBGAligner : public ISeedAndExtendAligner<AlignmentCompare> {
   public:
-    typedef DeBruijnGraph::node_index node_index;
-    typedef Alignment<node_index> DBGAlignment;
-    typedef QueryAlignment<node_index> DBGQueryAlignment;
-    typedef typename DBGAlignment::score_t score_t;
+    typedef ISeedAndExtendAligner<AlignmentCompare> BaseAligner;
```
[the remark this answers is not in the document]
hmusta: This was left over from a previous refactor.

metagraph/src/graph/alignment/dbg_aligner.hpp
hmusta: it's used in the implementation of AlignAggregate

metagraph/src/graph/alignment/dbg_aligner.hpp
hmusta: There are multiple parameters which may be used to rank alignments. Score is one of them, but whether you'd prefer long or short alignments is a design choice of the aligner, so I wanted to keep this here as a template parameter.

hmusta: For example, local alignment prefers having shorter, high-scoring alignments, whereas semi-global alignment prefers having the largest part of the query covered.

metagraph/src/graph/alignment/dbg_aligner.hpp
```diff
@@ -33,6 +33,8 @@ class IDBGAligner {
                              const AlignmentCallback &callback) const = 0;
     DBGQueryAlignment align(std::string_view query,
                             bool is_reverse_complement = false) const;
+
+    virtual const DBGAlignerConfig& get_config() const = 0;
```
karasikov, line 37: DBGAlignerConfig seems to be specific to the Seed-and-Extend approach. But this seems to be a more abstract class.
`ISeedAndExtendAligner` is only a child of this. So it's an inconsistency...

hmusta: The config has parameters which would be used in almost all aligners (such as match and mismatch scores), along with those used in seed-and-extend approaches.

One way around this would be to have `build_aligner` return `ISeedAndExtendAligner`

metagraph/src/graph/alignment/dbg_aligner.hpp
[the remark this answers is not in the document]
karasikov: Is it possible to inherit those new classes from this `SeedAndExtendAligner` (you would need to inherit `LabeledColumnExtender` from `IExtender`, which I think makes sense).

metagraph/src/graph/alignment/dbg_aligner.hpp
karasikov: Ok, I was confused by the existence of `std::less<Alignment<>>`, which assumes that it's the alignment you're looking for.

If there are these cases where different rankings may be preferred, it doesn't make sense to implement that operator< in Alignment<>.

Better implement the comparator outside Alignment<> and pass it here `template <..., class AlignmentCompare = BestAlignment>`.
Another way is to pass that comparator directly to Alignment<>.

Do you see what I mean?

metagraph/src/graph/alignment/dbg_aligner.hpp
karasikov: Yes, or better cast the thing it returns to `ISeedAndExtendAligner`, before checking `min_seed_length` in it's config.
Then you'll keep the generality and will remove the inconsistency.

hmusta: ok, I did it this way

metagraph/src/graph/alignment/dbg_aligner.hpp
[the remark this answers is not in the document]
hmusta: Essentially, I want to have two different structures, which are currently 1) `ISeedAndExtendAligner` and 2) `DBGAligner`
1) Given a query, seeder, and extender, run the alignment engine and aggregate the results
2) Given a batch of queries, generate (seeder,extender) pairs and pass them to the alignment engine

metagraph/src/graph/alignment/dbg_aligner.hpp
hmusta: I added a new struct called `LocalAlignmentLess` to handle this.

karasikov (approved):
Let's merge, it's getting big.

PR #280, opened 2021-01-28
workaround for filesystem::create_directories bug 87846 in gcc-8.2

hmusta:
See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=87846


hmusta:
If there's anywhere else in the code that you think may be affected that I missed, please fix it.

karasikov (approved):

PR #281, opened 2021-01-29
Test g++-8.0 in CI

hmusta:
This downgrades the version of gcc-8 in the CI to 8.0.1 to better check compatibility with older gcc versions.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -44,7 +44,7 @@ jobs:
         # If clang, use the default version. Otherwise the compiler install with apt-get.
         [ "${{ matrix.compiler }}" != "clang" ] && sudo apt-get install ${{ matrix.compiler }}
 
-        sudo apt-get install g++-8=8-20180414-1ubuntu2 libbz2-dev libhts-dev libjemalloc-dev libboost-all-dev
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -43,7 +43,18 @@ jobs:
...
+            gcc-8-base=8-20180414-1ubuntu2 \
+            gcc-8=8-20180414-1ubuntu2 \
+            libstdc++-8-dev=8-20180414-1ubuntu2 \
+            cpp-8=8-20180414-1ubuntu2 \
+            libgcc-8-dev=8-20180414-1ubuntu2 \
+            libmpx2=8-20180414-1ubuntu2 \
+            libbz2-dev \
+            libhts-dev \
+            libjemalloc-dev \
+            libboost-all-dev
```
karasikov, line 57: Can you add a special build with this version of gcc?
The existing gcc-8 will still use 8.4.0.

hmusta: so have both gcc-8.4 and 8.0? because currently it's using 8.0
```
-- The C compiler identification is GNU 8.0.1
-- The CXX compiler identification is GNU 8.0.1
```

karasikov: True. As you wish then, but maybe add a check and install all these only if it's the gcc-8 build.

hmusta: ok. The gcc-9 and gcc-10 builds still use the correct version, but it's just a waste of time downgrading.

hmusta: done

PR #214, opened 2020-10-18
Support sub-k seeding for DBGSuccinct wrapped in a CanonicalDBG

hmusta:

PR #282, opened 2021-01-29
Support seeds of length < k for primary DBGSuccinct

hmusta:
This is a simpler implementation which replaces #214


metagraph/src/graph/representation/succinct/dbg_succinct.cpp
```diff
@@ -270,27 +270,17 @@ void DBGSuccinct::map_to_nodes_sequentially(std::string_view sequence,
...
-void DBGSuccinct
-::call_nodes_with_suffix_matching_longest_prefix(
-            std::string_view str,
-            std::function<void(node_index, uint64_t /* match length */)> callback,
-            size_t min_match_length,
-            size_t max_num_allowed_matches) const {
-    if (!max_num_allowed_matches)
-        return;
-
+std::tuple<boss::BOSS::edge_index, boss::BOSS::edge_index, size_t>
```
karasikov, line 273: size + range (begin, end)?

metagraph/src/graph/alignment/aligner_helper.cpp
```diff
@@ -462,29 +462,50 @@ void Alignment<NodeType>::reverse_complement(const DeBruijnGraph &graph,
...
+                    if (edge_label == boss::BOSS::kSentinelCode) {
+                        // reverse complement not found
+                        *this = Alignment();
+                        return;
+                    }
+
+                    nodes_.push_back(dbg_succ->boss_to_kmer_index(edge));
+                    rev_seq.push_back(boss.decode(edge_label));
+                }
+            } else {
```
karasikov, line 487: Is `else` written for a wrapper of DBGSuccinct?

If so, add an assertion statement or throw a runtime error, or something.

metagraph/src/graph/alignment/aligner_helper.cpp
```diff
@@ -462,29 +462,50 @@ void Alignment<NodeType>::reverse_complement(const DeBruijnGraph &graph,
...
+                        last_char = c;
+                    });
+
+                    if (!next_node) {
+                        *this = Alignment();
+                        return;
+                    } else {
+                        nodes_.push_back(next_node);
+                        rev_seq.push_back(last_char);
+                    }
```
karasikov, line 505: I think you don't need this block. If you have an edge $****, there must be a path from it to a non-dummy edge, so the block above should always work. Just change the cast to `get_base_dbg_succ`.

karasikov (changes requested):

metagraph/src/graph/representation/succinct/dbg_succinct.cpp
hmusta: I added a comment describing its output. It's more convenient to return the indices of the last outgoing edge from the first and last BOSS node in the range (since `BOSS::tighten_range` can be called on it directly).

metagraph/src/graph/alignment/aligner_helper.cpp
hmusta: done

metagraph/src/graph/alignment/aligner_helper.cpp
hmusta: If `offset_` (the number of characters excluded from the prefix of the node sequence) is greater than the number of sentinel characters, it could be the case that a reverse complement can't be found (e.g., an isolated k-mer, a k-mer with outdegree 0 and a dummy k-mer parent)

karasikov: It's a succinct graph, and you're at a node which ends or starts with $. So there must be a path to a non-dummy node.
SO you can get there without involving the wrapper thing and doing these complex operations.
Do you see what I mean?

hmusta: I understand that, but once you get to a non-dummy k-mer, you may need to continue traversing forward to find the node you're looking for. In that case, you need the canonical wrapper.

hmusta: The idea behind this code block is that given a node with a suffix match of length L, you want to traverse forwards until you hit a node whose L-length prefix matches the starting node's suffix. Then, you reverse complement it to get a new node whose L-length suffix is the reverse complement of the original match.

For example, suppose your node is `$$$ATGTGCGAT`, but your seed is `GTGCGAT`. Then, you need to traverse forwards five steps to get to a node which starts with `GTGCGAT` so you can reverse complement it.

The problem is that if the number of sentinel characters is less than k - L, then you're not guaranteed to find the desired prefix-matching target node.

hmusta: but actually, I can speed this up a bit by using the BOSS graph to do the traversal until the first non-dummy node is reached, then use the canonical wrapper afterwards

metagraph/src/graph/representation/succinct/dbg_succinct.cpp
hmusta: Though I'm thinking that this would probably be better suited as a BOSS function since it doesn't really use anything specific to DBGSuccinct.

karasikov: Sounds good, seems like you operate on the level of the boss table here.

metagraph/src/graph/representation/succinct/dbg_succinct.cpp
```diff
@@ -299,18 +289,34 @@ ::call_nodes_with_suffix_matching_longest_prefix(
...
+}
+
+void DBGSuccinct
+::call_nodes_with_suffix_matching_longest_prefix(
+            std::string_view str,
+            std::function<void(node_index, uint64_t /* match length */)> callback,
+            size_t min_match_length,
+            size_t max_num_allowed_matches) const {
+    if (!max_num_allowed_matches || str.size() < min_match_length)
+        return;
```
karasikov, line 310: If you call this function just in one place, remove it.

karasikov (changes requested):

metagraph/src/graph/representation/succinct/dbg_succinct.hpp
```diff
@@ -72,16 +72,23 @@ class DBGSuccinct : public DeBruijnGraph {
...
-    // are found, or if the maximal prefix is shorter than |min_match_length|, return
+    // Find nodes with a common suffix matching the maximal prefix of the string |str|,
+    // and call these nodes. If more than |max_num_allowed_matches| are found,
+    // or if the maximal prefix is shorter than |min_match_length|, return
     // without calling.
     void call_nodes_with_suffix_matching_longest_prefix(
             std::string_view str,
             std::function<void(node_index, uint64_t /* match length */)> callback,
             size_t min_match_length = 1,
             size_t max_num_allowed_matches = std::numeric_limits<size_t>::max()) const;
```
karasikov, line 83: remove

metagraph/src/graph/representation/succinct/dbg_succinct.cpp
hmusta: it's used in a few places (seeding, CanonicalDBG, align.cpp)

metagraph/src/graph/representation/succinct/boss.hpp
```diff
@@ -721,6 +727,29 @@ BOSS::index_range(RandomAccessIt begin, RandomAccessIt end) const {
...
+
+    if (std::get<0>(range) == 0 || std::get<1>(range) == 0)
+        return {};
+
+    if (std::get<2>(range) == encoded.begin()) {
+        return {};
+    }
+
+    return { std::get<0>(range), std::get<1>(range), std::get<2>(range) - encoded.begin() };
+}
```
karasikov, line 751: better do this in DBGSuccinct. For example, you decode the sequence here, which is useful in line 289 in dbg_succinct.cpp

karasikov: I checked again. It's used in two places. But still, let's not add this.
All these checks can be compactly rewritten there, and, in fact, you don't need all of them.
For instance, instead of checking (std::get<2>(range) != encoded.begin()), you can just check if the range is non-zero.

metagraph/src/graph/alignment/aligner_methods.hpp
```diff
@@ -124,7 +124,9 @@ class SuffixSeeder : public BaseSeeder {
     template <typename... Args>
     SuffixSeeder(Args&&... args)
```
karasikov, line 125: The implementation is in aligner_seeder_methods.cpp.
So, move these declarations to aligner_seeder_methods.hpp?

metagraph/src/graph/alignment/aligner_seeder_methods.cpp
```diff
@@ -81,57 +84,234 @@ void ExactSeeder<NodeType>::call_seeds(std::function<void(Seed&&)> callback) con
...
+                    m < min_seed_length.size() && s > min_seed_length[m];
+                    ++m) {
+                assert(suffix_seeds[m].empty()
+                    || min_seed_length[m] == this->graph_.get_k()
+                                                - suffix_seeds[m][0].get_offset());
+                min_seed_length[m] = s--;
+                suffix_seeds[m].clear();
+            }
+
+            // e.g., match: $$$ATG, want ATG$$$
```
karasikov, line 275: `ATG$$$` doesn't exist (there may be only a single sentinel in the end).
Or you mean a wildcard symbol, not the sentinel? Then write `ATG***`.

metagraph/src/graph/alignment/aligner_seeder_methods.cpp
```diff
@@ -81,57 +84,234 @@ void ExactSeeder<NodeType>::call_seeds(std::function<void(Seed&&)> callback) con
...
-            last_seed_size = std::max(this->graph_.get_k() - 1,
-                                      this->config_.min_seed_length);
+            assert(min_seed_length[i] == this->graph_.get_k() - pos_seeds[0].get_offset());
+            if (pos_seeds.size() <= this->config_.max_num_seeds_per_locus) {
+                for (auto&& seed : pos_seeds) {
+                    callback(std::move(seed));
+                }
+            }
         }
     }
```
karasikov, line 303: if you anyway collect them into a vector, just return the vector.
There is no point in using a callback.

karasikov (approved):

PR #286, opened 2021-02-03
explicitly mark primary graphs

karasikov:


metagraph/src/cli/config/config.cpp
```diff
@@ -726,6 +726,8 @@ std::string Config::graphmode_to_string(DeBruijnGraph::Mode mode) {
             return "canonical";
         case DeBruijnGraph::PRIMARY:
             return "primary";
+        default:
+            throw std::runtime_error("Never happens");
```
hmusta, line 730: This isn't needed. Since it's an `enum`, adding or removing a value will lead to a compilation error anyway.

karasikov: gcc wasn't compiling it for some reason
https://github.com/ratschlab/metagraph/runs/1822432077

hmusta: It's because the function either needs to return a `DeBruijnGraph::Mode` type, or throw an exception. I'd move the `throw` to be outside the `switch` statement so we still get the desired behaviour. This way if we add a new mode, it'll fail to compile unless we add that mode here.

karasikov: I see what you mean

metagraph/src/cli/build.cpp
```diff
@@ -107,7 +107,7 @@ int build_graph(Config *config) {
 
             auto constructor = boss::IBOSSChunkConstructor::initialize(
                 boss_graph->get_k(),
-                config->canonical,
+                config->graph_mode == DeBruijnGraph::CANONICAL,
```
hmusta, line 110: why not pass `graph_mode` directly to `IBOSSChunkConstructor` and let it handle what to do with the mode (rather than losing precision here)

metagraph/src/cli/clean.cpp
```diff
@@ -135,16 +135,17 @@ int clean_graph(Config *config) {
                                           *node_weights,
                                           config->min_unitig_median_kmer_abundance))
                     callback(unitig, path);
-            }, num_threads, config->min_tip_size, graph->is_canonical_mode());
+            }, num_threads, config->min_tip_size, graph->get_mode() == graph::DeBruijnGraph::CANONICAL);
```
hmusta, line 138: similar idea here, why not just pass the mode and let `call_unitigs` figure out how to interpret it?

metagraph/src/cli/clean.cpp
```diff
@@ -135,16 +135,17 @@ int clean_graph(Config *config) {
...
 
         } else if (config->unitigs || config->min_tip_size > 1) {
             graph->call_unitigs(callback,
                                 num_threads,
                                 config->min_tip_size,
-                                graph->is_canonical_mode());
+                                graph->get_mode() == graph::DeBruijnGraph::CANONICAL);
```
hmusta, line 144: same

metagraph/src/cli/clean.cpp
```diff
@@ -135,16 +135,17 @@ int clean_graph(Config *config) {
...
 
         } else {
-            graph->call_sequences(callback, num_threads, graph->is_canonical_mode());
+            graph->call_sequences(callback, num_threads,
+                                  graph->get_mode() == graph::DeBruijnGraph::CANONICAL);
```
hmusta, line 148: same

metagraph/src/cli/clean.cpp
```diff
@@ -271,14 +272,14 @@ int clean_graph(Config *config) {
...
-                graph->is_canonical_mode()
+                true,
+                graph->get_mode()
             );
 
             // the outer for loop is parallelized, so don't start another thread
             // pool here
             dump_contigs_to_fasta(filebase, [&](auto dump_sequence, auto /* num_threads */) {
-                graph_slice.call_sequences(dump_sequence, 1, graph_slice.is_canonical_mode());
+                graph_slice.call_sequences(dump_sequence, 1, graph_slice.get_mode() == graph::DeBruijnGraph::CANONICAL);
```
hmusta, line 282: same

metagraph/src/cli/transform_graph.cpp
```diff
@@ -127,15 +128,28 @@ int transform_graph(Config *config) {
...
     }
 
-    logger->trace("Converting graph to state {}",
-                  Config::state_to_string(config->state));
-    timer.reset();
+    if (config->graph_mode == graph::DeBruijnGraph::PRIMARY
+            && dbg_succ->get_mode() == graph::DeBruijnGraph::BASIC) {
+        logger->info("Changing graph mode from basic to primary");
+        logger->warn("This doesn't rebuild the graph. Apply with caution,"
+                     " only to graphs that were constructed as primary!");
```
hmusta, line 135:
```suggestion
                     " only to graphs that were constructed from primary or canonicalized contigs!");
```

metagraph/src/graph/representation/bitmap/dbg_bitmap.cpp
```diff
@@ -37,7 +37,7 @@ DBGBitmap::DBGBitmap(DBGBitmapConstructor *builder) : DBGBitmap(2) {
 void DBGBitmap::map_to_nodes(std::string_view sequence,
                              const std::function<void(node_index)> &callback,
                              const std::function<bool()> &terminate) const {
-    for (const auto &[kmer, is_valid] : sequence_to_kmers(sequence, canonical_mode_)) {
+    for (const auto &[kmer, is_valid] : sequence_to_kmers(sequence, mode_ != BASIC)) {
```
hmusta, line 40: so DBGBitmap with the `PRIMARY` flag is essentially a graph which only stores the canonical k-mers? maybe we should mention this in the README somewhere

metagraph/src/graph/representation/masked_graph.cpp
```diff
@@ -12,26 +12,34 @@ MaskedDeBruijnGraph
...
         only_valid_nodes_in_mask_(only_valid_nodes_in_mask),
-        is_canonical_(canonical) {
+        mode_(mode) {
     assert(kmers_in_graph_.get());
     assert(kmers_in_graph_->size() == graph->max_index() + 1);
-    assert(!is_canonical_ || graph_->is_canonical_mode());
+
+    if (graph_->get_mode() == DeBruijnGraph::PRIMARY
+            && mode_ != DeBruijnGraph::PRIMARY) {
+        throw std::runtime_error("Any subgraph of a primary graph is primary");
```
hmusta, line 25: should we make this a warning instead of throwing? we can essentially just set the value ourselves and warn the user that we changed it

metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
```diff
@@ -958,13 +953,16 @@ class BOSSChunkConstructor : public IBOSSChunkConstructor {
                          size_t disk_cap_bytes)
         : swap_dir_(swap_dir),
           kmer_collector_(k + 1,
-                          both_strands_mode,
+                          both_strands
+                            ? (filter_suffix.empty()
```
hmusta, line 957: maybe put a comment here to explain why the choice of `KmerCollector::Mode` depends on `filter_suffix`

metagraph/src/kmer/kmer_collector.cpp
```diff
@@ -127,22 +129,20 @@ void count_kmers(std::function<void(CallStringCount)> generate_reads,
...
 ::KmerCollector(size_t k,
-                bool both_strands_mode,
+                Mode mode,
                 std::vector<typename Extractor::TAlphabet>&& filter_suffix_encoded,
                 size_t num_threads,
                 double memory_preallocated,
                 const std::filesystem::path &swap_dir,
-                size_t __attribute__((unused)) disk_cap_bytes,
-                bool canonical_only)
+                size_t __attribute__((unused)) disk_cap_bytes)
```
hmusta, line 137:
```suggestion
                size_t /* disk_cap_bytes */)
```

metagraph/src/cli/parse_sequences.hpp
```diff
@@ -54,7 +54,7 @@ void parse_sequences(const std::string &file,
                 [&](std::string_view, uint32_t count) {
                     count_hist[count] += (1 + config.forward_and_reverse);
                 },
-                !config.canonical && !config.forward_and_reverse
+                config.graph_mode != graph::DeBruijnGraph::CANONICAL && !config.forward_and_reverse
```
hmusta, line 57: if the graph mode is set to PRIMARY and a non-canonicalized/non-primary sequence set is passed, how does this affect the counting? will config.forward_and_reverse be true in this case?

hmusta (changes requested):

metagraph/src/kmer/kmer_collector.cpp
karasikov: it's used for some cases (defined by `if constexpr(...)`)

metagraph/src/graph/representation/bitmap/dbg_bitmap.cpp
karasikov: Yeah, but I think it goes too far.
It's good if we merge the wrapper into the graphs, but right now it's not perfect, because one can't extract contigs from such a primary graph. I'll change this, so it will work as the succinct graph -- take all input sequences and make the user responsible for not passing a k-mer together with its rev-compl.

metagraph/src/cli/clean.cpp
karasikov: I don't see why that's better. We need 2 regimes here, the current boolean flag does the job perfectly.

metagraph/src/cli/build.cpp
karasikov: Because it doesn't make sense. There are just 2 options. Why should we pass a thing which has 3 states?

metagraph/src/cli/build.cpp
```diff
@@ -71,7 +71,7 @@ int build_graph(Config *config) {
 
     Timer timer;
 
-    if (config->canonical)
+    if (config->graph_mode != DeBruijnGraph::BASIC)
```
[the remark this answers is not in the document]
karasikov: ok

metagraph/src/cli/clean.cpp
karasikov: Besides, the graph already knows its state.
Do you propose to remove this flag and the choice which sequences to extract?

metagraph/src/graph/representation/masked_graph.cpp
karasikov: By automatically correcting things that don't make sense, you allow writing code that doesn't make sense.

Just imagine, you would allow writing `MaskedGraph graph(..., BASIC)`, where `graph.get_mode() == BASIC` will return `false` in the next line.

metagraph/src/cli/parse_sequences.hpp
karasikov: This is complicated. I tried to take into account all scenarios.
I'll try to simplify this now and forbid some combinations that don't make sense in applications, to make it more straightforward.

karasikov:
@hmusta can you please check align.cpp and related? It needs to resolve some conflicts.

metagraph/src/graph/representation/bitmap/dbg_bitmap.cpp
karasikov: done

metagraph/integration_tests/test_clean.py
```diff
@@ -207,58 +207,30 @@ def setUp(self):
...
-
-        stats_command = '{exe} stats {graph}'.format(
-            exe=METAGRAPH,
-            graph=self.tempdir.name + '/graph' + graph_file_extension[representation],
-        )
-        res = subprocess.run(stats_command.split(), stdout=PIPE)
-        self.assertEqual(res.returncode, 0)
+        self._build_graph(input=TEST_DATA_DIR + '/transcripts_1000.fa',
+                          output=self.tempdir.name + '/graph',
+                          k=31, repr=representation, canonical=True,
```
hmusta: maybe let's add modes to `_build_graph_` as well

metagraph/src/cli/parse_sequences.hpp
```diff
@@ -34,6 +34,12 @@ void parse_sequences(const std::string &file,
                      CallWeighted call_weighted_sequence) {
     mtg::common::logger->trace("Parsing '{}'", file);
 
+    if (config.graph_mode == graph::DeBruijnGraph::PRIMARY && file_format(file) != "FASTA") {
```
hmusta, line 37: What if the input is a KMC file with only canonical k-mers?

metagraph/src/graph/representation/masked_graph.cpp
hmusta: Yes, true. In that case, how about we change the message here to explicitly say that `mode` must be set to `PRIMARY` when constructed from a primary graph?

karasikov: done

metagraph/src/cli/parse_sequences.hpp
karasikov: In that case, the graph would have too many gaps, which definitely doesn't work for SuccinctDBG.
So I think it's better to require that a primary graph must be always constructed from primary contigs, so the input must be only FASTA.

metagraph/integration_tests/test_clean.py
karasikov: done

hmusta (approved):

PR #287, opened 2021-02-08
faster map_to_nodes in canonical wrapper

karasikov:
If there is a mismatch, try to map the rev-compl sequence. If that has a mismatch there as well, map the rest of the forward and rev-compl sequences.

This change never makes things slower, but often should make mapping way faster.
For annotation, in particular (where usually one of the sequences can be mapped perfectly, and another one is missing), should become a few times faster.

### DIFF
--- a/metagraph/src/graph/representation/canonical_dbg.cpp
+++ b/metagraph/src/graph/representation/canonical_dbg.cpp
@@ -103,26 +103,22 @@ void CanonicalDBG
     if (const auto *dbg_succ = dynamic_cast<const DBGSuccinct*>(&graph_)) {
         // if it's boss, we can skip k-mers that have been found in the rev-compl
         const auto &boss = dbg_succ->get_boss();
-
-        path.assign(sequence.size() - get_k() + 1, npos);
-
         // the initial forward mapping stopped on this k-mer,
         // hence it's missing and we skip it
-        auto jt = path.begin() + 1;
+        path.push_back(npos);
         auto it = rev_path.rbegin() + 1;
         auto is_missing = get_missing_kmer_skipper(dbg_succ->get_bloom_filter(),
                                                    sequence.substr(1));
         boss.map_to_edges(sequence.substr(1),
             [&](boss::BOSS::edge_index edge) {
-                *jt = dbg_succ->boss_to_kmer_index(edge);
-                ++jt;
+                path.push_back(dbg_succ->boss_to_kmer_index(edge));
                 ++it;
             },
             []() { return false; },
             [&]() {
                 if (is_missing() || *it) {
+                    path.push_back(npos);
                     ++it;
-                    ++jt;
                     return true;
                 } else {
                     return false;
@@ -130,7 +126,6 @@ void CanonicalDBG
             }
         );
 
-        assert(jt == path.end());
         assert(it == rev_path.rend());
 
     } else {
--- a/metagraph/src/graph/representation/canonical_dbg.cpp
+++ b/metagraph/src/graph/representation/canonical_dbg.cpp
@@ -62,41 +62,65 @@ void CanonicalDBG
 ::map_to_nodes_sequentially(std::string_view sequence,
                             const std::function<void(node_index)> &callback,
                             const std::function<bool()> &terminate) const {
-    std::vector<node_index> path = map_sequence_to_nodes(graph_, sequence);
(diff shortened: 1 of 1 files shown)


metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -62,79 +62,100 @@ void CanonicalDBG
...
-    auto first_not_found = std::find(path.begin(), path.end(), DeBruijnGraph::npos);
+    if (sequence.size() < get_k())
+        return;
+
+    std::vector<node_index> path;
+    path.reserve(sequence.size() - get_k() + 1);
+
+    // map until the first mismatch
+    bool stop = false;
+    graph_.map_to_nodes_sequentially(sequence,
```
hmusta, line 73: if you use `graph_.traverse` instead, it should be slightly faster since it'll avoid the `get_missing_kmer_skipper` call

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -62,79 +62,100 @@ void CanonicalDBG
...
 
-    std::string rev_seq(sequence.begin() + (first_not_found - path.begin()),
-                        sequence.end());
+    std::string rev_seq(sequence);
     ::reverse_complement(rev_seq.begin(), rev_seq.end());
-    std::vector<node_index> rev_path;
-    if (const auto *dbg_succ = dynamic_cast<const DBGSuccinct*>(&graph_)) {
-        const auto &boss = dbg_succ->get_boss();
+    // map the reverse-complement
+    std::vector<node_index> rev_path = map_sequence_to_nodes(graph_, rev_seq);
```
hmusta, line 100: Wouldn't it be faster to again do a `graph_.traverse` or a `graph_.map_to_nodes_sequentially` call with rev_seq? then you can alternate between calling this on the forward and reverse complement orientation.

I'm not sure if doing things this way is that much faster, since you still end up doing the lookups and not finding k-mers on `rev_seq`.

hmusta (changes requested):

metagraph/src/graph/representation/canonical_dbg.cpp
karasikov: But get_missing_kmer_skipper is supposed to speed it up, that's why we have it.

hmusta: Ah, sorry, `get_missing_kmer_skipper` does help if there's a Bloom filter present.

What I had in mind instead was `utils::drag_and_mark_segments`. This is computed on the entire string, whereas all we need is to keep calling `fwd` and `get_edge` until we get `npos`.

hmusta: Though all of this is minor and switching to `traverse` won't really have a significant effect on mapping time.

metagraph/src/graph/representation/canonical_dbg.cpp
karasikov: To alternate between them, we'd need to synchronously map the opposite ends of the sequences.
This means we'd have to map the rev-compl sequence backwards, which is a bit slower.
So I decided to do it forwards for now. Though in this case, we can't alternate, so I just map the entire thing at once.

One can try later to do it via backward mapping. But that will need benchmarking, and it's not clear if that's going to be faster than the version implemented here.

The current version tries to map from both ends.
Consider these 3 important cases:
* all k-mers in forward are found   => Fast mapping, no misses. Though the old version would do well too.
* all k-mers in forward are missing, but all in reverse are found   => Fast mapping, 1 miss. The old version would do n cold searches of k-mers in forward. So the new version is almost k times faster.
* more general, a prefix of forward and prefix of rev-compl are found    => similar to the previous case
* symmetric: a suffix of forward and suffix of rev-compl are found    => same as before, no speedup

hmusta: thinking about this again, if it's done in an alternating fashion, then it'll end up doing more full-k lookups...

karasikov: During annotation, most of the mapped contigs should belong to case 1 or case 2.
So, this optimization should make it on average about k/2 times faster.

hmusta (approved):

PR #289, opened 2021-02-09
fixed anno transform flags and help messages

karasikov:


metagraph/src/cli/config/config.cpp
```diff
@@ -718,7 +731,8 @@ Config::GraphType Config::string_to_graphtype(const std::string &string) {
 }
 
 void Config::print_usage(const std::string &prog_name, IdentityType identity) {
-    const char annotation_list[] = "('column', 'row', 'bin_rel_wt_sdsl', 'bin_rel_wt', 'flat', 'rbfish', 'brwt', 'rb_brwt', 'row_diff')";
+    const char annotation_list[] = "\t\t( column, row, bin_rel_wt_sdsl, bin_rel_wt, flat, rbfish,\n"
+                                   "\t\t  brwt, rb_brwt, row_diff, row_diff_brwt, row_diff_sparse )";
```
hmusta, line 735: since we're changing this, how about we modify the order to put the ones we want to focus on near the front, like column, row_diff_brwt, etc.

hmusta (approved):

metagraph/src/cli/config/config.cpp
karasikov: what about this?

		( column, row_diff, row_diff_sparse, row_diff_brwt, brwt, rb_brwt,
		  row, flat, rbfish, bin_rel_wt_sdsl, bin_rel_wt )

PR #290, opened 2021-02-11
Feature/static builds using actions

izcram:
* adding static builds in CI
* upon pushing a tag of the form `v*` (e.g. `v0.1`) a new github release is created and binaries for protein and dna for linux are added for users to download via https://github.com/ratschlab/metagraph/releases


.github/workflows/main.yml
```diff
@@ -127,21 +174,68 @@ jobs:
...
+        with:
+          path: artifacts
+      - name: create release with static binaries
+        uses: svenstaro/upload-release-action@v2
+        with:
+          release_name: Release ${{ github.ref }}
+          body: |
+            The binaries are currently provided on an experimental basis.
+            The Linux binaries are statically linked and hence don't require additional libraries to be installed.
+            The Mac OS X binaries dynamically link to libhts, which would need to be available on your system, e.g. using `brew install htslib`). The Mac OS X binaries are currently nor signed nor notarized - you need to explicitely allow the binary to run in the security settings of your system.
```
karasikov, line 236: Can we link against libhts statically?

izcram: We probably could, but libhts has some dependencies like libcurl which make it a bit harder. The strategy would probably be to build libcurl and libhts from sources (the same is actually done for linux).
But IMHO this provides already some value and installing libhts shouldn't be a big burden. My intention is to allocate my time to create official conda packages and docker images. If we still think it's useful to statically link libhts for mac, we can still do it later.

karasikov: I checked it on my laptop and found a static lib `/usr/local/opt/htslib/lib/libhts.a` (installed with brew).

I'll try to fix this, maybe it's easy.
But otherwise, yeah, we can do it later if this is not trivial.

karasikov (approved):

karasikov:
It takes forever to run CI on MacOS... Maybe we should indeed release only the linux binaries...
It's super easy to compile on Mac (with brew) anyway.

PR #295, opened 2021-02-15
Adaptive distance estimation in sparse columns

karasikov:
More efficient subsampling technique for estimating the distance between very sparse columns, adaptive to sparsity.

The old algorithm (with full bitmap subsampling) is used by default.
The new algorithm can be run by adding the `--fast` flag:
`./metagraph transform_anno --linkage --fast ...`.


metagraph/src/annotation/binary_matrix/multi_brwt/clustering.cpp
```diff
@@ -225,9 +259,63 @@ Partition greedy_matching(const std::vector<sdsl::bit_vector> &columns,
...
+        }
+    }
+    while (it_first != col_first.end() && *it_first < size_merged) {
+        col_merged.push_back(*it_first);
+        ++it_first;
+    }
+    while (it_second != col_second.end() && *it_second < size_merged) {
+        col_merged.push_back(*it_second);
+        ++it_second;
+    }
```
hmusta, line 304: why not just use `std::merge`?
```suggestion
    std::merge(col_first.begin(), col_first.end(),
               col_second.begin(), col_second.end(),
               std::back_inserter(col_merged));
```

metagraph/src/cli/transform_annotation.cpp
```diff
@@ -41,6 +41,91 @@ void convert(std::unique_ptr<AnnotatorFrom> annotator,
...
+    uint64_t num_rows = 0;
+
+    ThreadPool subsampling_pool(get_num_threads(), 1);
+
+    // Load columns from disk
+    auto on_column = [&](uint64_t i, const std::string &label,
+                         std::unique_ptr<bit_vector> &&column) {
+        subsampling_pool.enqueue([&, i, label, column{std::move(column)}]() {
+            T *subvector;
+            #pragma omp critical
```
hmusta, line 60: hmm, I haven't seen `#pragma omp critical` being used with an `std::thread` pool. did you double-check this with a thread sanitizer to make sure it's working as intended?

hmusta (changes requested):

metagraph/src/annotation/binary_matrix/multi_brwt/clustering.cpp
```diff
@@ -94,9 +94,41 @@ std::vector<uint64_t> inverted_arrangement(const VectorPtrs &vectors) {
     return { init_arrangement.rbegin(), init_arrangement.rend() };
 }
 
+double inner_prod(const sdsl::bit_vector &first, const sdsl::bit_vector &second) {
+    assert(first.size() == second.size());
+    return static_cast<double>(::inner_prod(first, second)) / first.size();
+}
+
+double inner_prod(const SparseColumn &first, const SparseColumn &second) {
```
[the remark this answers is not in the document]
karasikov: I tried different measures in the past and found this work best.

The first one is indeed inner product, there is no need to rename it.

The second one is an estimation of the inner product divided by the size of the columns.
All of them have the same size (number of annotation rows), so this global scaling doesn't change the order.
I'll try to come up with a better name for that function.

metagraph/src/annotation/binary_matrix/multi_brwt/clustering.cpp
karasikov: std::merge will have duplications

hmusta: ah, I looked into it again and I think `std::set_union` should work then

metagraph/src/cli/transform_annotation.cpp
karasikov: Both of them have the same guarantees. I think externally they are pretty much equivalent.

metagraph/src/annotation/binary_matrix/multi_brwt/clustering.cpp
karasikov: Not exactly. I merge until the minimum size of these two.

Here we merge until the min(size_1, size_2).

I can make it work with std::set_union, but then I'd need to do a scan to find the new `end` first.

Is that worth it? I think the implementation of std::set_union is the same as here, anyway. No?

metagraph/src/annotation/binary_matrix/multi_brwt/clustering.cpp
```diff
@@ -225,9 +267,42 @@ Partition greedy_matching(const std::vector<sdsl::bit_vector> &columns,
...
+    const auto &[size_first, col_first] = first;
+    const auto &[size_second, col_second] = *second;
+
+    SparseColumn merged;
+    auto &[size_merged, col_merged] = merged;
+    size_merged = std::min(size_first, size_second);
+    col_merged.reserve(col_first.size() + col_second.size());
+
+    auto first_end = std::lower_bound(col_first.begin(), col_first.end(), size_merged);
+    auto second_end = std::lower_bound(col_second.begin(), col_second.end(), size_merged);
```
hmusta, line 288:
```suggestion
    auto first_end = size_first == size_merged
        ? col_first.end()
        : std::lower_bound(col_first.begin(), col_first.end(), size_merged);
    auto second_end = size_second == size_merged
        ? col_second.end()
        : std::lower_bound(col_second.begin(), col_second.end(), size_merged);
```
can't you avoid this call in some cases?

hmusta (approved):

metagraph/src/annotation/binary_matrix/multi_brwt/clustering.cpp
karasikov: I think it's not worth it. It's the last value, so the bin-search will immediately find it.

PR #296, opened 2021-02-16
Simpler and more accurate alignment extender

hmusta:


metagraph/src/graph/alignment/aligner_config.cpp
```diff
@@ -10,11 +10,21 @@ namespace align {
...
-    int8_t min_penalty_score = std::min(gap_opening_penalty, gap_extension_penalty);
+    int8_t min_penalty_score = std::numeric_limits<int8_t>::max();
     for (const auto &row : score_matrix_) {
         min_penalty_score = std::min(min_penalty_score, *std::min_element(row.begin(), row.end()));
     }
 
+    if (gap_opening_penalty * 2 >= min_penalty_score) {
+        std::cerr << "|gap_opening_penalty| * 2 should be greater than greatest mismatch penalty: "
+                  << gap_opening_penalty * 2 << " >= " << (score_t)min_penalty_score << std::endl;
+        return false;
```
karasikov, line 21: use spdlog?

metagraph/src/graph/alignment/dbg_aligner.cpp
```diff
@@ -40,6 +40,30 @@ ::align_core(std::string_view query,
...
+                range.first = std::min(range.first, idx_range.first);
+                range.second = std::max(range.second, idx_range.second);
+                inserted = true;
+            }
+        }
+
+        if (!inserted) {
+#ifndef NDEBUG
+            mtg::common::logger->trace("Skipping seed: {}", seed);
+#endif
```
karasikov, line 63: I think we should add a macro and simply write
```c++
DEBUG_LOG("Skipping seed: {}", seed);
```

karasikov (approved):
Is priority queue still used somewhere?

hmusta:
> Is priority queue still used somewhere?

yes, the priority deque is still used to keep track of the top-n endpoints in the DP table from which to start alignment

karasikov:
> > Is priority queue still used somewhere?
> 
> yes, the priority deque is still used to keep track of the top-n endpoints in the DP table from which to start reconstructing alignments (backtracking)

Ok, I was just wondering if it's time to get rid of that dependency :)

Maybe at some point later then...

PR #297, opened 2021-02-22
some crucial improvements in row-diff construction

karasikov:
* allow passing a custom path for disk swap
* enable constructing pred/succ/anchors and optimizing anchors without passing any input columns
* support updating an existing row-reduction vector to run on small chunks without creating a new vector on each run
* compress temp chunks with Elias-Fano to reduce their size by ~4 times


metagraph/src/common/elias_fano.cpp
```diff
@@ -223,6 +223,15 @@ EliasFanoEncoder<T>::~EliasFanoEncoder() {
     assert(!sink_internal_.is_open());
 }
 
+template <typename T>
+void EliasFanoEncoder<T>::append(const std::vector<T> &data,
+                                 const std::string &out_fname) {
+    std::ofstream sink(out_fname, std::ios::binary | std::ios::app);
+    std::ofstream sink_upper(out_fname + ".up", std::ios::binary | std::ios::app);
+    EliasFanoEncoder<T> encoder(data, &sink, &sink_upper);
+    encoder.finish();
```
hmusta, line 232: With the current implementation, it looks like it's only possible to append once since it finishes the encoder afterwards. This seems like an artificial limitation.

Why not call `finish` externally instead?

metagraph/src/cli/transform_annotation.cpp
```diff
@@ -46,6 +46,16 @@ int transform_annotation(Config *config) {
 
     const auto &files = config->fnames;
 
+    if (config->anno_type == Config::RowDiff && !files.size()) {
+        logger->trace("Passed no input annotations to transform. Prepare to RowDiff transform...");
```
hmusta, line 50: This message is kind of unintuitive, why not have it say something like "Preparing to RowDiff transform in output directory" or something like that"

metagraph/src/cli/config/config.cpp
```diff
@@ -442,6 +442,7 @@ Config::Config(int argc, char *argv[]) {
     if (identity != CONCATENATE
             && identity != STATS
             && identity != SERVER_QUERY
+            && !(identity == TRANSFORM_ANNOTATION && anno_type == Config::RowDiff)
             && !(identity == BUILD && complete)
             && !fnames.size())
         print_usage_and_exit = true;
```
hmusta, line 448: we should add a message here to explain that at least one file needs to be passed

hmusta (changes requested):

metagraph/src/common/elias_fano.cpp
karasikov: This thing writes in blocks, finish flushed the buffers.
Multiple appends are supported, no issue with that.

hmusta: Maybe it would be good to have a unit test for this.

https://github.com/ratschlab/metagraph/blob/b4c7aab2820e40a2b7543e83d009b6cc45276e17/metagraph/src/common/elias_fano.cpp#L274
On this line, there's a check to make sure that all elements have been added, so wouldn't this assert false?

hmusta: Unless there's something I'm missing here...

karasikov: It just pushed another block to the file.
I don't see what confuses you. Can you explain what seems to you?

karasikov: I mean, declared_size_ and size_ for this next block will be different. The blocks are independent. They just sit back to back in the file.

hmusta: ok, my bad. for some reason I thought it was appending to an existing block, but yes, since a new object is created, I see that it's a new independent block

hmusta (approved):

PR #301, opened 2021-03-03
push dual nodes to the back of the queue

karasikov:


metagraph/src/graph/representation/succinct/boss.cpp
```diff
@@ -2569,7 +2582,7 @@ call_path(const BOSS &boss,
                 // otherwise it's a fork which will be covered in the forward pass.
                 if (masked_pick_single_outgoing(boss, &next_edge, subgraph_mask)) {
                     if (!fetch_bit(visited.data(), next_edge, concurrent)) {
-                        edge_queue->emplace_back(next_edge,
+                        edge_queue->emplace_front(next_edge,
                                                  rev_comp_seq.begin() + i + 1,
                                                  rev_comp_seq.begin() + i + 1 + boss.get_k());
```
hmusta, line 2587:
```suggestion
                                                  rev_comp_seq.begin() + i + 1,
                                                  rev_comp_seq.begin() + i + 1 + boss.get_k());
```

hmusta (approved):
I'm guessing this is to minimize the chances of two threads trying to cover a path and its dual at the same time?

karasikov:
> I'm guessing this is to minimize the chances of two threads trying to cover a path and its dual at the same time?

Yes, this should make it a bit more connected. Though, I didn't see any difference in my tests, unfortunately.

But I think we should merge this anyway, conceptually this makes sense, even though the effect is little.

PR #309, opened 2021-04-06
query count annotation

karasikov:
* added support for querying k-mer counts (with `--count-kmers`)
* counts are stored in arrays and indexed by calling rank on columns of the base matrix (currently, Multi-BRWT)
* the construction is done as usually, via column-compressed annotation with the same layout of k-mer counts


hmusta:
Instead of making this specific to BRWT, why don't we instead introduce a trivial implementation of `get_row_ranks` for `BinaryMatrix` and allow this work for all annotation types?

metagraph/src/annotation/representation/annotation_matrix/static_annotators_def.hpp
```diff
@@ -61,6 +64,9 @@ template <>
 inline const std::string RowDiffBRWTAnnotator::kExtension = ".row_diff_brwt.annodbg";
 template <>
 inline const std::string RowDiffRowSparseAnnotator::kExtension = ".row_diff_sparse.annodbg";
+template <>
+inline const std::string ExtendedMultiBRWTAnnotator::kExtension = ".count_brwt.annodbg";
```
hmusta, line 68: Wouldn't it be much more extensible to have counts be stored separately as an int_vector?

metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
```diff
@@ -189,6 +247,30 @@ std::vector<BRWT::Column> BRWT::slice_rows(const std::vector<Row> &row_ids) cons
...
+            = slice_rows<std::pair<Column, uint64_t>>(row_ids);
+
+    assert(slice.size() >= row_ids.size());
+
+    auto row_begin = slice.begin();
+
+    for (size_t i = 0; i < row_ranks.size(); ++i) {
+        // every row in `slice` ends with `-1`
+        auto row_end = row_begin;
+        while (row_end->first != std::numeric_limits<Column>::max()) {
```
hmusta, line 264: add an assert to ensure that `row_end` never becomes `slice.end()`

hmusta (changes requested):

metagraph/src/annotation/representation/annotation_matrix/static_annotators_def.hpp
karasikov: Also answering to this:
> Instead of making this specific to BRWT, why don't we instead introduce a trivial implementation of `get_row_ranks` for `BinaryMatrix` and allow this work for all annotation types?

Only column-major annotations support rank on columns, so we can't implement `get_row_ranks` for all annotations.
For instance, the `row_flat` annotation would call rank on rows, hence, the layout of the k-mer counts would have to be different as well.

I thought about storing the counts always separately, but saw no advantages of that right now.
On the other hand, keeping them together with the matrix is easier because it never becomes out of sync.
If there is a need, we can easily change this later.

metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
```diff
@@ -58,38 +58,38 @@ BRWT::SetBitPositions BRWT::get_row(Row row) const {
...
-        return row_ranks;
+        return row;
     }
 
     // check all child nodes
     uint64_t index_in_child = rank - 1;
 
-    for (size_t i = 0; i < child_nodes_.size(); ++i) {
-        const auto &child = *child_nodes_[i];
+    for (size_t k = 0; k < child_nodes_.size(); ++k) {
```
hmusta, line 85: minor point, maybe call this `j` to avoid confusing it with the k-mer size

metagraph/src/annotation/int_matrix/extended/csc_matrix.hpp
```diff
@@ -57,62 +57,62 @@ class ColumnRankMatrix : public CountMatrix {
...
 
 template <class BaseMatrix>
-inline typename ColumnRankMatrix<BaseMatrix>::ColumnCounts
-ColumnRankMatrix<BaseMatrix>::get_row_counts(Row row) const {
-    const auto &column_ranks = binary_matrix_.get_row_ranks(row);
-    ColumnCounts row_counts;
-    row_counts.reserve(column_ranks.size());
+inline typename CSCMatrix<BaseMatrix>::RowValues
+CSCMatrix<BaseMatrix>::get_row_values(Row row) const {
+    const auto &column_ranks = binary_matrix_.get_column_ranks(row);
```
hmusta, line 67: `binary_matrix_.get_column_ranks` returns an r-value, won't this cause issues?
```suggestion
    auto column_ranks = binary_matrix_.get_column_ranks(row);
```

metagraph/src/annotation/int_matrix/extended/csc_matrix.hpp
```diff
@@ -57,62 +57,62 @@ class ColumnRankMatrix : public CountMatrix {
...
 }
 
 template <class BaseMatrix>
-inline std::vector<typename ColumnRankMatrix<BaseMatrix>::ColumnCounts>
-ColumnRankMatrix<BaseMatrix>::get_row_counts(const std::vector<Row> &rows) const {
-    const auto &column_ranks = binary_matrix_.get_row_ranks(rows);
-    std::vector<ColumnCounts> row_counts(rows.size());
+inline std::vector<typename CSCMatrix<BaseMatrix>::RowValues>
+CSCMatrix<BaseMatrix>::get_row_values(const std::vector<Row> &rows) const {
+    const auto &column_ranks = binary_matrix_.get_column_ranks(rows);
```
hmusta, line 80: Same comment as above
```suggestion
    auto column_ranks = binary_matrix_.get_column_ranks(rows);
```

hmusta (changes requested):

metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
```diff
@@ -57,6 +58,40 @@ BRWT::SetBitPositions BRWT::get_row(Row row) const {
...
+
+    Vector<std::pair<BRWT::Column, uint64_t>> row;
+
+    // check whether it is a leaf
+    if (!child_nodes_.size()) {
+        assert(assignments_.size() == 1);
+
+        // the bit is set
+        for (size_t j = 0; j < assignments_.size(); ++j) {
+            row.emplace_back(j, rank);
```
hmusta, line 77: is `row` every resized later? if not, reserve space

hmusta (changes requested):

metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
karasikov: I use `j` for indexing columns. This is the annotation module, there are no k-mers here.
So `k` is used to index children in a tree (from "k-ary tree")

metagraph/src/annotation/int_matrix/extended/csc_matrix.hpp
karasikov: Assigning const references to returned objects is always correct.
https://stackoverflow.com/questions/7340925/return-by-value-assigned-to-const-reference/7340943

I think it's a bit better because this implies that the object is not changed in this function, hence the function can, in principle, return const references as well, and no copy would be made in that case.

metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
karasikov: I don't know it's size. Setting it to num_columns() will hugely overestimate it in most of the cases.
So without the following shrink_to_fit this may accumulate and take in total tonnes of memory.

Also, I think it won't make a difference in the runtime anyway.

hmusta: For this base case, we know that `row`'s final size is gonna be `assignments_.size()`.

I'm not really concerned about runtime, it's that it might be less likely to use the small vector optimizations if the vector is not pre-allocated

karasikov: > For this base case, we know that row's final size is gonna be assignments_.size()

... which is equal to `num_columns()`

hmusta: actually, from the assertion, we know that it's final size is gonna be 1, so we can even just return `{ { 0, rank } }`

karasikov: Ok, this was written for a more general case, but yeah, currently there can only be a single column in leaves.

Though, I am sure this will have a zero effect on the memory or CPU usage...

PR #312, opened 2021-04-10
parallel small->stat transform

karasikov:
Conversion from small representations is very slow on a single thread, so we should make it parallel

### DIFF
--- a/metagraph/src/common/vectors/wavelet_tree.cpp
+++ b/metagraph/src/common/vectors/wavelet_tree.cpp
@@ -302,7 +302,9 @@ void wavelet_tree_dyn::clear() {
 
 sdsl::int_vector<> wavelet_tree_dyn::to_vector() const {
     sdsl::int_vector<> vector(dwt_.size(), 0, logsigma());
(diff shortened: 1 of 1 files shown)


metagraph/src/common/vectors/wavelet_tree.cpp
```diff
@@ -392,7 +393,10 @@ bool wavelet_tree_sdsl<t_wt_sdsl>::load(std::istream &in) {
 template <class t_wt_sdsl>
 sdsl::int_vector<> wavelet_tree_sdsl<t_wt_sdsl>::to_vector() const {
     sdsl::int_vector<> vector(wwt_.size(), 0, logsigma_);
-    std::copy(wwt_.begin(), wwt_.end(), vector.begin());
+    #pragma omp parallel for num_threads(get_num_threads()) schedule(static, 1024)
```
hmusta, line 396: maybe add a comment here that the block size (1024) has to be a multiple of 64, otherwise this loop will lead to race conditions

hmusta (approved):

PR #313, opened 2021-04-14
Add `print_signature` support to server `/search` endpoint

boydgreenfield:
👋  Hi there – very cool project. I'm experimenting with a use case where this is helpful for visually indicating which k-mers match the DBG vs. which are misses – the CLI `--print-signature` option accomplishes this, but there's no way to currently do that with the server.

This small patch adds support for sending a `print_signature: True` argument to the `/search` server endpoint (defaults to `False`), including in the Python client.


karasikov:
Thanks!

boydgreenfield:
> @karasikov karasikov changed the base branch from master to dev 9 hours ago

👍 . Thanks!

PR #314, opened 2021-04-16
improvements in column aggregation

karasikov:


metagraph/src/cli/transform_annotation.cpp
```diff
@@ -474,27 +483,70 @@ int transform_annotation(Config *config) {
...
 
-        sdsl::bit_vector intersection(sum.size(), false);
+        sdsl::bit_vector mask(sum.size(), false);
         for (uint64_t i = 0; i < sum.size(); ++i) {
-            if (sum[i] >= min_cols) {
-                intersection[i] = true;
+            if (sum[i] >= min_cols && sum[i] <= max_cols) {
+                mask[i] = true;
             }
         }
```
hmusta, line 494: you can use `call_nonzeros` from `vector_algorithm.hpp` to speed this up.

metagraph/src/cli/transform_annotation.cpp
```diff
@@ -434,15 +434,24 @@ int transform_annotation(Config *config) {
...
 
+        const uint64_t min_cols
+            = std::max((uint64_t)std::ceil(num_columns * config->min_fraction),
+                       (uint64_t)config->min_count);
+        const uint64_t max_cols
+            = std::min((uint64_t)std::floor(num_columns * config->max_fraction),
+                       (uint64_t)config->max_count);
+
+        // TODO: set width to log2(max_cols + 1) but make sure atomic
```
hmusta, line 453: The atomic increment will overflow if the slots are not wide enough. The only way to avoid this is to rewrite `atomic_fetch_and_add` to use a compare-and-swap loop, which would make the operations more expensive.

metagraph/src/cli/config/config.cpp
```diff
@@ -1090,8 +1094,13 @@ void Config::print_usage(const std::string &prog_name, IdentityType identity) {
...
-            fprintf(stderr, "\t   --intersect-columns \t\tcompute intersection of the annotation columns [off]\n");
-            fprintf(stderr, "\t   --intersect-ratio [FLOAT] \tinclude k-mer if it appears in this ratio of columns [1.0]\n");
+            fprintf(stderr, "\t   --aggregate-columns \t\taggregate annotation columns into a bitmask (new column) [off]\n");
+            fprintf(stderr, "\t   --anno-label [STR]\t\tname of the aggregated output column [mask]\n");
+            fprintf(stderr, "\t   --min-count [INT] \t\texclude k-mers appearing in fewer than this number of columns [1]\n");
+            fprintf(stderr, "\t   --min-fraction [FLOAT] \texclude k-mers appearing in fewer than this fraction of columns [0.0]\n");
+            fprintf(stderr, "\t   --max-count [INT] \t\texclude k-mers appearing in more than this number of columns [inf]\n");
+            fprintf(stderr, "\t   --max-fraction [FLOAT] \texclude k-mers appearing in more than this fraction of columns [1.0]\n");
+            fprintf(stderr, "\t   --intersect [STR] \tbitmask to intersect with annotation columns (compute inner product [off]\n");
```
hmusta, line 1103: it's really not clear from this description that this is supposed to take a file name as input

hmusta (changes requested):

metagraph/src/cli/config/config.cpp
```diff
@@ -432,8 +436,8 @@ Config::Config(int argc, char *argv[]) {
         print_usage_and_exit = true;
     }
 
-    if (intersect_ratio < 0 || intersect_ratio > 1) {
-        std::cerr << "Error: intersection ratio must be in range [0, 1]"
+    if (min_fraction < 0 || min_fraction > 1 || max_fraction < 0 || max_fraction > 1) {
```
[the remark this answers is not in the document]
karasikov: I think that should be possible. Why not.

By default, `min_count = 1`, so it will be 0 only if you specifically set it so.

metagraph/src/cli/transform_annotation.cpp
karasikov: Yes, so this is not trivial, that's why I added a TODO.

I think it's worth it, we would save a lot of RAM which is nice.
We just need to check if the slowdown is not too bad.

metagraph/src/cli/transform_annotation.cpp
karasikov: hm, not sure that makes sense. It's just a single pass, relative to summing up all columns, this should be cheap.
Also, this won't work for zeros, so I'll have to write a special case for `min_cols = 0`

metagraph/src/cli/config/config.cpp
karasikov: Thanks, I'll fix this

metagraph/src/cli/transform_annotation.cpp
hmusta: In this case, if the user sets `min_cols = 0` and `max_cols = inf`, then there's no need to even read the annotators and compute `sum`, we can just initialize `mask` to be all ones

hmusta (approved):

PR #322, opened 2021-05-23
optimized parallelization: annotate sequence batches

karasikov:
* accumulate batches before annotating to reduce the overhead in the thread pool
* `add_label_counts` is now thread-safe
* added `annotate_sequences` for efficiently annotating batches in parallel
* streamlined ColumnMajor

### DIFF
--- a/metagraph/src/cli/annotate.cpp
+++ b/metagraph/src/cli/annotate.cpp
@@ -218,6 +218,7 @@ void annotate_data(std::shared_ptr<graph::DeBruijnGraph> graph,
 
     ThreadPool thread_pool(get_num_threads() > 1 ? get_num_threads() : 0);
 
+    // not too small, not too large
     const size_t batch_size = 1'000;
     const size_t batch_length = 100'000;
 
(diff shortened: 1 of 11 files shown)


metagraph/src/graph/annotated_dbg.cpp
```diff
@@ -70,12 +70,50 @@ ::annotate_sequence(std::string_view sequence,
...
+        auto &indices = data[t].second == data[last].second ? ids[last] : ids[t];
+        indices.reserve(data[t].first.size());
+
+        graph_->map_to_nodes(data[t].first, [&](node_index i) {
+            if (i > 0)
+                indices.push_back(graph_to_anno_index(i));
+        });
+    }
+
+    std::lock_guard<std::mutex> lock(mutex_);
```
hmusta, line 90: Since the thread safety of annotators really depends on how they're implemented, how about we give the dynamic annotators their own mutexes and move this lock into the `add_labels` calls?

hmusta (approved):

metagraph/src/cli/annotate.cpp
```diff
@@ -216,8 +218,19 @@ void annotate_data(std::shared_ptr<graph::DeBruijnGraph> graph,
 
     ThreadPool thread_pool(get_num_threads() > 1 ? get_num_threads() : 0);
 
+    const size_t batch_size = 1'000;
+    const size_t batch_length = 100'000;
```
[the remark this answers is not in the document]
karasikov: There is no need for that, these are small enough to keep the overhead tiny and large enough to do the job.

metagraph/src/graph/annotated_dbg.cpp
karasikov: Currently, it only makes sense for add_kmer_counts to compute ranks in parallel.
For add_labels this doesn't make a difference. May be implemented at some point later.

PR #324, opened 2021-05-30
batch query for k-mer counts

karasikov:


metagraph/src/cli/query.cpp
```diff
@@ -313,6 +314,29 @@ slice_annotation(const AnnotatedDBG::Annotator &full_annotation,
...
 
+    if (const auto *mat = dynamic_cast<const IntMatrix *>(&full_annotation.get_matrix())) {
+        std::vector<uint64_t> row_indexes;
+        row_indexes.reserve(full_to_small.size());
+        for (const auto &[in_full, _] : full_to_small) {
+            assert(in_full < full_annotation.num_objects());
+            row_indexes.push_back(in_full);
+        }
+
+        auto slice = mat->get_row_values(row_indexes);
```
hmusta, line 325: how would this perform if it was done in batches with multiple threads?

hmusta (changes requested):

metagraph/src/annotation/int_matrix/csr_matrix/csr_matrix.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+// number of non-zero entries in the matrix
+uint64_t CSRMatrix::num_relations() const {
+    return std::accumulate(
+        vector_.begin(), vector_.end(), (uint64_t)0,
+        [](uint64_t sum, const auto &v) { return sum + v.size(); }
+    );
+}
+
+bool CSRMatrix::get(Row row, Column column) const {
```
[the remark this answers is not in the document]
karasikov: No, that woud be for `get_value`.
`get` is inherited from the interface of `BinaryMatrix`.

metagraph/src/cli/query.cpp
karasikov: One would need to test that. My focus was the single-thread mode.
Multithreading can be implemented inside `get_row_values` or here, as with binary annotations. One would need to do these experiments to make it parallel.

PR #329, opened 2021-06-16
support for multiple successors at forks

karasikov:


metagraph/src/annotation/annotation_converters.cpp
```diff
@@ -1172,6 +1173,20 @@ void convert_to_row_diff(const std::vector<std::string> &files,
...
         mem_bytes -= anchor_size;
+
+        const std::string rd_succ_fname = graph_fname + kRowDiffForkSuccExt;
+        if (!fs::exists(rd_succ_fname)) {
+            logger->error("Can't find row-diff successor bitmap at {}", rd_succ_fname);
+            exit(1);
+        }
+        uint64_t rd_succ_size = fs::file_size(rd_succ_fname);
+        if (rd_succ_size > mem_bytes) {
+            logger->warn("row-diff successor bitmap ({} MiB) is larger than"
```
hmusta, line 1184: Why is this just a warning instead of an error?

metagraph/src/graph/representation/succinct/dbg_succinct.hpp
```diff
@@ -191,6 +200,53 @@ class DBGSuccinct : public DeBruijnGraph {
     std::unique_ptr<mtg::kmer::KmerBloomFilter<>> bloom_filter_;
 };
 
+inline DBGSuccinct::node_index
+DBGSuccinct::row_diff_successor(node_index node, const bit_vector &rd_succ) const {
```
hmusta, line 204: why not make this a simple wrapper for `call_row_diff_successors`?

hmusta (changes requested):

metagraph/src/graph/representation/succinct/boss.hpp
```diff
@@ -153,17 +153,18 @@ class BOSS {
      */
     void row_diff_traverse(size_t num_threads,
                            size_t max_length,
-                           const bit_vector &rd_succ,
+                           const std::function<bool(edge_index)> &is_successor,
```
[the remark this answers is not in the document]
karasikov: I don't foresee the need of passing bit vectors or bitmaps to this function. A simple callback does the job and it's a simpler entity than `bitmap_lazy`. Hence, it's a better choice for this, I think.

metagraph/src/annotation/annotation_converters.cpp
karasikov: I think the idea is to let it run if the process doesn't get killed.

metagraph/src/graph/representation/succinct/dbg_succinct.hpp
karasikov: this thing calls just one successor and `call_row_diff_successors` calls all of them.

metagraph/src/graph/representation/succinct/dbg_succinct.cpp
```diff
@@ -1012,5 +1015,58 @@ void DBGSuccinct::print(std::ostream &out) const {
...
+            queue.pop_back();
+            // stop branch if anchor is reached
+            if (anchor[edge])
+                continue;
+
+            assert(W[edge] % boss_graph_->alph_size);
+
+            boss_graph_->call_outgoing(boss_graph_->fwd(edge, W[edge] % boss_graph_->alph_size),
+                [&](BOSS::edge_index next) {
+                    if (!fetch_bit(excluded.data(), next, async))
```
hmusta, line 1058: Would it make sense here to do `fetch_and_set_bit` here to better stop other threads from exploring this path?

metagraph/src/graph/representation/succinct/dbg_succinct.cpp
```diff
@@ -1012,5 +1015,58 @@ void DBGSuccinct::print(std::ostream &out) const {
...
+                        queue.push_back(next);
+                }
```
hmusta, line 1060: How often does this explore the same edges multiple times? I'm thinking there could be a better implementation of this traversal using ring buffers of length `max_length`. But, if this doesn't do too much extra traversal, this is a simpler approach.

PR #332, opened 2021-07-12
Various minor fixes and tweaks

hmusta:


metagraph/src/cli/align.cpp
```diff
@@ -393,12 +394,12 @@ int align_to_graph(Config *config) {
...
-    if (graph->get_mode() == DeBruijnGraph::PRIMARY) {
-        logger->trace("Primary graph wrapped into canonical");
-        graph = std::make_shared<CanonicalDBG>(graph);
-    }
-
     if (config->map_sequences) {
+        if (graph->get_mode() == DeBruijnGraph::PRIMARY) {
+            logger->trace("Primary graph wrapped into canonical");
+            graph = std::make_shared<CanonicalDBG>(graph);
+        }
```
karasikov, line 401: Why only for `map_sequences` and not always wrap the graph?

metagraph/src/cli/align.cpp
```diff
@@ -500,15 +501,18 @@ int align_to_graph(Config *config) {
...
 
                 logger->trace("Num minibatches: {}, minibatch size: {} KB",
                               num_minibatches, mbatch_size / 1e3);
             } else {
-                thread_pool.enqueue(process_batch, std::move(seq_batch), batch_size);
+                if (it != end) {
+                    thread_pool.enqueue(process_batch, std::move(seq_batch));
+                } else {
+                    process_batch(std::move(seq_batch));
+                }
```
karasikov, line 515: why not pass it to the pool as before?

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -195,21 +197,32 @@ void CanonicalDBG::append_next_rc_nodes(node_index node,
...
                 c = kmer::KmerExtractorBOSS::complement(c);
 
-                if (children[c] == npos)
+                if (children[c] != npos) {
+                    if (k_odd_) {
+                        common::logger->error(
+                            "Primary graph contains both forward and reverse complement: {} {} -> {} {}\t{} {}",
+                            node, graph_.get_node_sequence(node),
+                            children[c], graph_.get_node_sequence(children[c]),
+                            next, graph_.get_node_sequence(next));
```
karasikov, line 206: I think printing thousands of error messages is not helpful. Just throw an exception or make sure the behavior is correct, and hence, there is no need to print error messages. Otherwise, it's a half-correct function, isn't it?

karasikov (changes requested):

metagraph/src/graph/representation/canonical_dbg.cpp
hmusta: I think throwing an exception here is better since this message is only supposed to be printed if the input graph is invalid.

metagraph/src/cli/align.cpp
hmusta: I'll put a comment here to explain this. Later in the alignment code, I create a separate `CanonicalDBG` wrapper for each thread to avoid contention when accessing cached data.

karasikov: What about re-wrapping it there?

metagraph/src/graph/representation/canonical_dbg.cpp
karasikov: Yes, in that case just print a one-line exception.

metagraph/src/cli/align.cpp
hmusta: IIRC I was trying to debug something a while back. I don't need this anymore

metagraph/src/cli/align.cpp
hmusta: The graph is wrapped in all cases. It's just that for the alignment (non-`map_sequences`) case, I defer the wrapping until after a thread is spawned so I can keep a separate `CanoncialDBG` for each thread.

If the graph is wrapped from the beginning, then I'd have to do a `dynamic_cast` check, unwrap, then rewrap in the alignment section, which is a bit ugly.

metagraph/src/cli/align.cpp
```diff
@@ -211,15 +211,15 @@ void map_sequences_in_file(const std::string &file,
...
+            size_t num_unique_matching_kmers = graphindices.size() ? std::inner_product(
                 graphindices.begin() + 1, graphindices.end(),
                 graphindices.begin(),
                 size_t(graphindices.front() != DeBruijnGraph::npos),
                 std::plus<size_t>(),
                 [](DeBruijnGraph::node_index next, DeBruijnGraph::node_index prev) {
                     return next != DeBruijnGraph::npos && next != prev;
                 }
-            );
+            ) : 0;
```
karasikov, line 222: Wouldn't a simple loop look simpler? Simething like this?
```c++
            size_t num_unique_matching_kmers = 0;
            auto prev = DeBruijnGraph::npos;
            for (auto i : graphindices) {
                if (i && i != prev)
                    num_unique_matching_kmers++;
            }
```

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -195,21 +197,33 @@ void CanonicalDBG::append_next_rc_nodes(node_index node,
...
+                if (children[c] != npos) {
+                    if (!k_odd_) {
+                        is_palindrome_cache_.Put(next, true);
+                    } else {
+                        throw std::runtime_error(fmt::format(
+                            "Primary graph contains both forward and reverse complement: {} {} -> {} {}\t{} {}",
+                            node, graph_.get_node_sequence(node),
+                            children[c], graph_.get_node_sequence(children[c]),
+                            next, graph_.get_node_sequence(next)
+                        ));
```
karasikov, line 209: why not the conventional `logger->error` + `exit`?

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -195,21 +197,33 @@ void CanonicalDBG::append_next_rc_nodes(node_index node,
 
                 c = kmer::KmerExtractorBOSS::complement(c);
 
-                if (children[c] == npos)
+                if (children[c] != npos) {
+                    if (!k_odd_) {
+                        is_palindrome_cache_.Put(next, true);
+                    } else {
+                        throw std::runtime_error(fmt::format(
```
karasikov, line 204: in such cases, it's better to check for errors first:
```c++
if (k_odd_) {
    logger->error(...);
    exit(1);
}

is_palindrome_cache_.Put(next, true);
```

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -195,21 +197,33 @@ void CanonicalDBG::append_next_rc_nodes(node_index node,
...
-            rev_seq[0] = complement(alphabet[c]);
-            node_index next = graph_.kmer_to_node(rev_seq);
-            if (next != npos)
-                children[c] = next + offset_;
+            if (alphabet[c] != boss::BOSS::kSentinel && children[c] == npos) {
+                rev_seq[0] = complement(alphabet[c]);
+                node_index next = graph_.kmer_to_node(rev_seq);
+                if (next != npos)
+                    children[c] = next + offset_;
+            }
```
karasikov, line 226: writing some short comments would be nice

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -301,21 +314,33 @@ void CanonicalDBG::append_prev_rc_nodes(node_index node,
 
                     c = kmer::KmerExtractorBOSS::complement(c);
 
-                    if (parents[c] == npos)
+                    if (parents[c] != npos) {
+                        if (!k_odd_) {
+                            is_palindrome_cache_.Put(prev, true);
+                        } else {
+                            throw std::runtime_error(fmt::format(
```
karasikov, line 321: swap the order: check for errors first

metagraph/src/cli/align.cpp
karasikov: I see. But that might be not the only place where you want to isolate caches of two graph views.
A more general approach for handling such cases is wrapping the graph first and then create a new copy of the wrapped graph for each thread where you create it:
```
graph
if (graph is wrapped) {
    graph = new canonical_dbg(graph.base_graph());
}
use graph
```

karasikov (changes requested):

metagraph/src/graph/representation/canonical_dbg.cpp
hmusta: done

metagraph/src/graph/representation/canonical_dbg.cpp
hmusta: I changed it so all errors which derive from bad user input use the logger, whereas I kept the exception throw in one place where it can only be caused by programmer error.

metagraph/src/cli/align.cpp
hmusta: I didn't do this originally because `CanonicalDBG` doesn't have a copy constructor. I added one.

metagraph/src/graph/representation/canonical_dbg.cpp
karasikov: This can also be caused by a user's error. You can mark a non-primary graph as primary and then this error will pop up.

karasikov: For errors caused by a programmer's mistake use assertions.
There is no point to have three: exception / error+exit / assert.

hmusta: So I changed this line to be a logger error. The only place where it still throws is in `add_sequence`, where it throws if the input graph is const.

hmusta: anyway, I changed that to an assert as well

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -195,21 +211,36 @@ void CanonicalDBG::append_next_rc_nodes(node_index node,
...
+                            next, graph_.get_node_sequence(next)
+                        );
+                        exit(1);
+                    }
+
+                    is_palindrome_cache_.Put(next, true);
+
+                } else {
                     children[c] = next + offset_;
+                }
```
karasikov, line 229:
```suggestion
                if (children[c] == npos) {
                    children[c] = next + offset_;
                    return;
                }

                if (k_odd_) {
                    logger->error(
                        "Primary graph contains both forward and reverse complement: {} {} -> {} {}\t{} {}",
                        node, graph_.get_node_sequence(node),
                        children[c], graph_.get_node_sequence(children[c]),
                        next, graph_.get_node_sequence(next));
                    exit(1);
                }

                is_palindrome_cache_.Put(next, true);
```

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -301,21 +331,36 @@ void CanonicalDBG::append_prev_rc_nodes(node_index node,
...
+                                prev, graph_.get_node_sequence(prev)
+                            );
+                            exit(1);
+                        }
+
+                        is_palindrome_cache_.Put(prev, true);
+
+                    } else {
                         parents[c] = prev + offset_;
+                    }
```
karasikov, line 349: same, too many if - else.

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -48,13 +65,13 @@ uint64_t CanonicalDBG::num_nodes() const {
 void CanonicalDBG
 ::add_sequence(std::string_view sequence,
                const std::function<void(node_index)> &on_insertion) {
-    if (!graph_ptr_)
-        throw std::runtime_error("add_sequence only supported for non-const graphs.");
+    assert(graph_ptr_ && "add_sequence only supported for non-const graphs.");
```
karasikov, line 68: Are you sure this is checked somewhere in parents and there is no way to get this wrong with certain commands in CLI?

`assert` must only be called only if there is absolutely no way to get that error once it's compiled.

karasikov (changes requested):

metagraph/src/graph/representation/canonical_dbg.cpp
karasikov: Ok, we do use exceptions sometimes.

For instance, in that case, it should in principle never happen, so an assert would work, but it's really hard to trace the logic down to parsing commands in CLI, so I would not guarantee that it never happens and throw exceptions instead of writing assertions in such cases.

I hope this makes it clear.

metagraph/src/graph/representation/canonical_dbg.cpp
hmusta: I know. As far as I know, the `add_sequence` is only called in build, augment, and when constructing a query graph. CanonicalDBG is never used for these.

karasikov: if you're sure, you can keep the assert.

karasikov (approved):

metagraph/src/cli/align.cpp
```diff
@@ -211,15 +211,12 @@ void map_sequences_in_file(const std::string &file,
...
-                [](DeBruijnGraph::node_index next, DeBruijnGraph::node_index prev) {
-                    return next != DeBruijnGraph::npos && next != prev;
-                }
-            );
+            size_t num_unique_matching_kmers = 0;
+            auto prev = DeBruijnGraph::npos;
+            for (auto i : graphindices) {
+                if (i != DeBruijnGraph::npos && i != prev)
+                    ++num_unique_matching_kmers;
+            }
```
karasikov, line 219: hold on, `prev` is never updated. How does this pass tests?
```suggestion
            auto prev = DeBruijnGraph::npos;
            for (auto i : graphindices) {
                if (i != DeBruijnGraph::npos && i != prev)
                    ++num_unique_matching_kmers;
                prev = i;
            }
```

karasikov (changes requested):

metagraph/src/cli/align.cpp
hmusta: ah, I think there's no integration test for this

karasikov: Could you add one or two?

Maybe first before fixing this error, to make sure they would fail.

hmusta: ok, added

karasikov (approved):
ok, let's merge once it compiles fine

PR #333, opened 2021-07-12
Alignment refactor and rewrite

hmusta:
Features
- Refactor the aligner classes to make them more extensible
- Make UniMEMSeeder and SuffixSeeder work for different seed sizes, so the seeder doesn't have to be specified in align.cpp
- Rewrite DefaultColumnExtender to shrink the size of the dynamic programming table and allow it to converge faster
- Added RCDBG as a wrapper to allow for traversal backwards from a seed in Basic mode graphs

Broken off into other commits:
- DBGWrapper (#340)
- get_base_graph (#340)
- Removing templates from Alignment (commits 85d66a3c0ebf24d80ec1b70f5fe9ce5b5150024b to c083a6e6743acf9a1ebecbaadd222dffbe49d220)


metagraph/src/cli/config/config.cpp
```diff
@@ -198,6 +198,8 @@ Config::Config(int argc, char *argv[]) {
             dump_text_anno = true;
         } else if (!strcmp(argv[i], "--discovery-fraction")) {
             discovery_fraction = std::stof(get_value(i++));
+        } else if (!strcmp(argv[i], "--align-fraction-of-top")) {
+            alignment_fraction_of_top = std::stof(get_value(i++));
```
karasikov, line 202: Is there a usage message for this flag?

metagraph/src/graph/representation/rc_dbg.hpp
```diff
@@ -0,0 +1,129 @@
...
+
+#include "graph/representation/base/sequence_graph.hpp"
+#include "common/seq_tools/reverse_complement.hpp"
+
+namespace mtg {
+namespace graph {
+
+class RCDBG : public DeBruijnGraph {
+  public:
+    RCDBG(const DeBruijnGraph &graph) : graph_(graph) {}
```
karasikov, line 12: add a one or two-line comment description?

metagraph/src/cli/load/load_annotated_graph.cpp
```diff
@@ -23,15 +23,13 @@ using mtg::common::logger;
 
 std::unique_ptr<AnnotatedDBG> initialize_annotated_dbg(std::shared_ptr<DeBruijnGraph> graph,
                                                        const Config &config) {
+    uint64_t max_index = graph->get_base_graph().max_index();
+
     if (graph->get_mode() == DeBruijnGraph::PRIMARY) {
         graph = std::make_shared<CanonicalDBG>(graph);
         logger->trace("Primary graph was wrapped into canonical");
     }
```
karasikov, line 31: There too many degrees of freedom. Better make the logic as simple as possible while still supporting all features. Choose whether one should pass an unwrapped graph (and hence, no `get_base_graph()` is needed to be called) or never a primary graph (so it must always be wrapped when passed, and no wrapping is needed).

The code will be simpler and easier to maintain/use.

karasikov (changes requested):

metagraph/src/cli/load/load_annotated_graph.cpp
hmusta: ok, makes sense. I added an assert here to make sure only a base graph is passed

metagraph/src/cli/config/config.cpp
hmusta: added

metagraph/src/graph/representation/rc_dbg.hpp
hmusta: done

metagraph/src/cli/config/config.cpp
```diff
@@ -949,21 +944,21 @@ void Config::print_usage(const std::string &prog_name, IdentityType identity) {
...
+            fprintf(stderr, "\t   --align-one-strand \t\t\t\tdo not align backwards from a seed on basic-mode graphs [off]\n");
             fprintf(stderr, "\t   --align-alternative-alignments \t\tthe number of alternative paths to report per seed [1]\n");
             fprintf(stderr, "\t   --align-min-path-score [INT]\t\t\tthe minimum score that a reported path can have [0]\n");
-            fprintf(stderr, "\t   --align-edit-distance \t\t\tuse unit costs for scoring matrix [off]\n");
             fprintf(stderr, "\t   --align-max-nodes-per-seq-char [FLOAT]\tmaximum number of nodes to consider per sequence character [12.0]\n");
             fprintf(stderr, "\t   --align-max-ram [FLOAT]\t\t\tmaximum amount of RAM used per alignment in MB [200.0]\n");
+            fprintf(stderr, "\t   --align-xdrop [INT]\t\t\t\tthe maximum difference between the current and the best alignment [27]\n");
+            fprintf(stderr, "\t   --align-fraction-of-top [FLOAT]\t\tduring extension, use this fraction of the current best alignment as a lower bound on the alignment score [0.8]\n");
```
karasikov, line 953:
```suggestion
            fprintf(stderr, "\t   --align-fraction-of-top [FLOAT]\t\tmin score relative to the current best alignment for keeping paths during extension [0.8]\n");
```

karasikov: "fraction" is confusing, we use it to refer to subset sizes in other cases. Maybe `--min-relative-score` ?

metagraph/src/cli/config/config.hpp
```diff
@@ -120,6 +119,8 @@ class Config {
     size_t alignment_max_seed_length = std::numeric_limits<size_t>::max();
     size_t alignment_max_num_seeds_per_locus = std::numeric_limits<size_t>::max();
 
+    double alignment_fraction_of_top = 0.8;
```
karasikov, line 122: Rename, when looking at this name, I'd think this is about keeping 80% of the best paths at each step, which is not true. So I'd be confused.

metagraph/src/common/utils/simd_utils.hpp
```diff
@@ -5,6 +5,10 @@
 #include <immintrin.h>
 #endif
 
+#ifdef __SSE4_1__
+#include <smmintrin.h>
+#endif
+
 #ifdef __SSE2__
```
karasikov, line 12: should it be `#elif`?

karasikov (changes requested):

metagraph/src/cli/config/config.cpp
karasikov: or `--rel-score-cutoff`

metagraph/src/graph/alignment/aligner_alignment.hpp
```diff
@@ -249,20 +228,30 @@ class QueryAlignment {
...
+
+    const Alignment& operator[](size_t i) const { return alignments_[i]; }
+    iterator begin() { return alignments_.begin(); }
+    iterator end() { return alignments_.end(); }
+    const_iterator begin() const { return alignments_.begin(); }
+    const_iterator end() const { return alignments_.end(); }
+
+    iterator erase(const_iterator begin, const_iterator end) {
+        return alignments_.erase(begin, end);
+    }
```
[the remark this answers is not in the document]
hmusta: That's probably a better idea. The number of features supported by this container kept growing over time...

metagraph/src/common/utils/simd_utils.hpp
hmusta: Not sure if it matters too much since these headers have include guards.

metagraph/src/cli/config/config.cpp
hmusta: for the help message, how about
```min score relative to the current best alignment to use as a lower bound for subsequent extensions```

File: metagraph/src/common/utils/simd_utils.hpp (hunk 6 above)
Comment (hmusta) [reply metagraph/src/common/utils/simd_utils.hpp]: apparently including `x86intrin.h` is enough for all compilers, except MSVC, so I just used that

File: metagraph/src/graph/alignment/aligner_alignment.cpp [hunk 8]
```diff
@@ -43,8 +43,18 @@ Alignment<NodeType>::Alignment(std::string_view query,
...
 
-template <typename NodeType>
-void Alignment<NodeType>::append(Alignment&& other) {
+std::ostream& operator<<(std::ostream& out, const Alignment &alignment) {
+    out << (alignment.get_orientation() ? "-" : "+") << "\t"
+        << alignment.get_sequence() << "\t"
+        << alignment.get_score() << "\t"
+        << alignment.get_num_matches() << "\t"
+        << alignment.get_cigar().to_string() << "\t"
+        << alignment.get_offset();
```
karasikov, line 52: out << fmt::format(...);

metagraph/src/graph/alignment/aligner_extender_methods.cpp
```diff
@@ -50,709 +55,695 @@ DefaultColumnExtender<NodeType>::DefaultColumnExtender(const DeBruijnGraph &grap
...
+template <class Column, typename... RestArgs>
+Column alloc_column(size_t size, RestArgs... args) {
+    Column column { {}, {}, {}, args... };
+    auto &[S, E, F, node, i_prev, c, offset, max_pos, trim] = column;
 
-    if (max_score < min_path_score)
-        return;
+    S.reserve(size + kPadding);
+    E.reserve(size + kPadding);
+    F.reserve(size + kPadding);
```
karasikov, line 353: Are you sure a multiplication factor wouldn't be better?
This additive will work well only if you indeed extend the vectors by a constant string not dependent on the initial size.

i.e., `2 * size`

metagraph/src/graph/alignment/aligner_extender_methods.cpp
```diff
@@ -50,709 +55,695 @@ DefaultColumnExtender<NodeType>::DefaultColumnExtender(const DeBruijnGraph &grap
...
+                               size_t /* table idx */>;
+    TableIt best_score { 0, 0, 0 };
 
-        for (const auto &[next, c] : get_outgoing(prev)) {
-            auto &column_pair = table_[next];
-            auto &[column, converged] = column_pair;
-            if (converged)
-                continue;
+    // node traversal heap
+    std::priority_queue<TableIt> queue;
```
karasikov, line 395: do you actually ever call `std::get<1>(...)` on these tuples?

karasikov (changes requested):

metagraph/src/cli/config/config.cpp
karasikov: either is fine

metagraph/src/graph/alignment/aligner_extender_methods.cpp
hmusta: not really, they're just there for breaking ties

karasikov: Ok, please write this in a comment.

metagraph/src/graph/alignment/aligner_extender_methods.cpp
hmusta: the SIMD code reads in 128-bit blocks, so I'm doing this padding to ensure that it doesn't read out of bounds.

To make things more compact, I could just as well just round the size up to the nearest multiple of 128 bits.

karasikov: why not resize then?

karasikov: also, write this in a comment

hmusta: I explained this in a comment, it's to allow for AlignedVector methods (like size() and push_back()) to still function properly.

karasikov (changes requested):
converging...

metagraph/src/graph/alignment/aligner_extender_methods.hpp
```diff
@@ -1,125 +1,171 @@
...
+                                 const std::function<void(Alignment&&)> &callback) {
+        assert(path.size());
+        assert(ops.size());
+
+        if (cur_cell_score == 0 && ops.back().first == Cigar::MATCH
+                && end_score >= min_path_score) {
+            callback(construct_alignment(ops, clipping, window, path, match,
+                                         end_score, offset));
+        }
+    }
```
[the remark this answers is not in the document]
hmusta: the default behaviour is to call one alignment, but I wanted to leave the option open to generate multiple alignments from a single starting point, such as in #336

metagraph/src/cli/align.cpp
```diff
@@ -416,26 +373,26 @@ int align_to_graph(Config *config) {
...
+        bool is_reverse_complement = false;
 
         Timer data_reading_timer;
 
-        std::ostream *out = config->outfbase.size()
-            ? new std::ofstream(config->outfbase)
-            : &std::cout;
+        std::shared_ptr<std::ostream> out{ std::shared_ptr<std::ostream>{}, &std::cout };
+        if (config->outfbase.size())
+            out = std::make_shared<std::ofstream>(config->outfbase);
```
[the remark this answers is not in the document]
hmusta: hmm....I never explicitly make copies of this, but I guess it would be safer to use unique_ptr instead since I never modify this.

or, I could just make it only output to stdout and forget about writing to a file

metagraph/src/graph/alignment/aligner_config.hpp
```diff
@@ -52,22 +52,24 @@ class DBGAlignerConfig {
...
     size_t num_alternative_paths = 1;
-    size_t min_seed_length = 1;
-    size_t max_seed_length = std::numeric_limits<size_t>::max();
+    size_t min_seed_length = 0;
+    size_t max_seed_length = 0;
     size_t max_num_seeds_per_locus = std::numeric_limits<size_t>::max();
+
     // thresholds for scores
-    score_t min_cell_score = 0;
+    score_t min_cell_score = std::numeric_limits<score_t>::min() + 100;
```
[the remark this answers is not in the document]
hmusta: I'll leave a comment. It's there to prevent underflow

metagraph/src/cli/align.cpp
```diff
@@ -81,62 +74,29 @@ DBGAlignerConfig initialize_aligner_config(size_t k, const Config &config) {
...
                            ThreadPool *thread_pool = nullptr,
                            std::mutex *print_mutex = nullptr) {
     // TODO: multithreaded
     std::ignore = std::tie(thread_pool, print_mutex);
 
+    const DBGSuccinct *dbg = dynamic_cast<const DBGSuccinct*>(&graph);
+    if (!dbg) {
+        if (const auto *canonical = dynamic_cast<const CanonicalDBG*>(&graph))
+            dbg = dynamic_cast<const DBGSuccinct*>(&canonical->get_graph());
+    }
```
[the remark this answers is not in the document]
hmusta: `dbg` is only used for sub-k matching. I'll clarify this in a comment

metagraph/src/cli/align.cpp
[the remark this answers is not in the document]
hmusta: I just replaced it with a unique_ptr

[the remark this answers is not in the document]
karasikov: This will work, unique_ptr doesn't allow copying, where potential race conditions could happen.

metagraph/src/graph/alignment/aligner_extender_methods.hpp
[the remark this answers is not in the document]
karasikov: then write a comment for this method. Something like
```
// In fact, this function calls at most a single alignment
```

metagraph/src/cli/align.cpp
```diff
@@ -91,11 +91,8 @@ void map_sequences_in_file(const std::string &file,
     // TODO: multithreaded
     std::ignore = std::tie(thread_pool, print_mutex);
 
-    const DBGSuccinct *dbg = dynamic_cast<const DBGSuccinct*>(&graph);
-    if (!dbg) {
-        if (const auto *canonical = dynamic_cast<const CanonicalDBG*>(&graph))
-            dbg = dynamic_cast<const DBGSuccinct*>(&canonical->get_graph());
-    }
+    // used for sub-k sequence matching
+    const DBGSuccinct *dbg = dynamic_cast<const DBGSuccinct*>(&graph.get_base_graph());
```
karasikov, line 95: pass DBGSuccinct to the function. This conversion should happen in the caller. And in many cases it will be avoided there.

Otherwise you potentially enforce unnecessary conversions `DBGSuccinct -> DeBruijnGraph -> DBGSuccinct`

karasikov (changes requested):

metagraph/src/graph/alignment/aligner_aggregator.hpp
```diff
@@ -23,54 +34,69 @@ class AlignmentAggregator {
...
+template <class AlignmentCompare>
+inline void AlignmentAggregator<AlignmentCompare>
+::call_alignments(const std::function<void(Alignment&&)> &callback,
+                  const std::function<bool()> &terminate) {
+    auto &data = path_queue_.data();
+    for (auto it = data.rbegin(); it != data.rend() && !terminate(); ++it) {
+        boost::heap::pop_interval_heap_max(data.begin(), it.base(), path_queue_.cmp());
+        callback(std::move(*it));
     }
 }
```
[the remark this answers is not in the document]
hmusta: Using iterators here would be strange since each retrieval pops an element off the list.

metagraph/src/graph/representation/rc_dbg.hpp
```diff
@@ -0,0 +1,135 @@
...
+    virtual size_t get_k() const override final { return graph_.get_k(); }
+    virtual Mode get_mode() const override final { return graph_.get_mode(); }
+
+    virtual node_index traverse(node_index node, char next_char) const override final {
+        return graph_.traverse_back(node, complement(next_char));
+    }
+
+    virtual node_index traverse_back(node_index node, char prev_char) const override final {
+        return graph_.traverse(node, complement(prev_char));
+    }
```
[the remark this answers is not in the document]
hmusta: good idea! I've implemented this here

metagraph/src/graph/alignment/aligner_aggregator.hpp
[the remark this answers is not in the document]
karasikov: Then the name `call_alignments` is confusing, it doesn't make it clear that it changes the state and can be called only once.

btw, why don't you return them in a vector?

metagraph/src/graph/representation/base/dbg_wrapper.hpp
```diff
@@ -148,6 +154,9 @@ class DBGWrapper : public DeBruijnGraph {
   protected:
     std::shared_ptr<const Graph> graph_;
     std::shared_ptr<Graph> graph_ptr_;
+
+    // clear any internal storage the wrapper may have
+    virtual void flush() {}
```
karasikov, line 159: This is too much. Better overload method `load` in the children.

karasikov (changes requested):

metagraph/src/graph/representation/base/dbg_wrapper.hpp
hmusta: this method is called by `add_sequence` and `load`, so I designed it this way so that the children would not have to overload them

karasikov: I see, but this looks bad. You make it less more confusing by adding any methods which aren't in `DeBruijnGraph`.

The abstract wrapper should not have any new methods.

karasikov: You are essentially filling holes with hacks like this.

But you will never fill all possible holes. So if you stick to the same strategy, this class will grow and grow in the future, which is obviously not good.

metagraph/src/graph/alignment/aligner_aggregator.hpp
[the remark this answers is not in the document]
hmusta: It could be that the termination condition is dependent on the alignments that are returned (e.g., terminate after returning three alignments, or terminate when the score is below a value X, etc.).

metagraph/src/graph/representation/base/dbg_wrapper.hpp
hmusta: I figured that this is something that would be needed by any wrapper which stores auxiliary information, so I don't really see it as a hack.

One pitfall, though, is that I think it should be a pure abstract function in `DBGWrapper` since it needs to be overridden in most cases anyway. If it's like this, then it'll be easy to forget to do this.

karasikov: Hm, I think wrappers with aux stuff should override loaders/etc

metagraph/src/graph/alignment/aligner_aggregator.hpp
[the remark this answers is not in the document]
karasikov: but they're already computed, right? Why not always return all of them?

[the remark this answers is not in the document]
hmusta: The alignments are computed, but they should be returned in sorted (non-increasing score) order. So, if the queue is large, but you're only interested in, say, the top two alignments, then you only end up performing 2log(n) operations instead of nlog(n) to retrieve them.

metagraph/src/graph/representation/base/dbg_wrapper.hpp
hmusta: anyway, I don't think it's worth arguing over this too much...

I removed `flush`. Instead, I've made it so that `add_sequence` and `load` are pure virtual in `DBGWrapper` to remind the programmer to update their internal storage whenever the underlying graph is modified.

hmusta:
This has started to grow a lot.... I'm gonna break off DBGWrapper into its own pull request

metagraph/src/graph/alignment/aligner_aggregator.hpp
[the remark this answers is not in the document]
hmusta: `call_alignments` is not used here, so I just replaced it with `get_alignments`. We can add `call_alignments` later if needed.

metagraph/src/graph/alignment/aligner_alignment.hpp
```diff
@@ -200,45 +152,47 @@ struct LocalAlignmentGreater {
...
-    const_iterator cend() const { return alignments_.cend(); }
+
+    std::vector<Alignment>& data() { return alignments_; }
+    const std::vector<Alignment>& data() const { return alignments_; }
 
   private:
-    std::shared_ptr<std::string> query_;
-    std::shared_ptr<std::string> query_rc_;
+    std::shared_ptr<const std::string> query_;
+    std::shared_ptr<const std::string> query_rc_;
```
karasikov, line 195: I would avoid using shared pointers in such cases

karasikov: It's usually better to move things.

metagraph/src/graph/alignment/aligner_config.hpp
```diff
@@ -52,22 +52,24 @@ class DBGAlignerConfig {
...
     size_t num_alternative_paths = 1;
-    size_t min_seed_length = 1;
-    size_t max_seed_length = std::numeric_limits<size_t>::max();
+    size_t min_seed_length = 0;
+    size_t max_seed_length = 0;
     size_t max_num_seeds_per_locus = std::numeric_limits<size_t>::max();
+
     // thresholds for scores
-    score_t min_cell_score = 0;
+    score_t min_cell_score = std::numeric_limits<score_t>::min() + 100; // to prevent underflow
```
karasikov, line 60: Explain why 100 is enough?

metagraph/src/graph/alignment/aligner_extender_methods.cpp
```diff
@@ -50,696 +55,699 @@ DefaultColumnExtender::DefaultColumnExtender(const DeBruijnGraph &graph,
...
-        PF.push_back(Extender::NONE);
-        if (S.back() > 0) {
-            assert(S.back() == E.back());
-            updated = true;
-            PS.back() = Extender::CUR;
-            OS.back() = Cigar::INSERTION;
-            xdrop_cutoff = std::max(xdrop_cutoff, S.back() - config_.xdrop);
-        }
-    }
+    } else {
```
karasikov, line 151: no need for "else" after "return"

metagraph/src/graph/alignment/aligner_extender_methods.cpp
```diff
@@ -50,696 +55,699 @@ DefaultColumnExtender::DefaultColumnExtender(const DeBruijnGraph &graph,
...
+                            queue.emplace(next_score);
+                        }
+                    }
 
-                column.emplace_back(std::move(next_column));
-            } else if (!depth) {
-                table_.erase(next);
+                } else {
+                    table.pop_back();
+                }
```
karasikov, line 573: better
```c++
if (max_val < xdrop_cutoff) {
    table.pop_back();
    continue;
}

...

File: metagraph/src/graph/alignment/aligner_seeder_methods.cpp [hunk 23]
```diff
@@ -388,6 +405,17 @@ auto MEMSeeder::get_seeds() const -> std::vector<Seed> {
     return seeds;
 }
 
+template <class BaseSeeder>
+const DBGSuccinct& SuffixSeeder<BaseSeeder>
+::get_base_dbg_succ(const DeBruijnGraph &graph) {
+    try {
+      return dynamic_cast<const DBGSuccinct&>(graph.get_base_graph());
```
karasikov, line 412:
```suggestion
        return dynamic_cast<const DBGSuccinct&>(graph.get_base_graph());
```

metagraph/src/graph/representation/rc_dbg.hpp
```diff
@@ -0,0 +1,142 @@
...
+ * e.g., get_node_sequence(n) := reverse_complement(graph.get_node_sequence(n))
+ * e.g., traverse(n, c) := graph.traverse_back(n, complement(c))
+ */
+class RCDBG : public DBGWrapper<DeBruijnGraph> {
+  public:
+    template <typename... Args>
+    RCDBG(Args&&... args) : DBGWrapper<DeBruijnGraph>(std::forward<Args>(args)...) {}
+
+    virtual uint64_t num_nodes() const override final { return graph_->num_nodes(); }
+    virtual uint64_t max_index() const override final { return graph_->max_index(); }
```
karasikov, line 21: I think these should also go to `DBGWrapper`.

karasikov (approved):

metagraph/src/graph/alignment/aligner_alignment.hpp
hmusta: This was done out of necessity. Each alignment has a string_view window to its query sequence. If I don't make these shared pointers, then there's no way to guarantee that the views in each Alignment will stay alive and valid.

I've tried this with different combinations of making these `const std::string` and disabling constructors, but compiler optimizations seem to always cause weird stuff...

PR #334, opened 2021-07-12
Cache extracted values for DBGSuccinct

hmusta:
Useful for speeding up alignment to Primary and Basic mode graphs.
- [x] Cache the results of `get_minus_k_value(k - 2)` calls
- [x] Cache extracted k-mers for faster `get_node_sequence` calls
- [x] Cache results from reverse complement functions used in CanonicalDBG
- [x] Unit tests
- [x] Integration tests
- [x] Benchmarks


metagraph/src/graph/representation/succinct/dbg_succinct_cached.hpp
```diff
@@ -0,0 +1,235 @@
...
+    DBGSuccinctCached(const DeBruijnGraph &graph, size_t cache_size = 1000)
+          : DBGSuccinctCached(dynamic_cast<const DBGSuccinct&>(graph), cache_size) {}
+
+    virtual const DeBruijnGraph& get_base_graph() const override final { return graph_.get_base_graph(); }
+
+    const DBGSuccinct& get_dbg_succ() const { return graph_; }
+
+    inline void put_decoded_node(node_index node, const std::string &seq) const {
+        decoded_cache_.Put(node, seq);
+    }
```
karasikov, line 39: I think if we make this class, the cache should be efficient.

1) we should store k-mers compressed (probably as `KMer`?) and not std::string
2) we should use a more efficient cache library (a good hash table, not std::unordered_map which is used there at the moment) and std::deque instead of std::list.

Though we can postpone 2) for later

metagraph/src/graph/representation/succinct/dbg_succinct_cached.hpp
```diff
@@ -0,0 +1,235 @@
...
+    template <typename Key, typename Value>
+    using LRUCache = caches::fixed_sized_cache<Key, Value, caches::LRUCachePolicy<Key>>;
+
+    typedef LRUCache<node_index, std::string> DecodedCache;
+    typedef LRUCache<edge_index, edge_index> BwdFirstCache;
+
+    DBGSuccinctCached(const DBGSuccinct &dbg, size_t cache_size = 1000)
+          : graph_(dbg), boss_(graph_.get_boss()),
+            bwd_first_cache_(cache_size),
+            decoded_cache_(cache_size * dbg.alphabet().size()) {}
```
karasikov, line 28: Would be good to add benchmarks similar to https://github.com/ratschlab/metagraph/blob/master/metagraph/benchmarks/graph/bench_boss.cpp

So then we can work on improving this implementation more easily and get more confident improvements.

karasikov (changes requested):

metagraph/src/graph/representation/succinct/dbg_succinct_cached.hpp
hmusta: Storing k-mers in compressed form improved the overall alignment performance by 15%.

For now, I've added a TODO for the second point. I'll benchmark this thing to see if the cache operations are a significant bottleneck.

karasikov: I suggested this mainly to reduce the memory footprint, so you can make the cache size much larger.

metagraph/src/kmer/kmer_extractor.cpp
```diff
@@ -399,6 +399,44 @@ void KmerExtractorBOSS::sequence_to_kmers(std::string_view,
                                           Vector<Kmer256>*,
                                           bool);
 
+template <typename KMER>
+KMER KmerExtractorBOSS::sequence_to_kmer(std::string_view sequence,
+                                         bool canonical_mode) {
+    // encode sequence
+    std::vector<TAlphabet> seq(sequence.size(), alphabet.size());
```
karasikov, line 406:
```suggestion
    std::vector<TAlphabet> seq(sequence.size());
```

metagraph/src/graph/representation/succinct/dbg_succinct_cached.hpp
hmusta: That's one thing, but an additional benefit of having compressed k-mers is that it allowed for a sequence of adjacent k-mers to be efficiently generated and cached through small copies and bitwise operations, rather than copying and concatenating strings.

metagraph/src/kmer/kmer_extractor.cpp
```diff
@@ -399,6 +399,44 @@ void KmerExtractorBOSS::sequence_to_kmers(std::string_view,
...
+
+    std::transform(sequence.begin(), sequence.end(), seq.begin(), [](char c) {
+        return c != alphabet[0] ? encode(c) : 0;
+    });
+
+    if (std::find(seq.begin(), seq.end(), alphabet.size()) != seq.end())
+        throw std::runtime_error("Invalid characters encoded");
```
karasikov, line 413: I don't get it. It will never throw, you map '$' to 0 and then check for `alphabet.size()`

metagraph/src/graph/representation/succinct/dbg_succinct_cached.hpp
```diff
@@ -0,0 +1,362 @@
...
+#include "kmer/kmer_extractor.hpp"
+
+namespace mtg {
+namespace graph {
+
+
+class DBGSuccinctCached : public DeBruijnGraph {
+  public:
+    typedef boss::BOSS::edge_index edge_index;
+    typedef boss::BOSS::TAlphabet TAlphabet;
```
karasikov, line 18: Why don't we make it a template class?

Or, does this class use something specific to DBGSuccinct, i.e., sentinel characters?

karasikov (changes requested):

metagraph/src/kmer/kmer_extractor.cpp
hmusta: oops, I forgot to remove this check...

metagraph/src/graph/representation/succinct/dbg_succinct_cached.hpp
hmusta: I had a description for this in the pull request, but I'll also add a comment here.

This class caches the results for two things: 1) `get_node_sequence()`, and 2) `get_minus_k_value(k - 2)`

So, it has to be specific to DBGSuccinct. One could split this up such that a general class handles 1) and a more specific class handles 2), but I don't think it's worth implementing at this time.

metagraph/src/kmer/kmer_extractor.cpp
hmusta: If the input sequence has non-ASCII characters, this can still throw. This should never happen in my usage of this function, but I kept the throw here just in case.

metagraph/src/cli/align.cpp
```diff
@@ -433,10 +434,20 @@ int align_to_graph(Config *config) {
...
 
-            auto process_batch = [&](SeqBatch batch) {
-                auto aln_graph = graph;
-                if (auto *canonical = dynamic_cast<CanonicalDBG*>(graph.get()))
-                    aln_graph = std::make_shared<CanonicalDBG>(*canonical);
+            auto process_batch = [&,base_graph](SeqBatch batch) {
+                // aliasing constructor
+                auto aln_graph = base_graph;
+                auto *dbg_succ = dynamic_cast<DBGSuccinct*>(aln_graph.get());
+                bool is_primary = aln_graph->get_mode() == DeBruijnGraph::PRIMARY;
```
karasikov, line 441: why don't you always wrap primary into a canonical?

hmusta: `DBGSuccinctCached` needs to wrap on top of `DBGSuccinct` first, then `CanonicalDBG` on top of that.

metagraph/src/graph/representation/succinct/dbg_succinct_cached.hpp
karasikov: Why do you need to cache `get_minus_k_value(k - 2)`? Are you sure you don't cache the same nodes in both, and hence reconstruct the same nodes twice?

metagraph/src/cli/align.cpp
karasikov: ok, what is going on in this function?

You take the base graph, check if it's primary ... ???

Eventually, you want to create another view on the graph with its own cache, right?

Move this to a separate function.

karasikov (changes requested):

metagraph/src/graph/representation/succinct/dbg_succinct_cached.hpp
hmusta: `get_minus_k_value(k - 2)` returns two things: the decoded first character, and the corresponding BOSS node from the backwards traversal. So, the cache maps the input DBGSuccinct node to that BOSS node. It also caches the `bwd` of both of these, so the next time the function is called, it can fetch it from the cache.

The whole point of this is to speed up backwards traversal, so each step (on average) only needs to pay the cost of two `bwd` calls.

karasikov: Wouldn't caching `get_node_sequence()` be enough?

hmusta: Almost. If we modified `BOSS::get_node_seq` (or added a new helper function) to also return the last node after traversal, then we could have a single cache which keeps a pair of an encoded k-mer and that BOSS node.

karasikov: why just storing `get_node_sequence()` would not be enough with the current interface?

hmusta: Because that's only good enough for retrieving the value sequence of a node you've seen before. For any of its descendants, you'd need to recompute the sequence from scratch. If we have access to the last node that was accessed during the decoding, we could compute the full sequence of its descendent with a single `bwd` call (plus another `bwd` call to compute the corresponding key for the cache).

metagraph/src/cli/align.cpp
hmusta: done. I've also replaced the outdated comment to explain things a bit better

metagraph/src/cli/align.cpp
```diff
@@ -350,10 +342,7 @@ int align_to_graph(Config *config) {
     ThreadPool thread_pool(get_num_threads());
     std::mutex print_mutex;
 
-    if (graph->get_mode() == DeBruijnGraph::PRIMARY) {
-        logger->trace("Primary graph wrapped into canonical");
-        graph = std::make_shared<CanonicalDBG>(graph);
-    }
+    graph = wrap_graph(graph);
```
karasikov, line 345: bad name

karasikov (changes requested):

metagraph/src/cli/align.cpp
hmusta: How about something like `sanitize_graph`? Any other suggestions?

karasikov: Neither of them is descriptive.

It might be hard to come up with a good name because it has two things.

Just do the canonicalization outside and pass a graph for adding a cache.

Then it can be called `add_cache`. It doesn't matter if this has to re-apply the CanonicalDBG wrapper inside.

metagraph/src/graph/representation/succinct/dbg_succinct_cached.hpp
hmusta: I updated the implementation so it uses a single cache to store a pair with both the constructed node label, and the node resulting from the backwards traversals.

metagraph/src/cli/align.cpp
hmusta: ok, I split the two functions

metagraph/src/graph/representation/succinct/dbg_succinct_cached.hpp
```diff
@@ -0,0 +1,372 @@
...
+namespace mtg {
+namespace graph {
+
+
+// A wrapper which caches computed node sequences for DBGSuccinct graphs.
+// This allows for faster get_node_sequence and call_incoming_kmers calls.
+class DBGSuccinctCached;
+
+std::shared_ptr<DBGSuccinctCached>
+make_cached_dbgsuccinct(std::shared_ptr<const DBGSuccinct> graph, size_t cache_size = 1024);
```
karasikov, line 19: Better make it a public method of `DBGSuccinct`: `get_cached_view(size_t cache_size)`.

metagraph/src/graph/representation/succinct/dbg_succinct_cached.hpp
```diff
@@ -0,0 +1,372 @@
...
+
+class DBGSuccinctCached : public DBGWrapper<DBGSuccinct> {
```
karasikov, line 21: Could you write a list of methods that are re-defined (benefit from caching)?

metagraph/tests/graph/all/test_dbg_helpers.hpp
```diff
@@ -93,7 +100,25 @@ typedef ::testing::Types<DBGBitmap,
...
+                         DBGSuccinct,
+                         DBGSuccinctIndexed<1>,
+                         DBGSuccinctIndexed<2>,
+#if ! _PROTEIN_GRAPH
+                         DBGSuccinctIndexed<10>,
+#endif
+                         DBGSuccinctBloomFPR<1, 1>,
+                         DBGSuccinctBloomFPR<1, 10>,
+                         DBGSuccinctBloom<4, 1>,
+                         DBGSuccinctBloom<4, 50>> MutableGraphTypes;
```
karasikov, line 121: Why are `DBGSuccinctBloom` and `DBGSuccinctIndexed` are mutable, but not `DBGSuccinctCachedWrapped`?

metagraph/src/kmer/kmer_extractor.cpp
karasikov: You reinitialize everything in the next line, there is no point to set the initial value.

metagraph/src/graph/representation/succinct/boss.cpp
```diff
@@ -849,9 +849,12 @@ bool BOSS::compare_node_suffix(edge_index first, const TAlphabet *second) const
...
 /**
  * Given an edge index i, this function returns the k-mer sequence of its
- * source node.
+ * source node, and the node whose last character corresponds to the first
+ * character of the sequence. If the graph is suffix indexed, then the returned
+ * node is the last node visited after k - indexed_suffix_length_ bwd steps.
  */
-std::vector<TAlphabet> BOSS::get_node_seq(edge_index x) const {
+std::tuple<std::vector<TAlphabet>, edge_index, bool> BOSS
+::get_node_seq_with_end_node_indexed(edge_index x) const {
```
karasikov, line 857: The bool is essentially used to indicate that the returned edge_index is unusable.

So, just return npos for edge_index in that case, and you don't need to add this bool value.
```suggestion
std::tuple<std::vector<TAlphabet>, edge_index> BOSS
::get_node_seq_with_end_node_indexed(edge_index x) const {
```

metagraph/src/graph/representation/succinct/boss.cpp
```diff
@@ -893,7 +896,53 @@ std::vector<TAlphabet> BOSS::get_node_seq(edge_index x) const {
...
+
+        auto it = seq.rend() - indexed_suffix_length_;
+        assert(*it == get_node_last_value(last_node));
+        for (++it; it != seq.rend(); ++it) {
+            last_node = bwd(last_node);
+            assert(*it == get_node_last_value(last_node));
+        }
+
+        last_node_opt = last_node;
+    }
```
karasikov, line 940: there are too many optional and stuff, which complicates the whole design of these interfaces.

To simplify this, start with removing this optional argument -- just have two functions.

metagraph/src/graph/representation/succinct/boss.cpp
karasikov: Never mind, this whole thing should be changed.

See another comment for `get_node_seq_with_end_node`

karasikov (changes requested):

metagraph/src/graph/representation/succinct/boss.cpp
karasikov: So, get back the old method and add the one you need as a new separate method.

metagraph/tests/graph/all/test_dbg_helpers.hpp
hmusta: serialize and load are not implemented in `DBGSuccinctCachedWrapped`

karasikov: Why not? Just reset the cache and call the base functions.

hmusta: I had it at some point and you asked me to remove it, but I don't remember if this was for this class or the parent wrapper class

hmusta: it'll make things more complicated because the wrapper class would have to store a non-const shared_ptr

karasikov: Ok, this sounds like a reason.
For serialize/load nothing prevents us to implement those.

metagraph/src/graph/representation/succinct/dbg_succinct_cached.hpp
karasikov: or `get_const_cached_view()`

metagraph/src/graph/representation/succinct/dbg_succinct_cached.hpp
karasikov: So View would indicate it's a const thing.

metagraph/src/graph/representation/succinct/dbg_succinct_cached.hpp
hmusta: Sure, I can add `get_cached_view`, but I would still keep `make_cached_dbgsuccinct`. `get_cached_view` would have to be passed an alias of `DBGSuccinct`, so one would have to do extra work to keep that `DBGSuccinct` in scope once a cached graph is generated. This'll make it more annoying to implement `build_graph` from the unit tests.

metagraph/src/graph/representation/succinct/dbg_succinct.hpp
```diff
@@ -183,6 +185,8 @@ class DBGSuccinct : public DeBruijnGraph {
 
     const mtg::kmer::KmerBloomFilter<>* get_bloom_filter() const { return bloom_filter_.get(); }
 
+    std::shared_ptr<DBGSuccinctCached> get_cached_view(size_t cache_size = 1024) const;
```
karasikov, line 188: Why so small default cache?

Add a comment how many bytes are taken by one element in the cache.

karasikov (changes requested):

metagraph/src/cli/align.cpp
```diff
@@ -349,10 +348,8 @@ int align_to_graph(Config *config) {
     ThreadPool thread_pool(get_num_threads());
     std::mutex print_mutex;
 
-    if (graph->get_mode() == DeBruijnGraph::PRIMARY) {
-        logger->trace("Primary graph wrapped into canonical");
-        graph = std::make_shared<CanonicalDBG>(graph);
-    }
+    if (graph->get_mode() == DeBruijnGraph::PRIMARY)
+        graph = primary_to_canonical(graph);
```
karasikov, line 352: this doesn't make it clear that it will also have a cached view on the graph

karasikov (changes requested):

metagraph/src/graph/representation/succinct/dbg_succinct_cached.hpp
karasikov: I think this should not be necessary. If this is only needed for unit tests, move it to there. Then you'll probably see how you can get rid of it completely.

metagraph/src/graph/representation/succinct/dbg_succinct.hpp
```diff
@@ -10,6 +10,8 @@
 namespace mtg {
 namespace graph {
 
+class DBGSuccinctCached;
```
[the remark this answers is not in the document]
hmusta: This required a lot of reorganization to move stuff from `dbg_succinct_cached.hpp` to `dbg_succinct_cached.cpp`, but hopefully it works now.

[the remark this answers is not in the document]
hmusta: Including the hpp is not enough since this creates a circular dependence (since the main header and the cached header include each other)

metagraph/src/cli/align.cpp
hmusta: At this point, DBGSuccinctCached is not used, it's only introduced later when a batch is being aligned.

metagraph/src/graph/representation/succinct/boss.cpp
hmusta: I still want to keep `get_node_seq_with_end_node_indexed` to avoid having two functions (this and `get_node_seq`) that do nearly the same work. Later down the road, someone will make some changes to `get_node_seq` and it'll be their responsibility to make sure that the same changes get propagated to the other function (which most likely won't happen).

metagraph/src/graph/representation/succinct/dbg_succinct.hpp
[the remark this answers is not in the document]
karasikov: that's ok

metagraph/src/graph/representation/succinct/boss.cpp
karasikov: You also have code duplications right now. E.g., this block:
```
        auto it = seq.rend() - indexed_suffix_length_;
        assert(*it == get_node_last_value(last_node));
        for (++it; it != seq.rend(); ++it) {
            last_node = bwd(last_node);
            assert(*it == get_node_last_value(last_node));
        }
```
But on top you also have a bunch of confusing arguments that are most of the time aren't used.

karasikov: also lots of confusing if - else and stuff. So it's really better to get rid of all those optional arguments like `bool force_get_end_node`

metagraph/src/cli/align.cpp
karasikov: Then how do you explain this signature?
```c++
std::shared_ptr<graph::DeBruijnGraph> primary_to_canonical(std::shared_ptr<Graph> graph,
                                                           size_t cache_size = 1024);
```
I see a non-zero cache size.
Or, is this another argument that is dummy and not used?

karasikov (changes requested):

metagraph/src/cli/align.cpp
hmusta: `CanonicalDBG` and `DBGSuccinctCached` cache different things.

`CanonicalDBG` caches the child and parent nodes of each node (and works for every graph type), whereas `DBGSuccinctCache` caches decoded nodes labels and the BOSS edge of the first nucleotide (when backwards traversal is performed), and the boss edges of the reverse complement (when the reverse complement of a node is computed).

karasikov: Ok, I guess this needs to be clarified. Start with changing the returned type to something more specific instead of the most general `DeBruijnGraph`.

karasikov: `std::shared_ptr<CanonicalDBG<Graph>>`?

metagraph/src/graph/representation/succinct/boss.cpp
hmusta: I simplified this quite a bit. Now, there's a method called `get_node_seq_with_last_traversed_node` which returns the sequence, the last traversed node, and the number of steps still needed to reach k bwd steps. This way, if any downstream method can then call `get_minus_k_value(last_traversed_node, steps_remaining)` to get what it wants.

metagraph/src/graph/representation/succinct/dbg_succinct.hpp
hmusta: When I was testing this, it wasn't getting more efficient after 1024 elements, but I think that's just because I was testing with a small, non-redundant short read set

metagraph/benchmarks/graph/bench_boss.cpp
```diff
@@ -14,17 +15,47 @@ using mtg::graph::boss::BOSS;
...
+            );
+        } else if constexpr(std::is_same_v<Graph, DBGSuccinctCachedView>) {
+            std::shared_ptr<DeBruijnGraph> wrapped_graph
+                = make_cached_dbgsuccinct(graph, CACHE_SIZE);
+            if (graph->get_mode() == DeBruijnGraph::PRIMARY)
+                wrapped_graph = std::make_shared<CanonicalDBG>(wrapped_graph);
+
+            return wrapped_graph;
+        }
+    }
```
karasikov, line 58: Oh, this is really weird construct. Keep the old function. It was simple and interpretable.

metagraph/benchmarks/graph/bench_boss.cpp
```diff
@@ -14,17 +15,47 @@ using mtg::graph::boss::BOSS;
 const std::string filename = "./benchmark_graph.dbg";
 
 constexpr uint64_t NUM_DISTINCT_INDEXES = 1 << 21;
+constexpr uint64_t PATH_SIZE = 1 << 7;
+constexpr uint64_t CACHE_SIZE = 1 << 10;
```
karasikov, line 19: Just write the number, why do we need these powers if it's just 1024 or something smaller?

metagraph/src/cli/load/load_graph.cpp
```diff
@@ -71,5 +71,47 @@ std::shared_ptr<DeBruijnGraph> load_critical_dbg(const std::string &filename) {
...
+    return std::make_shared<CanonicalDBG>(graph, cache_size);
+}
+
+template std::shared_ptr<CanonicalDBG> primary_to_canonical(std::shared_ptr<const DeBruijnGraph>, size_t);
+template std::shared_ptr<CanonicalDBG> primary_to_canonical(std::shared_ptr<DeBruijnGraph>, size_t);
+
+std::shared_ptr<const DeBruijnGraph> make_cached_graph(const DeBruijnGraph &graph,
+                                                       const Config &config,
+                                                       size_t cache_size) {
+    std::shared_ptr<const DeBruijnGraph> ret_graph(std::shared_ptr<const DeBruijnGraph>{}, &graph);
```
karasikov, line 92: Currently, you can map wrap first to canonical with cache and then into a cached view, or the other way around.
Isn't one of these more optimal than the other? Did you think about this?
I think we should allow only a certain order of wrapping, to ensure things are wrapped optimally.

metagraph/src/cli/load/load_graph.hpp
```diff
@@ -25,6 +26,15 @@ std::shared_ptr<Graph> load_critical_graph_from_file(const std::string &filename
...
 std::shared_ptr<graph::DeBruijnGraph> load_critical_dbg(const std::string &filename);
 
+template <class Graph>
+std::shared_ptr<graph::CanonicalDBG> primary_to_canonical(std::shared_ptr<Graph> graph,
+                                                          size_t cache_size = 1024);
+
+std::shared_ptr<const graph::DeBruijnGraph>
+make_cached_graph(const graph::DeBruijnGraph &graph,
+                  const Config &config,
+                  size_t cache_size = 1024);
```
karasikov, line 36: So do we ever need larger cache?

metagraph/src/common/caches.hpp
```diff
@@ -0,0 +1,80 @@
...
+#include <utility>
+
+#include <tsl/hopscotch_map.h>
+
+namespace mtg {
+namespace common {
+
+// A fast LRU cache
+template <typename Key, typename Value>
+class LRUCache {
```
karasikov, line 17: Is this faster than the stuff in our external library?

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -90,8 +95,10 @@ ::map_to_nodes_sequentially(std::string_view sequence,
     std::vector<node_index> rev_path = map_sequence_to_nodes(*graph_, rev_seq);
 
     // map the forward
-    const auto *dbg_succ = dynamic_cast<const DBGSuccinct*>(graph_.get());
+    const auto *dbg_succ = get_dbg_succ(*graph_);
     if (dbg_succ && get_k() % 2) {
+        const auto *cached = dynamic_cast<const DBGSuccinctCachedView*>(graph_.get());
```
karasikov, line 100: So if it's divisible by 2, maybe we should print a warning saying that this thing is gonna be slower?

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -172,36 +185,57 @@ void CanonicalDBG::append_next_rc_nodes(node_index node,
...
-                        "Primary graph contains both forward and reverse complement: {} {} -> {} {}\t{} {}",
-                        node, graph_->get_node_sequence(node),
-                        children[c], graph_->get_node_sequence(children[c]),
-                        next, graph_->get_node_sequence(next));
-                    exit(1);
-                }
+            if (children[c] == npos) {
+                children[c] = next + offset_;
+                if (cached)
+                    cached->put_decoded_node(next, rev_seq);
```
karasikov, line 206: Why does Canonical wrapper care about the cache? Doesn't it have its own cache anyway?

karasikov: Doesn't it make sense to always keep a cached view on the graph in CanonicalDBG, so there will be no question if this thing is cached or not?

metagraph/src/cli/load/load_graph.cpp
karasikov: UP: Since canonical wrapper needs some caches, maybe it just makes sense to keep a cached view on the graph in it, and use its cache?

Then, if you want to create another cached view on that CanonicalDBG, you would copy the internal cache and create another CanonicalDBG<CachedView>.
So in addition to the initial CanonicalDBG<CachedView> you'll create another CanonicalDBG<CachedView> with independent cache, initialized from the first one, so you don't lose any things computed beforehand.

metagraph/src/graph/representation/succinct/boss.cpp
```diff
@@ -893,9 +893,10 @@ bool BOSS::compare_node_suffix(edge_index first, const TAlphabet *second) const
 
 /**
  * Given an edge index i, this function returns the k-mer sequence of its
- * source node.
+ * source node, and the node which is reached after k - size bwd steps.
```
karasikov, line 896: what is `size`?

metagraph/src/graph/representation/succinct/boss.hpp
```diff
@@ -273,6 +273,13 @@ class BOSS {
      */
     std::vector<TAlphabet> get_node_seq(edge_index i) const;
 
+    /**
+     * Given an edge index i, this function returns the k-mer sequence of its
+     * source node, and the node which is reached after k - size bwd steps.
```
karasikov, line 278: same

metagraph/src/common/caches.hpp
hmusta: For this case, it actually is...at some point we should open up a new PR to replace the external library with this

metagraph/src/graph/representation/canonical_dbg.cpp
hmusta: Caching is only supported for DBGSuccinct, whereas CanonicalDBG is more general. I guess one way around this is to use CanonicalDBG<DBGSuccinctCachedView> instead of CanonicalDBG<DeBruijnGraph>, so these if-statements can be made constexpr

metagraph/src/cli/load/load_graph.cpp
hmusta: The optimal wrapping strategy depends on the task. If you're only interested in exact k-mer query against the graph and annotations, then CanonicalDBG alone is enough for its mapping functions and won't use any caches. So, having the underlying graph be uncached is optimal.

However, for alignment, it's more optimal to have the underlying graph be a cached view since it can cache decoded nodes from map_to_nodes calls.

metagraph/src/cli/load/load_graph.hpp
hmusta: If you have a batch of very redundant short reads, or a single read of length > 1024, then yes, having a bigger cache is useful. I'll increase this value

metagraph/benchmarks/graph/bench_boss.cpp
hmusta: I split this up, so now there are two methods `load_graph` and `wrap_graph`

metagraph/src/graph/representation/canonical_dbg.cpp
hmusta: I added this warning to `load_graph.hpp`.

metagraph/src/graph/representation/canonical_dbg.cpp
hmusta: huh....I just noticed now that the `<....>` were discarded by GitHub's markdown :P

hmusta: anyway, I cleaned this part up a little. Changing the internals of CanonicalDBG to be templated is something that I think we can save for another PR

metagraph/src/cli/load/load_graph.cpp
```diff
@@ -71,5 +71,47 @@ std::shared_ptr<DeBruijnGraph> load_critical_dbg(const std::string &filename) {
...
+
+    // if alignment in both directions is not required, then there's no need to cache
+    if (ret_graph->get_mode() != DeBruijnGraph::CANONICAL && config.align_only_forwards)
+        return ret_graph;
```
karasikov, line 96: 1. So how is this function called `make_cached_graph` and it may return a graph which is not cached? Did you not ask yourself this question? If you keep it as is, you should rename it to `prepare_alignment_graph` or something like that.

2. It might be ok to just always cache. Is there any slowdown or downside of doing it if it's not required?

karasikov (changes requested):

metagraph/src/cli/load/load_graph.cpp
```diff
@@ -71,5 +71,47 @@ std::shared_ptr<DeBruijnGraph> load_critical_dbg(const std::string &filename) {
...
+    if (graph->get_mode() != DeBruijnGraph::PRIMARY) {
+        logger->error("Only primary mode graphs can be wrapped into canonical mode.");
+        exit(1);
+    }
+
+    logger->trace("Primary graph wrapped into canonical");
+    return std::make_shared<CanonicalDBG>(graph, cache_size);
+}
+
+template std::shared_ptr<CanonicalDBG> primary_to_canonical(std::shared_ptr<const DeBruijnGraph>, size_t);
```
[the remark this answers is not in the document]
hmusta: The output needs to be non-const `CanonicalDBG` for compatibility with `AnnotatedDBG`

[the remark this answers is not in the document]
karasikov: Then use the other version which initializes it from a non-const DeBruijnGraph!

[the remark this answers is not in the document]
hmusta: Every graph instance we have in the CLI is non-const, so it'll be easier to just make these functions take non-const and input and output non-const.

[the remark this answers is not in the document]
karasikov: I am not sure there is a need to change other things. So why don't use use the other initializer from a non-const DeBruijnGraph?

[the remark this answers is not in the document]
hmusta: What are you referring to when you say "other initializer". All I'm proposing is to make this single function have the signature
```c++
std::shared_ptr<graph::CanonicalDBG>
primary_to_canonical(std::shared_ptr<graph::DeBruijnGraph> graph, size_t cache_size = 100'000);
```

Any everything will compile and work as expected.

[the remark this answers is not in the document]
hmusta: And I'd have to change the signature of `make_cached_graph`
```c++
std::shared_ptr<graph::DeBruijnGraph> make_cached_graph(graph::DeBruijnGraph &graph,
                                                        const Config &config,
                                                        size_t cache_size = 100'000);
```

metagraph/tests/graph/all/test_dbg_basic.cpp
```diff
@@ -65,7 +61,7 @@ TYPED_TEST(DeBruijnGraphTest, SerializeEmpty) {
     EXPECT_TRUE(check_graph_nodes(graph));
 }
 
-TYPED_TEST(DeBruijnGraphTest, Serialize) {
```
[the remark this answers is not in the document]
hmusta: So we can either add load/serialize to `DBGSuccinctCachedView`, or to the parent `DBGWrapper`. My original plan was to add it to the latter (since I wanted to leave it up to the graph implementations to decide if they can be loaded or serialized), but you objected to that:
https://github.com/ratschlab/metagraph/pull/340#discussion_r676116419

metagraph/src/common/caches.hpp
```diff
@@ -0,0 +1,80 @@
...
+        std::unique_lock<std::shared_mutex> clear_lock(mu_);
+        values_.clear();
+        finder_.clear();
+    }
+
+  private:
+    size_t max_size_;
+
+    // TODO: switch to a more efficient storage vector
+    typedef std::list<std::pair<Key, Value>> Storage;
```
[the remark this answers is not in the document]
hmusta: I added more extensive comments to describe what's going on here

metagraph/src/graph/representation/succinct/dbg_succinct.hpp
```diff
@@ -5,11 +5,14 @@
 #include "kmer/kmer_bloom_filter.hpp"
 #include "graph/representation/base/sequence_graph.hpp"
 #include "boss.hpp"
+#include "dbg_succinct_cached.hpp"
 
 
 namespace mtg {
 namespace graph {
 
+class DBGSuccinctCachedView;
```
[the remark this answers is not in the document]
hmusta: Sure, this helped clean things up a bit.

metagraph/src/cli/load/load_graph.cpp
hmusta: I cleaned this up a bit, so now we don't need this function anymore

metagraph/tests/graph/all/test_dbg_basic.cpp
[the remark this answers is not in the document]
karasikov: You don't need to keep another non-const pointer to have load/serialize. So this doesn't contradict my last comment.

metagraph/src/cli/load/load_graph.cpp
hmusta: Also, it's not always a good idea to have a cached view. The internal caches use mutexes, so if you're using a single base graph in a multi-threaded setting, then having the cache can slow things down.

This is why I have a separate cached view for each thread

metagraph/tests/graph/all/test_dbg_basic.cpp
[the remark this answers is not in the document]
hmusta: Sure, this can be done with a `const_cast` if you trust the user to know what they're doing.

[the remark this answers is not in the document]
hmusta: anyway, I added this

metagraph/src/cli/load/load_graph.cpp
hmusta: Also, as mentioned below, having a cache leads to `map_to_nodes` locking for every node so it can populate the cache, so it's less optimal for use as a global object (i.e., each thread should have its own cached view).

metagraph/src/cli/load/load_graph.hpp
```diff
@@ -5,10 +5,16 @@
...
 #include "cli/config/config.hpp"
 
 
 namespace mtg {
+
+namespace graph {
+    class DBGSuccinct;
+}
+
 namespace cli {
```
karasikov, line 18: not needed
```suggestion
namespace mtg {
namespace cli {
```

metagraph/src/graph/representation/masked_graph.hpp
```diff
@@ -129,6 +129,23 @@ class MaskedDeBruijnGraph : public DBGWrapper<DeBruijnGraph> {
         });
     }
 
+    virtual void serialize(const std::string &) const override final {
+        throw std::runtime_error("serialize not supported for MaskedDeBruijnGraph");
+    }
+
+    virtual bool load(const std::string &) override final {
+        throw std::runtime_error("koad not supported for MaskedDeBruijnGraph");
```
karasikov, line 137:
```suggestion
        throw std::runtime_error("load not supported for MaskedDeBruijnGraph");
```

karasikov (changes requested):

metagraph/src/cli/align.cpp
```diff
@@ -84,7 +83,7 @@ template std::unique_ptr<IDBGAligner> build_aligner<DeBruijnGraph>(const DeBruij
 
 void map_sequences_in_file(const std::string &file,
                            const DeBruijnGraph &graph,
-                           const DBGSuccinct *dbg,
+                           const DBGSuccinct &dbg,
```
karasikov, line 86: why do we pass two graphs?

hmusta: I don't remember why we set it up this way, it's essentially just a dynamic cast of the first graph.

metagraph/src/graph/representation/succinct/dbg_succinct.hpp
```diff
@@ -53,13 +64,25 @@ class DBGSuccinct : public DeBruijnGraph {
                               const std::function<void(node_index)> &callback,
                               const std::function<bool()> &terminate = [](){ return false; }) const override;
 
+    // Traverse graph mapping sequence to the graph nodes
+    // and run callback for each node until the termination condition is satisfied.
+    // For each k-mer satisfying the skip condition, run the callback on npos.
+    // Guarantees that nodes are called in the same order as the input sequence.
+    // In canonical mode, non-canonical k-mers are NOT mapped to canonical ones
+    virtual void map_to_nodes_sequentially_checked(std::string_view sequence,
```
hmusta, line 72: Do you think it's worth adding this to `DeBruijnGraph` as well?

metagraph/src/cli/align.cpp
karasikov: I don't see why we need to pass both

hmusta: We actually don't, so I just got rid of it and added another dynamic cast into this function.

metagraph/src/graph/representation/succinct/dbg_succinct.hpp
karasikov: Would be better to just add this skipper to all functions `map_to_nodes_sequentially`

karasikov (changes requested):

metagraph/src/graph/representation/succinct/dbg_succinct.hpp
```diff
@@ -14,6 +16,15 @@ class DBGSuccinct : public DeBruijnGraph {
   public:
     friend class MaskedDeBruijnGraph;
 
+    /**
+     * A wrapper which caches computed node sequences for DBGSuccinct graphs.
+     * This allows for faster get_node_sequence and call_incoming_kmers calls.
+     * Traversal operations and map_to_nodes_sequentially are overridden to fill
+     * the cache alongside their normal functions.
+     * This wrapper stores ~144 bytes per cached node.
```
karasikov, line 24: doesn't this depend on the k-mer length?

hmusta: It does, so it'll be either 144, 152, or 168 depending in which kmertype is used

metagraph/src/graph/representation/succinct/dbg_succinct.hpp
```diff
@@ -195,6 +220,132 @@ class DBGSuccinct : public DeBruijnGraph {
...
+        typedef boss::BOSS::edge_index edge_index;
+        typedef boss::BOSS::TAlphabet TAlphabet;
+
+        virtual ~CachedView() {}
+
+        void call_outgoing_from_rev_comp(node_index node,
+                                         const std::function<void(node_index, TAlphabet)> &callback) const;
+
+        void call_incoming_to_rev_comp(node_index node,
+                                       const std::function<void(node_index, TAlphabet)> &callback) const;
```
karasikov, line 236: What is this method doing in this class `CachedView`?

karasikov: Move it to CanonicalDBG?

metagraph/src/graph/representation/canonical_dbg.hpp
```diff
@@ -139,6 +157,9 @@ class CanonicalDBG : public DBGWrapper<DeBruijnGraph> {
     // find all child nodes of node in the CanonicalDBG which are represented
     // in the reverse complement orientation in the underlying primary graph
     void append_next_rc_nodes(node_index node, std::vector<node_index> &children) const;
+
+    node_index get_rev_comp_node(node_index node) const;
```
hmusta, line 161: rename to `get_rev_comp_of_suffix`

metagraph/src/graph/representation/succinct/dbg_succinct.hpp
hmusta: I updated this now to reflect the new changes.

hmusta:
This should now be ready for review.

metagraph/src/graph/representation/succinct/dbg_succinct.hpp
```diff
@@ -183,6 +196,8 @@ class DBGSuccinct : public DeBruijnGraph {
 
     const mtg::kmer::KmerBloomFilter<>* get_bloom_filter() const { return bloom_filter_.get(); }
 
+    std::shared_ptr<CachedView> get_cached_view(size_t cache_size = 100'000) const;
```
karasikov, line 199: Better return `CachedView`. You can always create a shared_ptr from it (there should be a move constructor) if you need it.

metagraph/src/graph/representation/succinct/dbg_succinct.hpp
```diff
@@ -195,6 +210,108 @@ class DBGSuccinct : public DeBruijnGraph {
...
+        typedef boss::BOSS::TAlphabet TAlphabet;
+
+        virtual ~CachedView() {}
+
+        // get the encoding of the first character of this node's sequence
+        virtual TAlphabet get_first_value(edge_index i) const = 0;
+
+        /**
+         * Methods from DeBruijnGraph
+         */
```
karasikov, line 227: move this out of the class.
```
class DBGSuccinct {
  ...
  class CachedView;
};

class DBGSuccinct::CachedView {
  ...
};
```

metagraph/tests/graph/all/test_dbg_helpers.cpp
```diff
@@ -389,6 +406,21 @@ build_graph_batch<DBGSuccinctBloom<4, 50>>(uint64_t k,
...
+    auto base_graph = std::dynamic_pointer_cast<DBGSuccinct>(
+        build_graph_batch<DBGSuccinct>(k, sequences, mode)
+    );
+    auto cached = base_graph->get_cached_view();
+
+    // reset the underlying pointer to use base_graph instead of an alias of it
+    cached->set_graph(base_graph);
+    return cached;
+}
+
```
karasikov, line 423: The cached_view doesn't care how the underlying graph was constructed.

Thus, only one of `build_graph<DBGSuccinct::CachedView>` and `build_graph_batch<DBGSuccinct::CachedView>` should exist.

metagraph/src/graph/representation/succinct/dbg_succinct.hpp
```diff
@@ -14,6 +16,15 @@ class DBGSuccinct : public DeBruijnGraph {
...
     friend class MaskedDeBruijnGraph;
 
+    /**
+     * A wrapper which caches computed node sequences for DBGSuccinct graphs.
+     * This allows for faster get_node_sequence and call_incoming_kmers calls.
+     * Traversal operations and map_to_nodes_sequentially are overridden to fill
+     * the cache alongside their normal functions.
+     * Bytes stored per k-mer: ~56 (k<=21), ~64 (k<=42), ~80 (k<=85)
+     */
+    class CachedView;
```
karasikov, line 26: Where should this class be used?

I know you need it for alignment, but anything else?

This description doesn't make it clear how and in which cases this helps.
If there is a need to do some caching, where is the guarantee that this class will do a good job? Maybe there will be another caching strategy more useful for the application.

In general, I'm thinking about the expediency of this think if this indeed should be a general thing on this level rather than one of the alignment modules.

karasikov (changes requested):

metagraph/src/graph/representation/succinct/dbg_succinct.hpp
hmusta: `CachedView` is a pure abstract class, so this method can't return the actual object.

metagraph/src/graph/representation/succinct/dbg_succinct.hpp
hmusta: The caching strategy was designed to speed up these particular methods since they are the slowest of the frequently-used methods.

Any kind of operation which requires a sequence of backwards traversals can benefit from this. I don't think it's a good idea to make this an alignment module since it's a graph traversal helper and not something particular to alignment.

metagraph/tests/graph/all/test_dbg_helpers.cpp
hmusta: Having separate methods is useful because it can test that the CachedView works for graphs with (`build_graph`) and without (`build_graph_batch`) redundant dummy edges.

metagraph/src/graph/representation/succinct/dbg_succinct.hpp
karasikov: Then return unique_ptr. You can convert it to shared when needed. Otherwise, it will be more clear that this thing is a new object.

metagraph/src/graph/representation/succinct/dbg_succinct.hpp
karasikov: So what does it cache exactly, on what operation? Add this to the description.

hmusta: I updated the comment to add more details.

metagraph/src/graph/representation/succinct/dbg_succinct_cached.hpp
```diff
@@ -0,0 +1,152 @@
...
+        edge_index bwd_i = boss_->bwd(i);
+        auto fetch_bwd = decoded_cache_.TryGet(bwd_i);
+        if (!fetch_bwd) {
+            kmer_pair.second = boss_->bwd(*kmer_pair.second);
+            kmer_pair.first.to_prev(get_k(), boss_->get_node_last_value(*kmer_pair.second));
+            put_kmer(bwd_i, std::move(kmer_pair));
+        }
+
+        return c;
+    }
```
karasikov, line 39: move the bwd step to `get_kmer_pair` and call `get_kmer_pair().first[1]` instead of `get_first_value()`

metagraph/src/graph/representation/succinct/dbg_succinct_cached.hpp
```diff
@@ -0,0 +1,152 @@
...
+template <typename KmerType>
+class DBGSuccinctCachedViewImpl : public DBGSuccinct::CachedView {
+  public:
+    template <typename Graph>
+    DBGSuccinctCachedViewImpl(Graph&& graph, size_t cache_size)
+          : DBGSuccinct::CachedView(std::forward<Graph>(graph)),
+            boss_(&graph_->get_boss()),
+            decoded_cache_(cache_size) {}
+
+    virtual TAlphabet get_first_value(edge_index i) const override final {
```
karasikov, line 20: Comment: check cache and traverse to bwd to add the parent in cache

metagraph/src/graph/representation/succinct/dbg_succinct_cached.hpp
```diff
@@ -0,0 +1,152 @@
...
+        assert(i);
+
+        // Take k - 1 traversal steps to construct the node sequence. This way,
+        // the last node visited and its parent can be cached for subsequence calls.
+        auto kmer_pair = get_kmer_pair(i, true);
+        assert(kmer_pair.second);
+
+        TAlphabet c = kmer_pair.first[1];
+
+        edge_index bwd_i = boss_->bwd(i);
```
karasikov, line 30: You could also store this bwd(i) in cache, to avoid this extra call

metagraph/src/graph/representation/succinct/dbg_succinct_cached.hpp
```diff
@@ -0,0 +1,152 @@
...
+        auto fetch_bwd = decoded_cache_.TryGet(bwd_i);
+        if (!fetch_bwd) {
+            kmer_pair.second = boss_->bwd(*kmer_pair.second);
+            kmer_pair.first.to_prev(get_k(), boss_->get_node_last_value(*kmer_pair.second));
+            put_kmer(bwd_i, std::move(kmer_pair));
+        }
```
karasikov, line 36: An alternative:
instead of
```
i <- bwd(i) <- ... (k bwd steps) ... <- bwd(...(...))
^    *                                  ^
```
cache
```
i <- parent <- ... (arbitrary back steps) ... <- k-order-parent
^    ^                                           ^
```
and now you can cache arbitrary paths, not necessarily bwd ... as before

karasikov (changes requested):

metagraph/src/graph/representation/succinct/dbg_succinct_cached.hpp
```diff
@@ -0,0 +1,152 @@
...
+        const_cast<DBGSuccinct*>(graph_.get())->add_sequence(sequence, on_insertion);
+        decoded_cache_.Clear();
+    }
+
+  private:
+    const boss::BOSS *boss_;
+
+    // KmerType is the encoded k-mer
+    // edge_index is the boss node whose last character is the first character of the k-mer
+    typedef std::pair<KmerType, std::optional<edge_index>> CacheValue;
```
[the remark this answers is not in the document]
karasikov: Let's put this suggestion on hold and see how things look after implementing the other changes

metagraph/src/graph/representation/succinct/dbg_succinct_cached.hpp
```diff
@@ -0,0 +1,152 @@
...
+            if (std::optional<CacheValue> kmer = decoded_cache_.TryGet(key)) {
+                if (kmer->second) {
+                    assert(kmer->second == value.second);
+                    return;
+                }
+            }
+        }
+
+        decoded_cache_.Put(key, std::move(value));
+    }
```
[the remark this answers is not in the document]
karasikov: Let's put this suggestion on hold

PR #335, opened 2021-07-12
Chain local alignments to improve accuracy of long read alignment

hmusta:
A chain is a single alignment for a query that is constructed by concatenating local alignments to that query. If a few nodes (< k) are missing from the graph whose presence would allow for two alignments to be merged, then an affine gap penalty is added when merging them.

This replaces #212 
- [x] Use dynamic programming to create chains of local alignments
- [x] If two alignments overlap, trim prefix of second alignment to disjoin them
- [x] Assign gap scores for each disjoint alignment in a chain
- [x] Construct a single Alignment from a chain for simplicity
- [x] More extensive unit tests


metagraph/src/cli/config/config.cpp
```diff
@@ -218,6 +218,8 @@ Config::Config(int argc, char *argv[]) {
             align_one_strand = true;
         } else if (!strcmp(argv[i], "--align-edit-distance")) {
             alignment_edit_distance = true;
+        } else if (!strcmp(argv[i], "--align-chain")) {
+            alignment_chain = true;
```
karasikov, line 222: Add a line to the usage description? Otherwise, no one will know that this parameter exists.

metagraph/src/graph/alignment/aligner_alignment.cpp
```diff
@@ -55,19 +55,21 @@ std::ostream& operator<<(std::ostream& out, const Alignment &alignment) {
...
     assert(orientation_ == other.orientation_);
-    assert(cigar_.empty() || cigar_.back().first != Cigar::CLIPPED);
 
     nodes_.insert(nodes_.end(), other.nodes_.begin(), other.nodes_.end());
     sequence_ += std::move(other.sequence_);
     score_ += other.score_;
 
     cigar_.append(std::move(other.cigar_));
-    query_ = { query_.data(), query_.size() + other.query_.size() };
+    query_ = { query_.data(), other.query_.size() + (other.query_.data() - query_.data()) };
```
karasikov, line 66: Are you sure about the second argument? Describe what it represents in a comment.

metagraph/src/graph/alignment/aligner_alignment.cpp
```diff
@@ -507,13 +599,11 @@ ::load_from_json(const Json::Value &alignment, const DeBruijnGraph &graph) {
...
-                *(nodes_.rbegin() + 1),
-                [&](auto node, char c) {
-                    if (node == nodes_.back())
-                        sequence_ += c;
-                }
-            );
+            graph.call_outgoing_kmers(*(nodes_.rbegin() + 1),
+                                      [&](auto node, char c) {
+                if (node == nodes_.back())
+                    sequence_ += c;
```
karasikov, line 605:
```suggestion
                    sequence_.append(c);
```

metagraph/src/graph/alignment/aligner_alignment.cpp
```diff
@@ -574,36 +746,54 @@ bool spell_path(const DeBruijnGraph &graph,
...
-        });
-
-        if (!next) {
-            std::cerr << "ERROR: invalid edge " << path[i - 1] << " " << path[i] << std::endl;
+        if (num_dummy > graph.get_k()) {
+            std::string out_msg;
+            for (DeBruijnGraph::node_index node : path) {
+                out_msg += fmt::format("{} ", node);
+            }
+            logger->error("Too many dummy nodes\n{}", out_msg);
```
karasikov, line 766:
```suggestion
            logger->error("Too many dummy nodes\n{}", fmt::join(path, " "));
```

metagraph/src/graph/alignment/aligner_alignment.hpp
```diff
@@ -108,6 +108,13 @@ class Alignment {
 
     size_t trim_offset();
 
+    size_t trim_query_prefix(size_t n, const DeBruijnGraph &graph, const DBGAlignerConfig &config);
+
+    // When chaining together two alignments, use this method to adapt the refix
```
karasikov, line 113:
```suggestion
    // When chaining together two alignments, use this method to adapt the prefix
```

metagraph/src/graph/alignment/aligner_cigar.hpp
```diff
@@ -66,6 +67,9 @@ class Cigar {
         cigar_.insert(it, std::forward<Args>(args)...);
     }
 
+    template <typename... Args>
+    void erase(Args&&... args) { cigar_.erase(std::forward<Args>(args)...); }
+
     value_type& front() { return cigar_.front(); }
     value_type& back() { return cigar_.back(); }
     const value_type& front() const { return cigar_.front(); }
```
karasikov, line 75: so you allow doing anything with `cigar_`. Just add a getter, and then you don't need to write all these begin/end/insert/erase/...

karasikov (changes requested):

metagraph/src/graph/alignment/aligner_alignment.cpp
hmusta: ok. this is one of the strange things about `std::string_view`, it takes a length as the second argument instead of an end pointer (they added the latter in C++20)

[https://en.cppreference.com/w/cpp/string/basic_string_view/basic_string_view](url)

karasikov: yeah, but is that a length though?
`other.query_.size() + (other.query_.data() - query_.data())`

hmusta: I guess a simpler way to write it would be
```suggestion
    query_ = std::string_view(query_.data(), other.query_.end() - query_.begin());
```

before, it was complaining about passing a signed integer for the size, but that'll probably avoided by not initializing with `{}`

karasikov: Ok, looks simpler, it's still not obvious, why can you subtract end of one string from begin of another?
Are they segments of the same string and you want to take their union?

If so, writing this in a comment would help.

hmusta: Essentially, `query_` is a `string_view` of a much longer input sequence.

To avoid confusion, I think I should just rename `query_` to something like `query_view_`

metagraph/src/graph/alignment/aligner_cigar.hpp
hmusta: I cleaned this up a bit, but I still kept the `begin() const` and `end() const` to make it easier to iterate through a `Cigar` object

metagraph/src/graph/alignment/aligner_alignment.cpp
hmusta: I added this as a TODO

metagraph/src/graph/alignment/aligner_cigar.hpp
```diff
@@ -42,67 +45,65 @@ class Cigar {
...
+            cigar_.front().second += n;
+        }
     }
 
-    value_type& front() { return cigar_.front(); }
-    value_type& back() { return cigar_.back(); }
-    const value_type& front() const { return cigar_.front(); }
-    const value_type& back() const { return cigar_.back(); }
+    const_iterator begin() const { return cigar_.cbegin(); }
+    const_iterator end() const { return cigar_.cend(); }
```
karasikov, line 86: If you want to use range-iteration, why not write `auto x : cigar.data()`? So these begin/end would be redundant.

karasikov (approved):

metagraph/src/graph/alignment/aligner_cigar.hpp
hmusta: yes, but these iterators only add four extra lines, so it's not adding that much redundancy...

karasikov: I know, but there are still no benefits of having them, imo. It just adds another way to do the same thing, which is generally not a good thing.

PR #336, opened 2021-07-12
Alignment to subgraphs using labels and coordinates

hmusta:
Replaces and extends #150, #262, and #279 
- [x] Restrict backtracking to label- and coordinate-consistent paths
- [x] Load coordinates for each node
- [x] Handle coordinates for alignment in reverse direction (there's a regression somewhere...)
- [x] Label- and coordinate-consistent chaining
- [ ] Unit tests for coordinates and chaining

### DIFF
--- a/metagraph/src/cli/align.cpp
+++ b/metagraph/src/cli/align.cpp
@@ -305,14 +305,14 @@ std::string format_alignment(std::string_view header,
     if (!config.output_json) {
         sout += fmt::format("{}\t{}", header, paths.get_query());
         if (paths.empty()) {
-            sout += fmt::format("\t*\t*\t{}\t*\t*\t*", config.alignment_min_path_score);
+            sout += fmt::format("\t*\t*\t{}\t*\t*\t*\n", config.alignment_min_path_score);
         } else {
-            for (size_t i = 0; i < paths.size(); ++i) {
-                sout += fmt::format("\t{}", paths[i]);
+            for (const auto &path : paths.data()) {
+                sout += fmt::format("\t{}", path);
             }
+            sout += "\n";
         }
 
-        sout += "\n";
     } else {
         Json::StreamWriterBuilder builder;
         builder["indentation"] = "";
--- a/metagraph/src/cli/align.cpp
+++ b/metagraph/src/cli/align.cpp
@@ -347,36 +347,6 @@ std::string format_alignment(std::string_view header,
     return sout;
 }
 
-template <class LabelEncoder>
-std::string format_alignment_labels(const DeBruijnGraph &graph,
-                                    const LabelEncoder &label_encoder,
-                                    const Config &config,
-                                    std::string_view header,
-                                    QueryAlignment&& paths) {
-    std::vector<std::string> labels;
-
-    for (const auto &path : paths.data()) {
-        if (path.label_columns.empty()) {
-            labels.emplace_back("\t*");
-        } else {
-            auto &cur_label = labels.emplace_back("\t");
-            for (size_t i = 0; i < path.label_columns.size(); ++i) {
-                cur_label += label_encoder.decode(path.label_columns[i]);
-                if (path.label_coordinates.size()) {
-                    assert(path.label_coordinates.size() == path.label_columns.size());
-                    for (const auto &[first, last] : path.label_coordinates[i]) {
-                        cur_label += fmt::format(":{}-{}", first, last);
-                    }
-                }
-                cur_label += ";";
-            }
-            cur_label.pop_back();
-        }
-    }
-
-    return format_alignment(header, paths, graph, config, &labels);
-}
-
 int align_to_graph(Config *config) {
     assert(config);
 
@@ -492,28 +462,23 @@ int align_to_graph(Config *config) {
                 if (auto *canonical = dynamic_cast<CanonicalDBG*>(aln_graph.get()))
                     aln_graph = std::make_shared<CanonicalDBG>(*canonical);
 
+                std::shared_ptr<AnnotatedDBG> aln_anno_graph;
+                std::shared_ptr<IDBGAligner> aligner;
+
                 if (annotator) {
-                    AnnotatedDBG aln_anno_graph(aln_graph, annotator);
-                    const auto &label_encoder = annotator->get_label_encoder();
-                    build_aligner(aln_anno_graph, aligner_config)->align_batch(
-                        batch, [&](std::string_view header, auto&& paths) {
-                        std::string res = format_alignment_labels(
-                            *aln_graph, label_encoder, *config,
-                            header, std::move(paths)
-                        );
-                        std::lock_guard<std::mutex> lock(print_mutex);
-                        *out << res;
-                    });
+                    aln_anno_graph = std::make_shared<AnnotatedDBG>(aln_graph, annotator);
+                    aligner = build_aligner(*aln_anno_graph, aligner_config);
                 } else {
-                    build_aligner(*aln_graph, aligner_config)->align_batch(
-                        batch, [&](std::string_view header, auto&& paths) {
-                        std::string res = format_alignment(
-                            header, std::move(paths), *aln_graph, *config
-                        );
-                        std::lock_guard<std::mutex> lock(print_mutex);
-                        *out << res;
-                    });
+                    aligner = build_aligner(*aln_graph, aligner_config);
                 }
+
+                aligner->align_batch(batch, [&](std::string_view header, auto&& paths) {
+                    std::string res = format_alignment(
+                        header, std::move(paths), *aln_graph, *config
+                    );
+                    std::lock_guard<std::mutex> lock(print_mutex);
+                    *out << res;
+                });
             };
 
             ++num_batches;
--- a/metagraph/src/cli/align.cpp
+++ b/metagraph/src/cli/align.cpp
@@ -428,11 +428,11 @@ int align_to_graph(Config *config) {
                 std::shared_ptr<DeBruijnGraph> aln_graph(
                     std::shared_ptr<DeBruijnGraph>{}, graph.get()
                 );
+
                 if (auto *canonical = dynamic_cast<CanonicalDBG*>(aln_graph.get()))
                     aln_graph = std::make_shared<CanonicalDBG>(*canonical);
 
-                std::unique_ptr<IDBGAligner> aligner;
-                aligner = build_aligner(*aln_graph, aligner_config);
+                auto aligner = build_aligner(*aln_graph, aligner_config);
 
                 aligner->align_batch(batch, [&](std::string_view header, auto&& paths) {
                     std::string res = format_alignment(header, paths, *aln_graph, *config);
--- a/metagraph/src/cli/align.cpp
+++ b/metagraph/src/cli/align.cpp
@@ -5,12 +5,16 @@
 #include "common/logger.hpp"
 #include "common/unix_tools.hpp"
 #include "common/threads/threading.hpp"
+#include "common/utils/string_utils.hpp"
 #include "graph/representation/succinct/dbg_succinct.hpp"
 #include "graph/representation/canonical_dbg.hpp"
 #include "graph/alignment/dbg_aligner.hpp"
+#include "graph/alignment/aligner_labeled.hpp"
+#include "graph/annotated_dbg.hpp"
 #include "seq_io/sequence_io.hpp"
 #include "config/config.hpp"
 #include "load/load_graph.hpp"
+#include "load/load_annotated_graph.hpp"
 
 namespace mtg {
 namespace cli {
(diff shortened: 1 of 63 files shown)


metagraph/src/common/algorithms.hpp
```diff
@@ -97,6 +97,73 @@ namespace utils {
...
     }
 
+    // Return true if the two sorted ranges share a common element
+    template <class InputIt1, class InputIt2>
+    constexpr bool share_element(InputIt1 first1,
+                                 InputIt1 last1,
+                                 InputIt2 first2,
+                                 InputIt2 last2) {
+        assert(std::is_sorted(first1, last1));
+        assert(std::is_sorted(first2, last2));
```
karasikov, line 107: last -> end ?

karasikov (changes requested):

metagraph/src/common/algorithms.hpp
hmusta: sure, we can do `begin` and `end`. I was just copying the names used on cppreference

metagraph/src/graph/annotated_dbg.cpp
```diff
@@ -743,6 +738,8 @@ AnnotatedDBG::get_top_labels(const std::vector<std::pair<row_index, size_t>> &in
     assert(check_compatibility());
 
     if (with_kmer_counts) {
+        // TODO: Don't take into account counts when comparing to min_count.
+        //       It should be compared to sum_rows and not sum_row_values.
```
karasikov, line 742: I think I have fixed this

metagraph/src/common/algorithms.hpp
```diff
@@ -66,37 +66,93 @@ namespace utils {
...
+    }
 
-            if (*first_begin == *second_begin) {
-                ++count;
-                ++first_begin;
-                ++second_begin;
+    // Intersect the sorted ranges a1 and b1 with corresponding sorted ranges of
+    // sorted ranges a2 and b2 (of equal length).
+    // i.e., For each shared element between a1 and b1, intersect the corresponding
+    // ranges in a2 and b2.
```
karasikov, line 115: This is unclear

metagraph/src/cli/align.cpp
```diff
@@ -74,13 +77,25 @@ DBGAlignerConfig initialize_aligner_config(const Config &config) {
     return aligner_config;
 }
 
+template <class Graph, template <class ... Types> class Aligner>
+inline std::unique_ptr<graph::align::IDBGAligner>
+build_aligner_infer_args(const Graph &graph, const DBGAlignerConfig &aligner_config) {
+    return std::make_unique<Aligner<>>(graph, aligner_config);
+}
```
karasikov, line 84: Why is this needed? Can't you just write `std::make_unique<...>(graph, aligner_config);`?

metagraph/src/cli/align.cpp
```diff
@@ -290,14 +305,14 @@ std::string format_alignment(std::string_view header,
     if (!config.output_json) {
         sout += fmt::format("{}\t{}", header, paths.get_query());
         if (paths.empty()) {
-            sout += fmt::format("\t*\t*\t{}\t*\t*\t*", config.alignment_min_path_score);
+            sout += fmt::format("\t*\t*\t{}\t*\t*\t*\n", config.alignment_min_path_score);
         } else {
             for (const auto &path : paths.data()) {
                 sout += fmt::format("\t{}", path);
             }
+            sout += "\n";
```
karasikov, line 313: use `fmt::join`
```suggestion
            sout += fmt::format("\t{}\n", fmt::join(paths.data(), "\t"));
```

metagraph/src/cli/align.cpp
```diff
@@ -432,13 +454,24 @@ int align_to_graph(Config *config) {
...
 
-                aligner->align_batch(batch, [&](std::string_view header, auto&& paths) {
-                    std::string res = format_alignment(header, paths, *aln_graph, *config);
-                    std::lock_guard<std::mutex> lock(print_mutex);
-                    *out << res;
-                });
+                aligner->align_batch(batch,
+                    [&](std::string_view header, QueryAlignment&& paths) {
+                        std::string res = format_alignment(header, std::move(paths),
+                                                           *aln_graph, *config);
```
karasikov, line 470:
```suggestion
                        std::string res = format_alignment(header, paths, *aln_graph, *config);
```

metagraph/src/cli/load/load_annotated_graph.cpp
```diff
@@ -32,9 +32,11 @@ std::unique_ptr<AnnotatedDBG> initialize_annotated_dbg(std::shared_ptr<DeBruijnG
         logger->trace("Primary graph was wrapped into canonical");
     }
 
-    auto annotation_temp = config.infbase_annotators.size()
+    std::shared_ptr<AnnotatedDBG::Annotator> annotation_temp{
```
karasikov, line 35:
```suggestion
    std::shared_ptr<AnnotatedDBG::Annotator> annotation_temp {
```

metagraph/src/cli/load/load_annotated_graph.cpp
```diff
@@ -73,8 +75,7 @@ std::unique_ptr<AnnotatedDBG> initialize_annotated_dbg(std::shared_ptr<DeBruijnG
     }
 
     // load graph
-    auto anno_graph
-            = std::make_unique<AnnotatedDBG>(std::move(graph), std::move(annotation_temp));
+    auto anno_graph = std::make_unique<AnnotatedDBG>(std::move(graph), annotation_temp);
```
karasikov, line 78:
```suggestion
    auto anno_graph = std::make_unique<AnnotatedDBG>(graph, annotation_temp);
```

metagraph/src/common/algorithms.hpp
```diff
@@ -66,37 +66,93 @@ namespace utils {
...
 
         uint64_t count = 0;
 
-        while (first_begin != first_end && second_begin != second_end) {
-            first_begin = std::lower_bound(first_begin, first_end, *second_begin);
+        while (a_begin != a_end && b_begin != b_end) {
+            if (*a_begin < *b_begin) {
+                a_begin = std::lower_bound(a_begin, a_end, *b_begin);
+            } else if (*b_begin < *a_begin) {
+                b_begin = std::lower_bound(b_begin, b_end, *a_begin);
```
karasikov, line 82: If this is called a lot, would be good to check if linear search is indeed not faster.

metagraph/src/graph/alignment/dbg_aligner.hpp
```diff
@@ -94,7 +94,8 @@ class ISeedAndExtendAligner : public IDBGAligner {
                     const ISeeder &seeder,
                     IExtender &extender,
                     const std::function<void(Alignment&&)> &callback,
-                    const std::function<score_t(const Alignment&)> &get_min_path_score) const;
+                    const std::function<score_t(const Alignment&)> &get_min_path_score,
+                    bool force_fixed_seed) const;
```
karasikov, line 98: what is `force_fixed_seed`?

metagraph/src/graph/alignment/dbg_aligner.cpp
```diff
@@ -63,7 +63,7 @@ ::align_batch(const std::vector<IDBGAligner::Query> &seq_batch,
         auto seeder = build_seeder(this_query, is_reverse_complement, nodes);
         auto extender = build_extender(this_query, aggregator);
 
-        size_t num_explored_nodes = 0;
+        double num_explored_nodes = 0;
```
karasikov, line 66: why double?

metagraph/src/graph/alignment/aligner_labeled.hpp
```diff
@@ -0,0 +1,278 @@
...
+    typedef annot::binmat::BinaryMatrix::Column Column;
+    typedef annot::binmat::BinaryMatrix::Row Row;
+    typedef annot::matrix::MultiIntMatrix::Tuple Tuple;
+
+    typedef std::reference_wrapper<const Alignment::LabelSet> LabelSet;
+    typedef std::reference_wrapper<const Alignment::CoordinateSet> CoordsSet;
+
+    static constexpr Row nrow = std::numeric_limits<Row>::max();
+
+    DynamicLabeledGraph(const AnnotatedDBG &anno_graph);
```
karasikov, line 38: Why do you need this guy?

metagraph/src/graph/alignment/aligner_alignment.hpp
```diff
@@ -49,7 +57,8 @@ class Alignment {
               bool orientation = false,
               size_t offset = 0);
 
-    void append(Alignment&& other);
+    // return true if the internal labels were modified
+    bool append(Alignment&& other);
```
karasikov, line 61: Does it only add labels or may remove some as well?

karasikov (changes requested):

metagraph/src/graph/alignment/aligner_labeled.hpp
hmusta: Maybe I should call this `AnnotationBuffer` instead?

metagraph/src/graph/alignment/aligner_alignment.hpp
hmusta: This concatenates two alignments and intersects their labels. If the alignments have coordinates, then it makes sure that the two alignments have adjacent index ranges

metagraph/src/graph/alignment/dbg_aligner.cpp
hmusta: For readability, this should probably stay as `size_t`. I changed it to clean up some of the `static_cast`s in the log message below.

metagraph/src/graph/alignment/dbg_aligner.hpp
hmusta: I'll write a comment about this. If it's true, then it forces the extender to use the full length of the seed in alignment (i.e., the resulting alignment has the seed as a prefix). Otherwise, the extender just uses the seed as a starting node in the graph and can take whichever path it chooses.

hmusta: It's done this way because the seeder does not take coordinates or labels into account, so parts of the input seed may have to be ignored.

hmusta: However, for the backwards traversal part of alignment, we know that the input seed (a forwards alignment) is coordinate consistent, so we can tell the extender to use the full seed.

metagraph/src/graph/alignment/aligner_labeled.hpp
karasikov: maybe

metagraph/src/common/algorithms.hpp
hmusta: I guess it's very input-dependent...I'll double-check gcc's implementation to see if we can gain any insights from it.

metagraph/src/graph/alignment/aligner_alignment.hpp
karasikov: Sorry, my question was about this comment: "... internal labels were modified"

metagraph/src/cli/align.cpp
hmusta: Actually, there's a mistake here, I'm not passing template arguments for the Aligner construction...

Anyway, this is needed because `ILabeledAligner` takes an `AnnotatedDBG` as input instead of `DeBruijnGraph`

metagraph/src/cli/align.cpp
hmusta: I tried this, but it doesn't compile because `formatter<Alignment>` needs to be implemented. I guess if we do need to use `fmt::join` more often in the future, I'll implement it.

metagraph/src/graph/alignment/aligner_alignment.hpp
hmusta: ah, ok. So I'll expand the comment to describe exactly what this function is doing. It returns true if some of the labels or coordinates had to be discarded during the append.

metagraph/src/cli/align.cpp
karasikov: Hm, I still don't see why you need this function.

hmusta: I ended up removing it...I actually don't remember why I needed the helper...

metagraph/src/graph/alignment/dbg_aligner.cpp
hmusta: I changed this back and kept the static_cast

metagraph/src/common/algorithms.hpp
hmusta: hmm, looks like the implementation in libstdc++ does a linear search
https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/include/bits/stl_algo.h#L5241

hmusta: same with clang

karasikov: yeah, I think that should generally work faster, unless you have very large sets with a small intersection.

PR #337, opened 2021-07-14
query k-mer coordinates

karasikov:


metagraph/src/annotation/int_matrix/rank_extended/tuple_csc_matrix.hpp
```diff
@@ -0,0 +1,203 @@
...
+            assert(r >= 1 && "matches can't have zero-rank");
+            size_t begin = delimiters_[j].select1(r) + 1 - r;
+            size_t end = delimiters_[j].select1(r + 1) - r;
+            Tuple tuple;
+            tuple.reserve(end - begin);
+            for (size_t t = begin; t < end; ++t) {
+                tuple.push_back(column_values_[j][t]);
+            }
+            row_tuples[i].emplace_back(j, std::move(tuple));
+        }
```
hmusta, line 156: I agree, I think it might be more cache efficient to first aggregate a list of all accessed column IDs, then iterate through each column once to populate this vector.

hmusta (changes requested):
Looks good so far. I'll propose a list of TODOs
- [ ] A few integration tests to make sure serializing/loading works
- [ ] Unit tests
- [ ] Make get_row_tuples in TupleCSCMatrix more cache efficient
- [ ] Implement TupleCSRMatrix (optional?)

metagraph/src/graph/annotated_dbg.hpp
```diff
@@ -122,6 +123,11 @@ class AnnotatedDBG : public AnnotatedSequenceGraph {
                               double presence_ratio,
                               const std::vector<double> &count_quantiles) const;
 
+    std::vector<std::pair<Label, std::vector<SmallVector<uint64_t>>>>
+    get_kmer_coordinates(std::string_view sequence,
```
hmusta, line 127: How about splitting this function into two: one which takes a sequence, and another which takes annotation indices (like the other methods)? It'll be more efficient to query if the nodes have already been mapped.

karasikov: done

metagraph/src/cli/query.cpp
```diff
@@ -85,6 +86,26 @@ std::string QueryExecutor::execute_query(const std::string &seq_name,
...
+
+        output += seq_name;
+
+        for (const auto &[label, tuples] : result) {
+            output += "\t<" + label + ">";
+            for (const auto &coords : tuples) {
+                output += ":";
+                output += fmt::format("{}", fmt::join(coords, ","));
+            }
+        }
```
hmusta, line 105: This produces a lot of unnecessary output for sequences with long matching stretches. So how about reporting something like `<chrX>:1-2:5-6` (with inclusive ranges similar to what `samtools` does)

metagraph/src/graph/annotated_dbg.cpp
```diff
@@ -376,6 +377,114 @@ AnnotatedDBG::get_label_count_quantiles(std::string_view sequence,
...
+        } else {
+            ids.push_back(-1);
+        }
+    }
+
+    uint64_t min_count = std::max(1.0, std::ceil(presence_ratio * path.size()));
+    if (rows.size() < min_count)
+        return {};
+
+    auto rows_tuples = dynamic_cast<const MultiIntMatrix &>(annotator_->get_matrix())
```
hmusta, line 432: When an annotator not supporting coordinates is loaded, this throws a `std::bad_cast` without giving more useful feedback. Either catch this is `load_annotated_graph`, or here and rethrow it with a more useful message.

hmusta (changes requested):

metagraph/src/cli/query.cpp
karasikov: I suggest we use this for now and change it if we see that it's objectively better or others request this.

metagraph/src/graph/annotated_dbg.cpp
karasikov: done

metagraph/src/cli/query.cpp
hmusta: I was suggesting this to make the output more human readable, not so much for performance reasons.

karasikov: I think the ease of parsing is more important. Not clear which is easier to parse, though. I suggest we keep things as is and think about this and change this when we collect more data.

PR #339, opened 2021-07-23
filter queries by the number of k-mers matched to the joint graph

karasikov:


metagraph/src/cli/config/config.cpp
```diff
@@ -1191,6 +1196,7 @@ void Config::print_usage(const std::string &prog_name, IdentityType identity) {
             fprintf(stderr, "\t   --print-signature \t\tprint vectors indicating present/absent k-mers [off]\n");
             fprintf(stderr, "\t   --num-top-labels \t\tmaximum number of frequent labels to print [off]\n");
             fprintf(stderr, "\t   --discovery-fraction [FLOAT] fraction of labeled k-mers required for annotation [0.7]\n");
+            fprintf(stderr, "\t   --presence-fraction [FLOAT] \tfraction of k-mers required for presence in the graph [0.0]\n");
```
hmusta, line 1199: This sounds a little confusing, what about this?
```suggestion
            fprintf(stderr, "\t   --presence-fraction [FLOAT] \tfraction of k-mers required to be present in the graph [0.0]\n");
```

metagraph/src/graph/annotated_dbg.hpp
```diff
@@ -93,9 +93,11 @@ class AnnotatedDBG : public AnnotatedSequenceGraph {
 
     /*********************** Special queries **********************/
 
-    // return labels that occur at least in |presence_ratio| k-mers
+    // return labels that occur at least in |discovery_fraction| k-mers
+    // but skip the sequence if fewer than |presence_fraction| k-mers are matched
```
hmusta, line 97:
```suggestion
    // but skip the sequence if fewer than |presence_fraction| k-mers are matched against the graph
```

PR #340, opened 2021-07-24
Added DBGWrapper as an abstract parent class for all DeBruijnGraph wrappers

hmusta:
This was grabbed from #333. It adds the `DBGWrapper` abstract parent class and `SequenceGraph::get_base_graph` (for easy type checking)

### DIFF
--- a/metagraph/src/graph/representation/base/dbg_wrapper.hpp
+++ b/metagraph/src/graph/representation/base/dbg_wrapper.hpp
@@ -215,6 +215,110 @@ class DBGWrapper : public DeBruijnGraph {
     std::shared_ptr<const Graph> graph_;
 };
 
+template <typename Graph = DeBruijnGraph>
+class DBGNodeModifyingWrapper : public DBGWrapper<Graph> {
+  public:
+    typedef typename Graph::node_index node_index;
+    typedef typename Graph::CallPath CallPath;
+    typedef typename Graph::OutgoingEdgeCallback OutgoingEdgeCallback;
+    typedef typename Graph::IncomingEdgeCallback IncomingEdgeCallback;
+
+    template <typename... Args>
+    explicit DBGNodeModifyingWrapper(Args&&... args)
+          : DBGWrapper<Graph>(std::forward<Args>(args)...) {}
+
+    virtual ~DBGNodeModifyingWrapper() {}
+
+    virtual void map_to_nodes(std::string_view sequence,
+                              const std::function<void(node_index)> &callback,
+                              const std::function<bool()> &terminate
+                                  = [](){ return false; }) const override = 0;
+
+    virtual void map_to_nodes_sequentially(std::string_view sequence,
+                                           const std::function<void(node_index)> &callback,
+                                           const std::function<bool()> &terminate
+                                               = [](){ return false; }) const override = 0;
+
+    virtual void
+    adjacent_outgoing_nodes(node_index node,
+                            const std::function<void(node_index)> &callback) const override = 0;
+
+    virtual void
+    adjacent_incoming_nodes(node_index node,
+                            const std::function<void(node_index)> &callback) const override = 0;
+
+    virtual void call_outgoing_kmers(node_index kmer,
+                                     const OutgoingEdgeCallback &callback) const override = 0;
+
+    virtual void call_incoming_kmers(node_index kmer,
+                                     const IncomingEdgeCallback &callback) const override = 0;
+
+    virtual void call_nodes(const std::function<void(node_index)> &callback,
+                            const std::function<bool()> &stop_early
+                                = [](){ return false; }) const override = 0;
+    virtual void call_kmers(const std::function<void(node_index, const std::string&)> &callback) const override = 0;
+
+    virtual uint64_t num_nodes() const override = 0;
+    virtual uint64_t max_index() const override = 0;
+
+    virtual std::string get_node_sequence(node_index index) const override = 0;
+
+    virtual node_index traverse(node_index node, char next_char) const override = 0;
+    virtual node_index traverse_back(node_index node, char prev_char) const override = 0;
+
+    virtual size_t outdegree(node_index node) const override = 0;
+    virtual size_t indegree(node_index node) const override = 0;
+
+    virtual void traverse(node_index start,
+                          const char *begin,
+                          const char *end,
+                          const std::function<void(node_index)> &callback,
+                          const std::function<bool()> &terminate = [](){ return false; }) const override {
+        DeBruijnGraph::traverse(start, begin, end, callback, terminate);
+    }
+
+    virtual void call_sequences(const CallPath &callback,
+                                size_t num_threads = 1,
+                                bool kmers_in_single_form = false) const override {
+        DeBruijnGraph::call_sequences(callback, num_threads, kmers_in_single_form);
+    }
+
+    virtual void call_unitigs(const CallPath &callback,
+                              size_t num_threads = 1,
+                              size_t min_tip_size = 1,
+                              bool kmers_in_single_form = false) const override {
+        DeBruijnGraph::call_unitigs(callback, num_threads, min_tip_size, kmers_in_single_form);
+    }
+
+    virtual bool has_single_outgoing(node_index node) const override {
+        return DeBruijnGraph::has_single_outgoing(node);
+    }
+
+    virtual bool has_multiple_outgoing(node_index node) const override {
+        return DeBruijnGraph::has_multiple_outgoing(node);
+    }
+
+    virtual bool has_no_incoming(node_index node) const override {
+        return DeBruijnGraph::has_no_incoming(node);
+    }
+
+    virtual bool has_single_incoming(node_index node) const override {
+        return DeBruijnGraph::has_single_incoming(node);
+    }
+
+    virtual node_index kmer_to_node(std::string_view kmer) const override {
+        return DeBruijnGraph::kmer_to_node(kmer);
+    }
+
+    virtual bool find(std::string_view sequence, double discovery_fraction = 1) const override {
+        return DeBruijnGraph::find(sequence, discovery_fraction);
+    }
+
+    virtual void call_source_nodes(const std::function<void(node_index)> &callback) const override {
+        return DeBruijnGraph::call_source_nodes(callback);
+    }
+};
+
 } // namespace graph
 } // namespace mtg
 
--- a/metagraph/src/graph/representation/base/dbg_wrapper.hpp
+++ b/metagraph/src/graph/representation/base/dbg_wrapper.hpp
@@ -71,7 +71,7 @@ class IDBGWrapper : public DeBruijnGraph {
 };
 
 /**
- * This wrapper uses the default methods from DeBruijnGraph when available
+ * This wrapper uses the default methods from Graph when available
  */
 template <typename Graph = DeBruijnGraph>
 class DBGNodeModifyingWrapper : public IDBGWrapper<Graph> {
@@ -85,8 +85,7 @@ class DBGNodeModifyingWrapper : public IDBGWrapper<Graph> {
     virtual ~DBGNodeModifyingWrapper() {}
 
     /**
-     * These methods should be implemented. The DeBruijnGraph defaults cannot
-     * be used.
+     * These methods should be implemented. The Graph defaults cannot be used.
      */
     virtual void call_nodes(const std::function<void(node_index)> &callback,
                             const std::function<bool()> &stop_early
--- a/metagraph/src/graph/representation/base/dbg_wrapper.hpp
+++ b/metagraph/src/graph/representation/base/dbg_wrapper.hpp
@@ -2,7 +2,6 @@
 #define __DBG_WRAPPER__
 
 #include "sequence_graph.hpp"
-#include "common/logger.hpp"
 
 
 namespace mtg {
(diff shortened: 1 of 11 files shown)


metagraph/src/graph/representation/base/dbg_wrapper.hpp
```diff
@@ -0,0 +1,325 @@
...
+    }
+
+    virtual ~DBGWrapper() {}
+
+    /**
+     * Added methods
+     */
+    virtual const Graph& get_graph() const { return *graph_; }
+    virtual std::shared_ptr<const Graph> get_graph_ptr() const { return graph_; }
+    virtual std::shared_ptr<Graph> get_mutable_graph_ptr() const { return graph_ptr_; }
```
karasikov, line 47: Any reason to have this mutable version as well? I think this is not needed at the moment and can be removed for simplicity.

add_sequence should just throw errors

metagraph/src/graph/representation/base/dbg_wrapper.hpp
```diff
@@ -0,0 +1,325 @@
...
+                                           const std::function<void(node_index)> &callback,
+                                           const std::function<bool()> &terminate
+                                               = [](){ return false; }) const override = 0;
+
+    virtual void
+    adjacent_outgoing_nodes(node_index node,
+                            const std::function<void(node_index)> &callback) const override = 0;
+
+    virtual void
+    adjacent_incoming_nodes(node_index node,
```
karasikov, line 247: what is this class about?

karasikov (changes requested):

metagraph/src/graph/representation/base/dbg_wrapper.hpp
hmusta: For wrappers like `DBGSuccinctCached` which don't modify the external view of the graph and only keep an internal cache, allowing for non-const graphs to work has very little extra overhead.

metagraph/src/graph/representation/base/dbg_wrapper.hpp
hmusta: I'll add a comment. The idea here is that if the graph modifies the external view of the graph (adds/removes nodes), then these newly-made pure virtual methods are the bare minimum that need to be implemented. For the rest of the methods, this uses the `DeBruijnGraph` versions instead of the template type's by default.

karasikov: But how many classes will inherit from this? Isn't this needed only for canonical_dbg?

hmusta: `MaskedDeBruijnGraph` and `RCDBG` (#333) inherit from this

hmusta: while `DBGSuccinctCache` inherits from `DBGWrapper`

metagraph/src/graph/representation/base/dbg_wrapper.hpp
karasikov: Yes, but we aren't using this. So it's still better to remove this.

metagraph/src/graph/representation/base/dbg_wrapper.hpp
karasikov: why do you need to add/remove virtual nodes in RCDBG?

metagraph/src/graph/representation/base/dbg_wrapper.hpp
hmusta: The only place where it's used is in the unit tests for `DBGSuccinctCached` (since I add the cached version as another graph type)

metagraph/src/graph/representation/base/dbg_wrapper.hpp
karasikov: what is the point of this class, anyway? that you defined methods like
```
    virtual bool has_single_outgoing(node_index node) const override {
        return DeBruijnGraph::has_single_outgoing(node);
    }
```
?

But isn't that already defined the same way in DBGWrapper?

hmusta: The wrapper reverses the direction of all traversal and returns the reverse complement for its node labels, so its external view is different from the underlying graph (using the same node indices).

metagraph/src/graph/representation/base/dbg_wrapper.hpp
karasikov: Maybe just test it like all other static graphs?

metagraph/src/graph/representation/base/dbg_wrapper.hpp
hmusta: > what is the point of this class, anyway? that you defined methods like
> 
> ```
>     virtual bool has_single_outgoing(node_index node) const override {
>         return DeBruijnGraph::has_single_outgoing(node);
>     }
> ```
> 
> ?
> 
> But isn't that already defined the same way in DBGWrapper?

`DBGWrapper` calls all of its methods from the underlying graph, whereas `DBGNodeModifyingWrapper` tends to call the `DeBruijnGraph` versions

karasikov: This definitely sounds too complex, even if it makes sense.

Is it possible to do things without this class?

metagraph/src/graph/representation/base/dbg_wrapper.hpp
hmusta: We don't really have a framework for testing these other than copying and pasting unit tests, so it's not ideal.

hmusta: That can be saved for a later pull request.

karasikov: but how do we test DBGBitmap, for instance?

hmusta: It's still modifiable in the sense that `load` is implemented. So when adding unit tests for `DBGSuccinctCached`, I had to implement serialize and load. At that point, I figured why not also implement add_sequence since it's trivial

hmusta: So the graph pointer had to be mutable to implement load.

karasikov: I see, but we can add many things, though each of them makes the code more complex and harder to extend. So if those features are not used, it's better not to add them, even if it looks natural.

Do we actually call load on these wrappers? I thought we always load the graph and only then wrap it into something else.

hmusta: load and serialize is only used in the unit tests. One could add another set of graphs called StaticGraphs to test_dbg_helpers so that the load and serialize tests are only run for non-const graph types.

metagraph/src/graph/representation/base/dbg_wrapper.hpp
hmusta: The approaches I see are:
1. have each method run `graph_->myop(...)` and trust the child class to override everything relevant
2. have each method run `DeBruijnGraph::myop(...)` and trust the child class to override based on efficiency concerns (e.g., manually checking if the type is DBGSuccinct)
3. make every method which has non-negligible run time pure virtual
4. some combination of the above

karasikov: The child must override everything relevant in any case. I don't see how making these intermediate classes may help.

hmusta: Not exactly. In `CanonicalDBG`, for example, methods like `has_single_outgoing`, 'has_multiple_outgoing`, `traverse(node_index, const char*, ...)`, etc. can use the default implementations from `DeBruijnGraph`. With this structure, one doesn't have to manually override them.

hmusta: I cleaned up the two wrappers now and added descriptions, so it should be easier to understand now.

karasikov: Just always re-set those to `DeBruijnGraph::...`

hmusta: But if everything needs to be reset anyway, there's no point in having a parent wrapper class.

metagraph/src/graph/representation/base/dbg_wrapper.hpp
```diff
@@ -0,0 +1,253 @@
...
+        assert(!graph_ptr_);
+    }
+
+    // aliasing constructor
+    template <class InGraph>
+    explicit IDBGWrapper(const InGraph &graph)
+          : graph_(std::shared_ptr<const Graph>{}, dynamic_cast<const Graph*>(&graph)) {
+        assert(graph_);
+        assert(!graph_ptr_);
+    }
```
karasikov, line 38: move to `protected`

metagraph/src/graph/representation/base/dbg_wrapper.hpp
karasikov: we didn't have load/serialize in CanonicalDBG, but you still were able to test it. Why don't we keep things as they were?

metagraph/src/graph/representation/base/dbg_wrapper.hpp
karasikov: the point of the class is to avoid writing forwards like `return graph_.traverse(...);` in every wrapper.

This abstract parent delegator-wrapper should be a single thing. We don't need 2.

metagraph/src/graph/representation/base/dbg_wrapper.hpp
hmusta: Anyway, for this pull request we don't need a non-const pointer, and for #334 one can do the unit tests without needing to add a non-const pointer, so I just removed this feature...

metagraph/src/graph/representation/base/dbg_wrapper.hpp
hmusta: done

metagraph/src/graph/representation/base/dbg_wrapper.hpp
hmusta: Yes, but then sometimes you want to delegate to the base graph's method, and other times to `DeBruijnGraph`'s. If only one of them is implemented, then you need to spend time writing delegations for the other one.

karasikov: My point is, let's make this wrapped graph with overloaded methods behave exactly the same way as if this was overloaded for the base DeBruijnGraph. In this case, things will be natural and you don't need to check every second the code to double-check how things work. Because this is the ultimate goal of the interfaces -- they should assume the most natural and simple behavior you could expect.

karasikov: And even having the name `DBGNodeModifyingWrapper` clearly doesn't satisfy this condition.

hmusta: So I'm guessing you want it to look something like this? 0085837a56e0faaf22f40ab89408c1960c1bf7a6

The issue I have with this implementation is that there are so many methods in `DeBruijnGraph` that it's really easy to forget what needs to be overloaded and make mistakes.

With `DBGSuccinctCached`, for example, one would have to look through `sequence_graph.hpp` and find which methods have default implementations for `DeBruijnGraph` so they can be overloaded.

karasikov: Think of one of those wrappers: CanonicalDBG, etc.
The purpose of this abstract wrapper is to avoid forwarding the calls duplicated among all of them.

So just define those forwards in here, and keep the rest pure abstract, so they are defined in the children.

karasikov: But it might also be that this idea with writing common interface class for wrappers is not that good after all and just rewriting all methods in all wrappers is better and more straighforward.

hmusta: I think it's still useful to have an abstract wrapper parent to deal with stuff like `get_k`, `alphabet`, etc, and also constructors.

I've reduced the class down to a minimum so that anyone who uses it has to be mindful of what should be implemented in derived classes, while not having to deal with the trivial stuff that's implemented in the parent.

metagraph/src/graph/representation/base/dbg_wrapper.hpp
```diff
@@ -0,0 +1,96 @@
...
+     */
+    virtual const Graph& get_graph() const { return *graph_; }
+    virtual std::shared_ptr<const Graph> get_graph_ptr() const { return graph_; }
+
+    /**
+     * Methods shared by all wrappers
+     */
+    virtual std::string file_extension() const override final { return graph_->file_extension(); }
+    virtual size_t get_k() const override final { return graph_->get_k(); }
+    virtual const std::string& alphabet() const override final { return graph_->alphabet(); }
```
karasikov, line 53: let's define `get_mode()` too.

hmusta: ok

karasikov (approved):
Now I see that this class doesn't seem to add much. So I'd be totally ok with abandoning it and keeping things as they were, where we redefine all required methods in child wrappers.

But as you wish.

PR #342, opened 2021-07-27
support row-diff transform in memory without disk swap

karasikov:
Necessary if the number of columns is too large
and either buffers or temp files can't be created.

### DIFF
--- a/metagraph/src/annotation/row_diff_builder.cpp
+++ b/metagraph/src/annotation/row_diff_builder.cpp
@@ -796,7 +796,13 @@ void convert_batch_to_row_diff(const std::string &pred_succ_fprefix,
         }
     }
 
-    const fs::path tmp_path = utils::create_temp_dir(swap_dir, "col");
+    fs::path tmp_path;
+    bool swap_disk = !swap_dir.empty();
+    if (swap_disk) {
+        tmp_path = utils::create_temp_dir(swap_dir, "col");
+    } else {
+        logger->info("Diff-transform in memory without disk swap");
+    }
 
     // stores the row indices that were set because of differences to incoming/outgoing
     // edges, for each of the sources, per chunk. set_rows_fwd is already sorted
@@ -809,12 +815,67 @@ void convert_batch_to_row_diff(const std::string &pred_succ_fprefix,
     auto tmp_file = [&](size_t s, size_t j, size_t chunk) {
         return tmp_path/fmt::format("{}/col_{}_{}/chunk_{}", s / 100, s, j, chunk);
     };
-    auto dump_chunk_to_disk = [&](const std::vector<T> &v,
-                                  size_t s, size_t j, size_t chunk) {
-        assert(std::is_sorted(v.begin(), v.end()) && "all bits in chunks must be sorted");
+    auto dump_chunk_to_disk = [&](std::vector<T> &v, size_t s, size_t j, size_t chunk) {
+        if (!swap_disk)
+            return;
+        std::sort(v.begin(), v.end());
         Encoder<T>::append_block(v, tmp_file(s, j, chunk));
         row_diff_bits[s][j] += v.size();
+        v.resize(0);
     };
+
+    const uint32_t chunks_open_per_thread
+            = MAX_NUM_FILES_OPEN / std::max((uint32_t)1, num_threads) / (2 + with_values);
+    if (chunks_open_per_thread < 3) {
+        logger->error("Can't merge with less than 3 chunks per thread open. "
+                      "Max num files open: {}. Current number of threads: {}.",
+                      MAX_NUM_FILES_OPEN, num_threads);
+        exit(1);
+    }
+
+    using CallOnes = std::function<void(const std::function<void(uint64_t)> &)>;
+    std::function<CallOnes(size_t, size_t)> call_ones = [&](size_t s, size_t j) {
+        CallOnes caller = [&,s,j](const std::function<void(uint64_t)> &call) {
+            std::vector<std::string> filenames;
+            // skip chunk with fwd bits which have already been counted if stage 1
+            for (uint32_t chunk = compute_row_reduction ? 1 : 0;
+                                chunk < num_chunks[s][j]; ++chunk) {
+                filenames.push_back(tmp_file(s, j, chunk));
+            }
+
+            const bool remove_chunks = true;
+            uint64_t r = 0;
+            elias_fano::merge_files<T>(filenames, [&](T v) {
+                call(utils::get_first(v));
+                if constexpr(with_values) {
+                    assert(v.second && "zero diffs must have been skipped");
+                    values[s][j][r++] = matrix::encode_diff(v.second);
+                }
+            }, remove_chunks, chunks_open_per_thread);
+        };
+        return caller;
+    };
+    if (!swap_disk) {
+        call_ones = [&](size_t s, size_t j) {
+            CallOnes caller = [&,s,j](const std::function<void(uint64_t)> &call) {
+                auto &v = set_rows_fwd[s][j];
+                v.insert(v.end(), set_rows_bwd[s][j].begin(), set_rows_bwd[s][j].end());
+                set_rows_bwd[s][j] = {};
+
+                std::sort(v.begin(), v.end());
+
+                uint64_t r = 0;
+                for (size_t i = 0; i < v.size(); ++i) {
+                    call(utils::get_first(v[i]));
+                    if constexpr(with_values) {
+                        assert(v[i].second && "zero diffs must have been skipped");
+                        values[s][j][r++] = matrix::encode_diff(v[i].second);
+                    }
+                }
+            };
+            return caller;
+        };
+    }
 
     // In the first stage, only one buffer is created per column (`bwd`).
     // In the last stage, two buffers (`fwd` and `bwd`) are created per column.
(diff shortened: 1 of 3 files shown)


metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -809,12 +815,67 @@ void convert_batch_to_row_diff(const std::string &pred_succ_fprefix,
...
+                    values[s][j][r++] = matrix::encode_diff(v.second);
+                }
+            }, remove_chunks, chunks_open_per_thread);
+        };
+        return caller;
     };
+    if (!swap_disk) {
+        call_ones = [&](size_t s, size_t j) {
+            CallOnes caller = [&,s,j](const std::function<void(uint64_t)> &call) {
+                auto &v = set_rows_fwd[s][j];
```
hmusta, line 881: it would be nice if this was named something like `set_rows_fwd_s_j` so it would be easier to find how `set_rows_fwd` is used throughout this function

metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -999,8 +1059,10 @@ void convert_batch_to_row_diff(const std::string &pred_succ_fprefix,
 
     async_writer.join();
 
-    set_rows_fwd.clear(); // free up memory
-    set_rows_bwd.clear();
+    if (swap_disk) {
+        set_rows_fwd.clear(); // free up memory
+        set_rows_bwd.clear();
+    }
```
hmusta, line 1084: Unless I missed something, it looks like `set_rows_fwd` and `set_rows_bwd` are not used after this anyway, so why not clear them in all cases?

hmusta: Also, I'm struggling to see what these vectors actually do when in memory mode. It looks like you populate them and do nothing...

hmusta: it would be nice to have some comments describing the general flow to better understand this

hmusta (changes requested):

metagraph/src/annotation/row_diff_builder.cpp
hmusta: same applies for `set_rows_bwd`

karasikov: too long, this block is too technical anyway

metagraph/src/annotation/row_diff_builder.cpp
karasikov: For in-memory transforms, these buffers store all the diffs, they are extracted in `call_diffs` (invoked below).
I added a comment.

metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -809,13 +811,91 @@ void convert_batch_to_row_diff(const std::string &pred_succ_fprefix,
...
+
+                std::sort(v.begin(), v.end());
+
+                uint64_t r = 0;
+                for (size_t i = 0; i < v.size(); ++i) {
+                    call(utils::get_first(v[i]));
+                    if constexpr(with_values) {
+                        assert(v[i].second && "zero diffs must have been skipped");
+                        values[s][j][r++] = matrix::encode_diff(v[i].second);
+                    }
```
hmusta, line 893: you can probably put this block in a lambda so it can be used here and above

metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -16,6 +16,7 @@ const uint64_t BLOCK_SIZE = 1 << 25;
 const uint64_t BUFFER_SIZE = 1024 * 1024; // 1 MiB
 const uint64_t ROW_REDUCTION_WIDTH = 32;
 const uint32_t MAX_NUM_FILES_OPEN = 2000;
+const uint64_t MAX_COLUMNS_IN_BATCH = 1'000'000;
```
hmusta, line 19: It would be nice to have this exposed in the CLI. That way, the users can have the option of keeping the thread count fixed and playing around with this accordingly.

hmusta: or ideally, having the batch size set automatically

metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -16,6 +16,7 @@ const uint64_t BLOCK_SIZE = 1 << 25;
 const uint64_t BUFFER_SIZE = 1024 * 1024; // 1 MiB
 const uint64_t ROW_REDUCTION_WIDTH = 32;
 const uint32_t MAX_NUM_FILES_OPEN = 2000;
```
hmusta, line 18: is there a way to read the `ulimit` from the system so this doesn't have to be hard-coded?

hmusta (changes requested):

metagraph/src/annotation/row_diff_builder.cpp
hmusta: the default ulimit is 1024, so this could cause problems if you eventually open 2000 files at once

metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -809,13 +811,91 @@ void convert_batch_to_row_diff(const std::string &pred_succ_fprefix,
...
+        }
+
+        if (total_num_labels > MAX_COLUMNS_IN_BATCH / 2) {
+            logger->warn("The number of columns to transform is large: {}."
+                         " Consider disabling disk swap (pass --disk-swap \"\").",
+                         total_num_labels);
+        }
+        if (total_num_labels > MAX_COLUMNS_IN_BATCH) {
+            logger->error("Too many columns to transform with disk swap: {} (> MAX {}).",
+                          total_num_labels, MAX_COLUMNS_IN_BATCH);
```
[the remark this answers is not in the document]
karasikov: done

metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -809,13 +811,91 @@ void convert_batch_to_row_diff(const std::string &pred_succ_fprefix,
...
+            exit(1);
+        }
+        tmp_path = utils::create_temp_dir(swap_dir, "col");
+
+        const uint32_t chunks_open_per_thread
+                = MAX_NUM_FILES_OPEN / std::max((uint32_t)1, num_threads) / (2 + with_values);
+        if (chunks_open_per_thread < 3) {
+            logger->error("Can't merge with less than 3 chunks per thread open. "
```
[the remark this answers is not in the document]
karasikov: No, this is unrelated to that new card.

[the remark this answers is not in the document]
karasikov: extended the comment

metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -999,8 +1078,13 @@ void convert_batch_to_row_diff(const std::string &pred_succ_fprefix,
...
     async_writer.join();
 
-    set_rows_fwd.clear(); // free up memory
-    set_rows_bwd.clear();
+    // For transforms with disk swap, the diffs are dumped to temp chunks, so the
+    // fwd and bwd buffers can be removed. Otherwise, the diffs are stored there
+    // and hence must be kept (they are extracted in `call_diffs` invoked below).
+    if (swap_disk) {
+        set_rows_fwd.clear(); // free up memory
+        set_rows_bwd.clear();
```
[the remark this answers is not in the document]
karasikov: it's a vector of vectors, so everything is cleaned up

metagraph/src/annotation/row_diff_builder.cpp
karasikov: it would be far above, I think it's better this way

metagraph/src/annotation/row_diff_builder.cpp
karasikov: it's a different thing, not a batch size.

I just set this guy so that if there are too many tiny columns, we don't want to create millions of temp files, even though all of them would be small.

There are 4 different things you might be confusing:
1. trade-offs with threads when annotating (what you saw in the new trello card)
2. batch size, already set automatically -- we load as many annotation columns as we can fit in RAM.
3. number of open file descriptors (pre-merge chunks if there are too many of them, but throw error if we can't do the pre-merging in parallel, i.e. the number of threads is roughly 1000 or above.)
4. the number of temp files we'd be ok to create. For annotations with 50M columns, transforming them with disk-swap is hard because we can't keep 50M folders with chunks, that's inefficient, so I print warnings and errors if that happens.

metagraph/src/annotation/row_diff_builder.cpp
karasikov: I thought the default is 4096.

hmusta: one can increase the limit to 4096 without root privileges (unless someone changes this), but the default base value is 1024

karasikov: OK, it's 4k on the cluster. I wouldn't like to change this, it has worked so far.

We can add this parameter when someone faces the issue.

hmusta: Ahh, I see where the confusion is. On our login nodes, it's 1024, but on the compute nodes, it's 4096...

hmusta (approved):

PR #343, opened 2021-07-28
Auxiliary methods for taxonomic classification

heracle:
Add some small methods that will be further used for tax class

Signed-off-by: Radu Muntean <radu.muntean.2@gmail.com>

### DIFF
--- a/metagraph/src/common/utils/string_utils.cpp
+++ b/metagraph/src/common/utils/string_utils.cpp
@@ -7,6 +7,13 @@
 
 namespace utils {
 
+bool starts_with(const std::string &str, const std::string &prefix) {
+    if (prefix.size() > str.size()) {
+        return false;
+    }
+    return prefix == str.substr(0, static_cast<int>(prefix.size()));
+}
+
 bool ends_with(const std::string &str, const std::string &suffix) {
     auto actual_suffix = str.substr(
         std::max(0, static_cast<int>(str.size())
--- a/metagraph/src/common/utils/string_utils.cpp
+++ b/metagraph/src/common/utils/string_utils.cpp
@@ -11,7 +11,7 @@ bool starts_with(const std::string &str, const std::string &prefix) {
     if (prefix.size() > str.size()) {
         return false;
     }
-    return prefix == str.substr(0, static_cast<int>(prefix.size()));
+    return prefix == std::string_view(str).substr(0, prefix.size());
 }
 
 bool ends_with(const std::string &str, const std::string &suffix) {
--- a/metagraph/src/graph/annotated_dbg.cpp
+++ b/metagraph/src/graph/annotated_dbg.cpp
@@ -776,6 +776,34 @@ void AnnotatedSequenceGraph
     );
 }
 
+void AnnotatedDBG::call_annotated_rows(const std::vector<node_index> &rows,
+                                       std::function<void(const std::string&)> callback_cell,
+                                       std::function<void()> callback_row) const {
+    assert(check_compatibility());
+
+    auto unique_matrix_rows = annotator_->get_matrix().get_rows(rows);
+
+    //TODO make sure that this function works even if we have duplications in 'rows'. Then, delete this error catch.
+    if (rows.size() != unique_matrix_rows.size()) {
+        throw std::runtime_error("The current 'call_annotated_rows' call contains duplication.");
+    }
+
+    if (unique_matrix_rows.size() >= std::numeric_limits<uint32_t>::max()) {
+        throw std::runtime_error("The current 'call_annotated_rows' call has returned, "
+                                  + std::to_string(unique_matrix_rows.size())  +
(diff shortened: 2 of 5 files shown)


metagraph/src/graph/annotated_dbg.cpp
```diff
@@ -776,6 +776,34 @@ ::call_annotated_nodes(const Label &label,
...
+    if (rows.size() != unique_matrix_rows.size()) {
+        throw std::runtime_error("The current 'call_annotated_rows' call contains duplication.");
+    }
+
+    if (unique_matrix_rows.size() >= std::numeric_limits<uint32_t>::max()) {
+        throw std::runtime_error(
+                folly::to<string>("The current 'call_annotated_rows' call has returned, ", unique_matrix_rows.size(),
+                                  "rows. The maximum number of rows that can be returned is ",
+                                  std::numeric_limits<uint32_t>::max(),
+                                  ". Please reduce the query batch size"));
```
karasikov, line 796: use fmt::format

metagraph/src/common/utils/string_utils.cpp
```diff
@@ -7,6 +7,13 @@
 
 namespace utils {
 
+bool starts_with(const std::string &str, const std::string &prefix) {
+    if (prefix.size() > str.size()) {
+        return false;
+    }
+    return prefix == str.substr(0, static_cast<int>(prefix.size()));
```
karasikov, line 14:
```suggestion
    return prefix == std::string_view(str).substr(0, prefix.size());
```

metagraph/src/graph/annotated_dbg.cpp
```diff
@@ -776,6 +776,34 @@ ::call_annotated_nodes(const Label &label,
...
+    }
+    const auto &label_encoder = annotator_->get_label_encoder();
+    for (auto row : unique_matrix_rows) {
+        for (auto cell : row) {
+            callback_cell(label_encoder.decode(cell));
+        }
+        callback_row();
+    }
+}
```
karasikov, line 805: I don't quite see a reason to add this function. You have access to all members, so can implement this directly where needed.

karasikov (changes requested):

metagraph/src/graph/annotated_dbg.cpp
```diff
@@ -776,6 +776,34 @@ ::call_annotated_nodes(const Label &label,
...
+
+    if (unique_matrix_rows.size() >= std::numeric_limits<uint32_t>::max()) {
+        throw std::runtime_error(
+                folly::to<string>("The current 'call_annotated_rows' call has returned, ", unique_matrix_rows.size(),
+                                  "rows. The maximum number of rows that can be returned is ",
+                                  std::numeric_limits<uint32_t>::max(),
+                                  ". Please reduce the query batch size"));
+    }
+    const auto &label_encoder = annotator_->get_label_encoder();
+    for (auto row : unique_matrix_rows) {
```
hmusta, line 799: this is unnecessarily making a copy
```suggestion
    for (const auto &row : unique_matrix_rows) {
```

metagraph/src/graph/annotated_dbg.hpp
```diff
@@ -156,6 +156,10 @@ class AnnotatedDBG : public AnnotatedSequenceGraph {
                                      int32_t match_score = 1,
                                      int32_t mismatch_score = 2) const;
 
+    void call_annotated_rows(const std::vector<node_index> &rows,
+                             std::function<void(const std::string&)> callback_cell,
+                             std::function<void()> callback_row) const;
```
hmusta, line 161: What's the second callback supposed to be used for? It's not clear...

karasikov: I suggest to remove this function

heracle: Thanks for the comments! I need to have access to  `annotator_` field and I can't implement it directly where needed.

heracle: It is using 2 callback functions: one for doing a preprocessing per cell (i.e. get the taxid from label) and one for a final computation on an annotated row (i.e. get the LCA taxid). If you think that it would be more clear, I can reimplement it to use only one callback that receives the list of labels on one row (but this will use more memory -> store list of labels instead of list of integers)

hmusta: But do you need to have a non-const reference to the annotator? If not, you can just call `get_annotator().get_matrix()`

metagraph/src/graph/annotated_dbg.cpp
heracle: thanks! answered in the other comment

metagraph/src/graph/annotated_dbg.hpp
heracle: Cool, I missed that `get_annotator()`. Did the changes now, thanks for the feedback

PR #346, opened 2021-07-30
taxonomic classification skeleton with the RMQ preprocessing 

heracle:
This PR draws the entire structure of the taxonomic classification approach (without CLI cmds):

The header script `tax_classifier.hpp` contains the description of all the taxonomic classification classes:

- `TaxonomyBase` -> implements the methods and fields that will be used by both classification options (via anno matrix or via taxDB)
- `TaxonomyClsAnno` -> implements the methods for `via anno matrix` option
- `TaxonomyClsImportDB` -> implements the methods for `via taxDB` option

In this PR there are fully implemented the following methods that are used for the preprocessing steps:

- label to taxid parsing (which is different for GenBank or Kraken2 sequence labels)
- parsing the taxonomic tree file
- parsing the accession version to taxid lookup table file
- DFS and RMQ precomputation


heracle:
Please let me know if the structure of the classes and the files looks ok:

- At this moment there is one class `TaxonomyBase` with the methods that will be used by both classification options. Then `TaxonomyClsImportDB` and `TaxonomyClsAnno` do a public inheritance.
- There is a file `label_to_taxid.cpp` where there are implemented all the methods regarding the label parsing. I created a separate file just for the reason to be able to extend the `enum LabelType` by changing one file only

metagraph/src/annotation/taxonomy/tax_classifier.hpp
```diff
@@ -0,0 +1,191 @@
...
+     * @param [input] tax_tree_filepath ->  path to a taxonomic tree ("nodes.dmp" file).
+     * @param [input] label_taxid_map_filepath ->  path to acccession version to taxid lookup table (".accession2taxid" file). Mandatory if the label doesn't contain the 'taxid'.
+     */
+    TaxonomyClsAnno(const graph::AnnotatedDBG &anno,
+                    const double lca_coverage_rate,
+                    const double kmers_discovery_rate,
+                    const std::string &tax_tree_filepath,
+                    const std::string &label_taxid_map_filepath = "");
+    TaxonomyClsAnno() {};
+    ~TaxonomyClsAnno() {};
```
hmusta, line 122: there's no need to define this, it's already implicitly defined

metagraph/src/annotation/taxonomy/tax_classifier.hpp
```diff
@@ -0,0 +1,191 @@
...
+     * fast_log2 is a table for a fast compute of log2(x).
+     */
+    std::vector<uint32_t> fast_log2;
+
+    /**
+     * fast_pow2 is a table for a fast compute of pow2(x).
+     */
+    std::vector<uint32_t> fast_pow2;
+
+    const graph::AnnotatedDBG *_anno_matrix = NULL;
```
hmusta, line 185: do you need this class to be copy assignable (`operator=`)? if not, then it's cleaner to make this a const reference instead

metagraph/src/annotation/taxonomy/tax_classifier.hpp
```diff
@@ -0,0 +1,191 @@
...
+using TaxId = std::uint32_t;
+using ChildrenList = tsl::hopscotch_map<TaxId, std::vector<TaxId>>;
+
+class TaxonomyBase {
+public:
+    using KmerId = annot::MultiLabelEncoded<std::string>::Index;
+    using node_index = graph::SequenceGraph::node_index;
+
+    enum LabelType {
+        UNASSIGNED, GEN_BANK, KRAKEN
```
hmusta, line 29: what do these label types represent? what is a Kraken label here?

metagraph/src/annotation/taxonomy/tax_classifier.cpp
```diff
@@ -0,0 +1,199 @@
...
+        std::vector<std::string> parts = utils::split_string(line, "\t");
+        if (parts.size() <= 2) {
+            logger->error("The Taxonomic tree filepath contains incomplete lines. Please make sure that this file was not manually modified: {}.",
+                          tax_tree_filepath);
+            exit(1);
+        }
+        uint32_t act = static_cast<uint32_t>(std::stoull(parts[0]));
+        uint32_t parent = static_cast<uint32_t>(std::stoull(parts[2]));
+        full_parents_list[act] = parent;
+        this->node_parent[act] = parent;
```
hmusta, line 83: I don't think you need to use `this->` for these unless this member is stored in the parent class and the parent is a templated class

metagraph/src/annotation/taxonomy/tax_classifier.hpp
```diff
@@ -0,0 +1,191 @@
...
+
+    /**
+     * fast_log2 is a table for a fast compute of log2(x).
+     */
+    std::vector<uint32_t> fast_log2;
+
+    /**
+     * fast_pow2 is a table for a fast compute of pow2(x).
+     */
+    std::vector<uint32_t> fast_pow2;
```
hmusta, line 183: is this really faster than `1llu << x`?

metagraph/src/annotation/taxonomy/tax_classifier.hpp
```diff
@@ -0,0 +1,191 @@
...
+    /**
+     * node_to_linearization_idx[node] returns the index of the first occurrence of node
+     * in the tree linearization order. This array will be further used inside a RMQ query.
+     */
+    tsl::hopscotch_map<TaxId, uint32_t> node_to_linearization_idx;
+
+    /**
+     * fast_log2 is a table for a fast compute of log2(x).
+     */
+    std::vector<uint32_t> fast_log2;
```
hmusta, line 178: it'll probably be faster to do
```c++
#define LOG2(X) ((unsigned) (8*sizeof (unsigned long long) - __builtin_clzll((X)) - 1))
```
source: https://stackoverflow.com/questions/11376288/fast-computing-of-log2-for-64-bit-integers

hmusta (changes requested):

metagraph/src/annotation/taxonomy/tax_classifier.hpp
hmusta: ah, after reading above, it looks like `KRAKEN`-type labels are taxonomic IDs. So we should call this `TAXID` instead of `KRAKEN`

metagraph/src/annotation/taxonomy/tax_classifier.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+using mtg::common::logger;
+
+TaxonomyClsAnno::TaxonomyClsAnno(const graph::AnnotatedDBG &anno,
+                                 const double lca_coverage_rate,
+                                 const double kmers_discovery_rate,
+                                 const std::string &tax_tree_filepath,
+                                 const std::string &label_taxid_map_filepath) : _anno_matrix(&anno) {
+    _lca_coverage_rate = lca_coverage_rate;
+    _kmers_discovery_rate = kmers_discovery_rate;
```
hmusta, line 23:
```suggestion
                                 const std::string &label_taxid_map_filepath)
          : _anno_matrix(&anno),
            _lca_coverage_rate(lca_coverage_rate),
            _kmers_discovery_rate(kmers_discovery_rate) {
```

metagraph/src/annotation/taxonomy/tax_classifier.cpp
```diff
@@ -0,0 +1,199 @@
...
+        this->fast_log2[i] = 1 + this->fast_log2[i/2];
+        if (this->fast_log2[i] > this->fast_log2[i-1]) {
+            this->fast_pow2.push_back(i);
+        }
+    }
+}
+
+TaxId TaxonomyClsAnno::assign_class(const std::string &sequence) const {
+    std::cerr << "assign class not implemented " << sequence << "\n\n";
+    return 0;
```
hmusta, line 195: This'll be hard to catch if people are redirecting stderr to some other place.

Use `throw std::runtime_error` instead

metagraph/src/annotation/taxonomy/tax_classifier.hpp
heracle: Thanks for the review! This `enum` contains a list of labels formats where we need to hardcode how to get the accession version or the taxid.

For example:

- one GEN_BANK label looks like this: ">gi|1070643132|ref|**NC_031224.1**| Arthrobacter phage Mudcat, complete genome"
- one Kraken2 label looks like: ">kraken:taxid|2016032|**NC_047834.1** Alteromonas virus vB_AspP-H4/4, complete genome"

Maybe there are better ideas than hardcoding a "get_accession_version" method for all these formats, but I didn't see any other way around this

metagraph/src/annotation/taxonomy/tax_classifier.cpp
heracle: I tried to implement this, but `_lca_coverage_rate` and `_kmers_discovery_rate` are inherited base class fields and got an error. Should I create a constructor for `TaxonomyBase` to initialize these two `_lca_coverage_rate` and `_kmers_discovery_rate`?

metagraph/src/annotation/taxonomy/tax_classifier.hpp
heracle: you are right, this vector is a useless complication. Better to use `<<` instead and the same for `fast_log2`, thanks!

metagraph/src/annotation/taxonomy/tax_classifier.cpp
hmusta: If they're stored in the base class, then it's best to have a constructor for the base, then call the base's constructor here before setting `_anno_matrix`

metagraph/src/annotation/taxonomy/tax_classifier.hpp
heracle: This  class has an empty constructor `TaxonomyClsAnno() {};` that is used for all the unit tests. If i make the `_anno_matrix` a const reference, then I get a compilation error because I am not assigning it inside this empty class constructor. Is there any other better way to implement it?

metagraph/src/annotation/taxonomy/tax_classifier.hpp
karasikov: Better call `sdsl::bits::hi(x)`

metagraph/src/annotation/taxonomy/label_to_taxid.cpp
```diff
@@ -0,0 +1,95 @@
...
+        *require_accversion_to_taxid_map = true;
+    } else if (utils::starts_with(utils::split_string(label, ":")[1], "taxid|")) {
+        // e.g.   >kraken:taxid|2016032|NC_047834.1 Alteromonas virus vB_AspP-H4/4, complete genome
+        label_type = TAXID;
+        *require_accversion_to_taxid_map = false;
+    } else {
+        logger->error("Can't determine the type of the given label {}. Please make sure that the labels are in a recognized format.", label);
+        std::exit(1);
+    }
+}
```
karasikov, line 25: return bool instead of passing a pointer

metagraph/src/annotation/taxonomy/label_to_taxid.cpp
```diff
@@ -0,0 +1,95 @@
...
+    if (utils::starts_with(label, "gi|")) {
+        // e.g.   >gi|1070643132|ref|NC_031224.1| Arthrobacter phage Mudcat, complete genome
+        label_type = GEN_BANK;
+        *require_accversion_to_taxid_map = true;
+    } else if (utils::starts_with(utils::split_string(label, ":")[1], "taxid|")) {
+        // e.g.   >kraken:taxid|2016032|NC_047834.1 Alteromonas virus vB_AspP-H4/4, complete genome
+        label_type = TAXID;
+        *require_accversion_to_taxid_map = false;
+    } else {
+        logger->error("Can't determine the type of the given label {}. Please make sure that the labels are in a recognized format.", label);
```
karasikov, line 22: break the line
```suggestion
        logger->error("Can't determine the type of the given label {}."
                      " Make sure the labels are in the correct format.", label);
```

metagraph/src/annotation/taxonomy/label_to_taxid.cpp
```diff
@@ -0,0 +1,95 @@
...
+        std::string acc_version = get_accession_version_from_label(label);
+        if (not accversion_to_taxid_map.count(acc_version)) {
+            return false;
+        }
+        *taxid = accversion_to_taxid_map.at(acc_version);
+        return true;
+    }
+
+    logger->error("Run get_taxid_from_label() for unknown label {}.", label);
+    std::exit(1);
```
karasikov, line 41: `get_taxid_from_label()` is not exposed in the CLI, so printing this error message to the user will be strange.

Ether print something that is exposed to the CLI or make this an assert (in case this can only happen if there is a bug in the code or the flow)

metagraph/src/annotation/taxonomy/label_to_taxid.cpp
```diff
@@ -0,0 +1,95 @@
...
+            logger->error("The accession to taxid map table contains empty lines. Please make sure that this file was not manually modified {}.", filepath);
+            exit(1);
+        }
+        std::vector<std::string> parts = utils::split_string(line, "\t");
+        if (parts.size() <= 2) {
+            logger->error("The accession to taxid map table contains incomplete lines. Please make sure that this file was not manually modified {}.", filepath);
+            exit(1);
+        }
+        if (input_accessions.size() == 0 || input_accessions.count(parts[1])) {
+            accversion_to_taxid_map[parts[1]] = static_cast<TaxId>(std::stoull(parts[2]));
```
karasikov, line 89: this will be casted to TaxId automatically
```suggestion
            accversion_to_taxid_map[parts[1]] = std::stoull(parts[2]);
```

metagraph/tests/annotation/taxonomy/test_taxonomy.cpp
```diff
@@ -0,0 +1,91 @@
...
+        {0, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0},
+        {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0}
+    };
+
+    tax->rmq_preprocessing(linearization);
+    EXPECT_EQ(expected_rmq, tax->rmq_data);
+}
+
+}
+}
```
karasikov, line 91: Write which namespace is closed to make it easier to follow.

```suggestion
} // namespace test
} // namespace mtg
```

metagraph/tests/annotation/taxonomy/test_taxonomy.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+#define TESTING
+
+#include <tsl/hopscotch_map.h>
+#include <tsl/hopscotch_set.h>
+
+#include "annotation/taxonomy/tax_classifier.hpp"
+
+namespace mtg {
+namespace test {
```
karasikov, line 11: don't we use unnamed namespace for tests?
```suggestion
namespace {
```

metagraph/src/annotation/taxonomy/tax_classifier.hpp
```diff
@@ -0,0 +1,194 @@
...
+    tsl::hopscotch_map<TaxId, TaxId> node_parent;
+
+    tsl::hopscotch_map<std::string, TaxId> accversion_to_taxid_map;
+
+    double _lca_coverage_rate;
+    double _kmers_discovery_rate;
+};
+
+class TaxonomyClsImportDB : public TaxonomyBase {
+public:
```
karasikov, line 105:
```suggestion
  public:
```

metagraph/src/annotation/taxonomy/tax_classifier.hpp
```diff
@@ -0,0 +1,194 @@
...
+    // todo implement
+    TaxonomyClsImportDB(const std::string &taxdb_filepath,
+                        const double lca_coverage_rate,
+                        const double kmers_discovery_rate);
+
+PRIVATE_TESTABLE:
```
karasikov, line 111: We write `private`, but then add `#define private public` in tests to access private methods. Check out how it's done in other tests.

```suggestion
  private:
```

metagraph/src/annotation/taxonomy/tax_classifier.cpp
```diff
@@ -0,0 +1,205 @@
...
+namespace annot {
+
+using mtg::common::logger;
+
+TaxonomyClsAnno::TaxonomyClsAnno(const graph::AnnotatedDBG &anno,
+                                 const double lca_coverage_rate,
+                                 const double kmers_discovery_rate,
+                                 const std::string &tax_tree_filepath,
+                                 const std::string &label_taxid_map_filepath) :
+                                 TaxonomyBase(lca_coverage_rate, kmers_discovery_rate), _anno_matrix(&anno) {
```
karasikov, line 22: Break lines to make it easier to read and see logical blocks rather than look nice and aligned.
```suggestion
                                 const std::string &label_taxid_map_filepath)
      : TaxonomyBase(lca_coverage_rate, kmers_discovery_rate),
        _anno_matrix(&anno) {
```

metagraph/src/annotation/taxonomy/tax_classifier.cpp
```diff
@@ -0,0 +1,205 @@
...
+    if (!std::filesystem::exists(tax_tree_filepath)) {
+        logger->error("Can't open taxonomic tree file {}.", tax_tree_filepath);
```
karasikov, line 24:
```suggestion
        logger->error("Can't open taxonomic tree file {}", tax_tree_filepath);
```

metagraph/src/annotation/taxonomy/tax_classifier.cpp
```diff
@@ -0,0 +1,205 @@
...
+        std::exit(1);
+    }
+
+    bool require_accversion_to_taxid_map = false;
+    assign_label_type(_anno_matrix->get_annotation().get_all_labels()[0], &require_accversion_to_taxid_map);
+
+    Timer timer;
+    if (require_accversion_to_taxid_map) {
+        logger->trace("Parsing label_taxid_map file..");
```
karasikov, line 33:
```suggestion
        logger->trace("Parsing label_taxid_map file...");
```

metagraph/src/annotation/taxonomy/tax_classifier.cpp
```diff
@@ -0,0 +1,205 @@
...
+        read_accversion_to_taxid_map(label_taxid_map_filepath, _anno_matrix);
+        logger->trace("Finished label_taxid_map file in {}s", timer.elapsed());
+    }
+
+    timer.reset();
+    logger->trace("Parsing taxonomic tree..");
+    ChildrenList tree;
+    read_tree(tax_tree_filepath, &tree);
+    logger->trace("Finished taxonomic tree read in {}s.", timer.elapsed());
```
karasikov, line 42:
```suggestion
    logger->trace("Finished taxonomic tree read in {} sec", timer.elapsed());
```

karasikov (changes requested):

metagraph/tests/annotation/taxonomy/test_taxonomy.cpp
heracle: replaced with unnamed namespace

metagraph/src/annotation/taxonomy/label_to_taxid.cpp
heracle: replaced, thanks!

metagraph/src/annotation/taxonomy/label_to_taxid.cpp
heracle: I was returning a `bool` in the previous implementation version, but thought that it could be misunderstood as a valid/failed assignation. Also, it could be a possibility to return 2 different boolean values in the future, depending on what precomputation we need to manage for different label types. 

But I can return a `bool` if you consider that it is more clear, just let me know.

heracle:
Thank you very much for the review!

metagraph/src/annotation/taxonomy/label_to_taxid.cpp
karasikov: The meaning of `require_accversion_to_taxid_map` is not clear at all either.

Better return a bool and either write a short comment explaining what is returned or consider renaming the method so it's self-explanatory.

karasikov (commented):
Please close resolved comments and also do a pass to fix all similar cases.

metagraph/src/annotation/taxonomy/label_to_taxid.cpp
heracle: I have changed as you suggested, thanks for the review! For the moment, I will let this comment as not solved in case of any follow-ups.

heracle:
Cool. Please let me know about anything not proper. Usually, I am trying my best to pass through the entire PR and fix all the similar cases, but for some of them, maybe I am not generalizing too well yet

metagraph/src/annotation/taxonomy/tax_classifier.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+TaxId TaxonomyClsAnno::find_lca(const std::vector<TaxId> &taxids) const {
+    throw std::runtime_error("find_lca TaxonomyClsAnno not implemented. Received taxids size"
+                             + to_string(taxids.size()));
+}
+
+TaxId TaxonomyClsImportDB::find_lca(const std::vector<TaxId> &taxids) const {
+    throw std::runtime_error("find_lca TaxonomyClsImportDB not implemented. Received taxids size"
+                             + to_string(taxids.size()));
+}
```
heracle, line 291: these will be implemented in the next PR. The error messages are not very suggestive, but I had to avoid the "unused variable" cpp error

metagraph/tests/annotation/taxonomy/test_taxonomy.cpp
```diff
@@ -0,0 +1,90 @@
...
+            {3, 2},
+            {4, 2},
+            {5, 1},
+            {6, 1},
+            {7, 1},
+            {8, 1},
+    };
+
+    std::vector<uint32_t> linearization = {
+            0, 1, 4, 7, 4, 8, 4, 1, 5, 1, 0, 2, 0, 3, 6, 3, 0
```
karasikov, line 76:
```suggestion
        0, 1, 4, 7, 4, 8, 4, 1, 5, 1, 0, 2, 0, 3, 6, 3, 0
```

metagraph/tests/annotation/taxonomy/test_taxonomy.cpp
```diff
@@ -0,0 +1,90 @@
...
+            {0, 4},
+            {1, 3},
+            {2, 1},
+            {3, 2},
+            {4, 2},
+            {5, 1},
+            {6, 1},
+            {7, 1},
+            {8, 1},
+    };
```
karasikov, line 73: 4 spaces

metagraph/tests/annotation/taxonomy/test_taxonomy.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+    std::vector<uint32_t> tree_linearization;
+    tax->dfs_statistics(0, tree, &tree_linearization);
+    EXPECT_EQ(expected_linearization, tree_linearization);
+    EXPECT_EQ(expected_node_depths, tax->node_depth);
+    EXPECT_EQ(expected_node_to_linearization_idx, tax->node_to_linearization_idx);
+}
+
+TEST (TaxonomyTest, ClsAnno_RmqPreprocessing) {
+    mtg::annot::TaxonomyClsAnno *tax = new mtg::annot::TaxonomyClsAnno();
```
karasikov, line 61: Never deleted -> Memory leak. Better use unique pointers

metagraph/tests/annotation/taxonomy/test_taxonomy.cpp
```diff
@@ -0,0 +1,90 @@
...
+    };
+
+    std::vector<uint32_t> tree_linearization;
+    tax->dfs_statistics(0, tree, &tree_linearization);
+    EXPECT_EQ(expected_linearization, tree_linearization);
+    EXPECT_EQ(expected_node_depths, tax->node_depth);
+    EXPECT_EQ(expected_node_to_linearization_idx, tax->node_to_linearization_idx);
+}
+
+TEST (TaxonomyTest, ClsAnno_RmqPreprocessing) {
```
karasikov, line 60:
```suggestion
TEST(TaxonomyTest, ClsAnno_RmqPreprocessing) {
```
same everywhere

metagraph/tests/annotation/taxonomy/test_taxonomy.cpp
```diff
@@ -0,0 +1,90 @@
...
+                {0, {1, 2, 3}},      // node 0 -> root
+                {1, {4, 5}},         // node 1
+                {2, {}},             // node 2
+                {3, {6}},            // node 3
+                {4, {7, 8}},         // node 4
+                {5, {}},
+                {6, {}},
+                {7, {}},
+                {8, {}},
+    };
```
karasikov, line 25: 4 spaces

metagraph/src/annotation/taxonomy/tax_classifier.hpp
```diff
@@ -0,0 +1,192 @@
...
+class TaxonomyClsAnno : public TaxonomyBase {
+  public:
+    /**
+     * TaxonomyCls constructor
+     *
+     * @param [input] anno -> the annotation matrix
+     * @param [input] lca_coverage_rate -> threshold used for taxonomic classification.
+     * @param [input] kmers_discovery_rate -> threshold used for taxonomic classification.
+     * @param [input] tax_tree_filepath ->  path to a taxonomic tree ("nodes.dmp" file).
+     * @param [input] label_taxid_map_filepath ->  path to acccession version to taxid lookup table (".accession2taxid" file). Mandatory if the label doesn't contain the 'taxid'.
```
karasikov, line 123: the line is too long

metagraph/src/annotation/taxonomy/tax_classifier.hpp
karasikov: private members have underscore in the end and not in the beginning
```suggestion
    const graph::AnnotatedDBG *anno_matrix_ = NULL;
```

karasikov: Why do you need this as a member, anyway? It's used only in that constructor.

metagraph/src/annotation/taxonomy/tax_classifier.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+void TaxonomyClsAnno::dfs_statistics(const TaxId node,
+                                     const ChildrenList &tree,
+                                     std::vector<TaxId> *tree_linearization) {
+    node_to_linearization_idx[node] = tree_linearization->size();
+    tree_linearization->push_back(node);
+    uint32_t depth = 0;
+    if (tree.count(node)) {
+        for (const TaxId &child : tree.at(node)) {
```
karasikov, line 230: Double lookup (`count` + `at`), which could be avoided with `tree.find(node)`

metagraph/src/annotation/taxonomy/tax_classifier.cpp
```diff
@@ -0,0 +1,294 @@
...
+        node_parent[act] = parent;
+    }
+
+    std::vector<TaxId> relevant_taxids;
+    // 'considered_relevant_taxids' is used to make sure that there are no duplications in 'relevant_taxids'.
+    tsl::hopscotch_set<TaxId> considered_relevant_taxids;
+
+    if (accversion_to_taxid_map.size()) {
+        // Store only the taxonomic nodes that exists in the annotation matrix.
+        for (const pair<std::string, TaxId> &it : accversion_to_taxid_map) {
```
karasikov, line 178: pair is not an iterator
```suggestion
        for (const std::pair<std::string, TaxId> &pair : accversion_to_taxid_map) {
```

metagraph/src/annotation/taxonomy/tax_classifier.hpp
heracle: I've split it into multiple lines. I was not considering the comments, but now I fixed this in the entire PR

metagraph/src/annotation/taxonomy/tax_classifier.cpp
heracle: `_anno_matrix` will be used by `assign_class()` method in the next PR. 

Replaced the count+at with an `it = tree.find(node)`

Thanks for the review!

metagraph/src/annotation/taxonomy/tax_classifier.cpp
```diff
@@ -0,0 +1,299 @@
...
+
+#include <cmath>
+#include <string>
+#include <vector>
+
+#include "annotation/representation/annotation_matrix/annotation_matrix.hpp"
+#include "common/unix_tools.hpp"
+#include "common/utils/string_utils.hpp"
+
+#include "common/logger.hpp"
```
karasikov, line 11:
```suggestion
#include "common/utils/string_utils.hpp"
#include "common/logger.hpp"
```

metagraph/src/annotation/taxonomy/tax_classifier.cpp
```diff
@@ -0,0 +1,299 @@
...
+    } else if (utils::starts_with(utils::split_string(sample_label, ":")[1], "taxid|")) {
+        // e.g.   >kraken:taxid|2016032|NC_047834.1 Alteromonas virus vB_AspP-H4/4, complete genome
+        label_type = TAXID;
+        return false;
+    }
+
+    logger->error("Can't determine the type of the given label {}. "
+                  "Make sure the labels are in a recognized format.", sample_label);
+    exit(1);
+}
```
karasikov, line 32: it's quite unclear what this function is doing. Also, it's called only in one place, so I'd recommend removing it and move the code right there.

Or, if you need to call this function also in child classes, make it a properly: return label_type -- and try/false can be checked instantly after the call, e.g.:
```c++
... label_type = get_label_type(sample_label);

bool require_accversion_to_taxid_map = (label_type == GEN_BANK);
```

metagraph/src/annotation/taxonomy/tax_classifier.hpp
```diff
@@ -0,0 +1,198 @@
...
+
+    /**
+     *  node_parent stores a taxonomic tree representation as a taxid to taxid parent list.
+     */
+    tsl::hopscotch_map<TaxId, TaxId> node_parent;
+
+    tsl::hopscotch_map<std::string, TaxId> accversion_to_taxid_map;
+
+    double _lca_coverage_rate;
+    double _kmers_discovery_rate;
```
karasikov, line 103: you know what to do
```suggestion
    double lca_coverage_rate_;
    double kmers_discovery_rate_;
```

metagraph/src/annotation/taxonomy/tax_classifier.hpp
```diff
@@ -0,0 +1,198 @@
...
+
+    LabelType label_type;
+
+    /**
+     * node_depth returns the depth for each node in the taxonomic tree.
+     * The root is the unique node with maximal depth and all the leaves have depth equal to 1.
+     */
+    tsl::hopscotch_map<TaxId, uint32_t> node_depth;
+
+    TaxId root_node;
```
karasikov, line 93:
```suggestion
    TaxId root_node_;
```

metagraph/src/annotation/taxonomy/tax_classifier.hpp
```diff
@@ -0,0 +1,198 @@
...
+     */
+    virtual std::vector<TaxId> get_lca_taxids_for_seq(const std::string_view &sequence, bool reversed) const = 0;
+
+    LabelType label_type;
+
+    /**
+     * node_depth returns the depth for each node in the taxonomic tree.
+     * The root is the unique node with maximal depth and all the leaves have depth equal to 1.
+     */
+    tsl::hopscotch_map<TaxId, uint32_t> node_depth;
```
karasikov, line 91:
```suggestion
    tsl::hopscotch_map<TaxId, uint32_t> node_depth_;
```

metagraph/src/annotation/taxonomy/tax_classifier.hpp
```diff
@@ -0,0 +1,198 @@
...
+                               TaxId *best_lca,
+                               uint32_t *best_lca_dist_to_root) const;
+
+    /**
+     * Get the list of LCA taxid for each kmer in a given sequences.
+     * The sequence can be given in forward or in reversed orientation.
+     */
+    virtual std::vector<TaxId> get_lca_taxids_for_seq(const std::string_view &sequence, bool reversed) const = 0;
+
+    LabelType label_type;
```
karasikov, line 85:
```suggestion
    LabelType label_type_;
```

metagraph/src/annotation/taxonomy/tax_classifier.cpp
```diff
@@ -0,0 +1,299 @@
...
+        if (not accversion_to_taxid_map.count(acc_version)) {
+            return false;
+        }
+        *taxid = accversion_to_taxid_map.at(acc_version);
+        return true;
+    }
+
+    logger->error("Error: Could not get the taxid for label {}", label);
+    exit(1);
+}
```
karasikov, line 49: I'm confused by these helpers.

You are setting `label_type` somewhere else, then it's used here. So it requires a certain sequence of these calls right?
What is the point to separate these functions then?

metagraph/src/annotation/taxonomy/tax_classifier.cpp
```diff
@@ -0,0 +1,299 @@
...
+bool TaxonomyBase::get_taxid_from_label(const std::string &label, TaxId *taxid) const {
+    if (label_type == TAXID) {
+        *taxid = std::stoul(utils::split_string(label, "|")[1]);
+        return true;
+    } else if (TaxonomyBase::label_type == GEN_BANK) {
+        std::string acc_version = get_accession_version_from_label(label);
+        if (not accversion_to_taxid_map.count(acc_version)) {
+            return false;
+        }
+        *taxid = accversion_to_taxid_map.at(acc_version);
```
karasikov, line 43: Again, double lookup: `.count` + `.at` ...

metagraph/src/annotation/taxonomy/tax_classifier.cpp
```diff
@@ -0,0 +1,299 @@
...
+            } else {
+                rmq_data[row][i] = rmq_data[row - 1][i + delta];
+            }
+        }
+        delta *= 2;
+    }
+}
+
+std::vector<TaxId> TaxonomyClsAnno::get_lca_taxids_for_seq(const std::string_view &sequence, bool reversed) const {
+    cerr << "Assign class not implemented reversed = " << reversed << "\n";
```
karasikov, line 277: `std::cerr`

Why does it not require `std::`?
Do we have `using namespace std;` written somewhere?

karasikov: why not use the logger?

karasikov (changes requested):

metagraph/src/annotation/taxonomy/tax_classifier.cpp
heracle: I have moved it directly to the constructor. Thanks for the suggestion!

metagraph/src/annotation/taxonomy/tax_classifier.hpp
heracle: ah, sorry, I missed that. Replaced everywhere now, all the fields in the two objects `TaxonomyClsAnno` and `TaxonomyBase` have an `_` termination

metagraph/src/annotation/taxonomy/tax_classifier.cpp
heracle: I will need to use this `label_type` from different enterpoints from the CLI cmd. Currently, it is computed in the `TaxonomyClsAnno` constructor and will be further used inside `tax_class(sequence)` method where we will need to iterate the column labels and take the associated taxids. If you think that it would be more clear, I can move it as part of the `tax_class` and run it multiple times (should be fast, I guess).

metagraph/src/annotation/taxonomy/tax_classifier.cpp
heracle: added everything to the `throw std::runtime_error` line. I had to use all the parameters to not get compilation errors.

There is no `using namespace std;` in `metagraph/src`, but there are a few in the external libraries. Maybe it is taken from there.

metagraph/src/annotation/taxonomy/tax_classifier.hpp
```diff
@@ -0,0 +1,200 @@
+#ifndef __TAX_CLASSIFIER_HPP__
+#define __TAX_CLASSIFIER_HPP__
+
+#include <tsl/hopscotch_set.h>
+#include <tsl/hopscotch_map.h>
+#include "graph/annotated_dbg.hpp"
```
karasikov, line 6:
```suggestion

#include "graph/annotated_dbg.hpp"
```

metagraph/src/annotation/taxonomy/tax_classifier.hpp
```diff
@@ -0,0 +1,200 @@
...
+    enum LabelType {
+        UNASSIGNED,
+        GEN_BANK, // e.g. ">gi|1070643132|ref|NC_031224.1| Arthrobacter phage Mudcat, complete genome"
+        TAXID, // e.g. ">kraken:taxid|2016032|NC_047834.1 Alteromonas virus vB_AspP-H4/4, complete genome"
+    };
+
+    TaxonomyBase() {};
+    TaxonomyBase(const double lca_coverage_rate, const double kmers_discovery_rate)
+        : lca_coverage_rate_(lca_coverage_rate),
+          kmers_discovery_rate_(kmers_discovery_rate) {};
```
karasikov, line 28:
```suggestion
    TaxonomyBase() {}
    TaxonomyBase(double lca_coverage_rate, double kmers_discovery_rate)
        : lca_coverage_rate_(lca_coverage_rate),
          kmers_discovery_rate_(kmers_discovery_rate) {}
```

metagraph/src/annotation/taxonomy/tax_classifier.hpp
```diff
@@ -0,0 +1,200 @@
...
+     * @param [optional input] label_taxid_map_filepath ->  path to the acc_version to taxid lookup table (".accession2taxid").
+     *                                                      Mandatory if the taxid is not mentioned in the label string.
+     */
+    TaxonomyClsAnno(const graph::AnnotatedDBG &anno,
+                    const double lca_coverage_rate,
+                    const double kmers_discovery_rate,
+                    const std::string &tax_tree_filepath,
+                    const std::string &label_taxid_map_filepath = "");
+    TaxonomyClsAnno() {};
+    virtual ~TaxonomyClsAnno() {};
```
karasikov, line 127:
```suggestion
    TaxonomyClsAnno() {}
    virtual ~TaxonomyClsAnno() {}
```

metagraph/src/annotation/taxonomy/tax_classifier.hpp
```diff
@@ -0,0 +1,200 @@
...
+    using node_index = graph::SequenceGraph::node_index;
+
+    enum LabelType {
+        UNASSIGNED,
+        GEN_BANK, // e.g. ">gi|1070643132|ref|NC_031224.1| Arthrobacter phage Mudcat, complete genome"
+        TAXID, // e.g. ">kraken:taxid|2016032|NC_047834.1 Alteromonas virus vB_AspP-H4/4, complete genome"
+    };
+
+    TaxonomyBase() {};
+    TaxonomyBase(const double lca_coverage_rate, const double kmers_discovery_rate)
```
karasikov, line 26: virtual destructor is missing

metagraph/src/annotation/taxonomy/tax_classifier.hpp
karasikov: the virtual destructor is not needed unless you this class is used as a parent

metagraph/src/annotation/taxonomy/tax_classifier.hpp
```diff
@@ -0,0 +1,200 @@
...
+     *
+     * Given a set of kmers taken from one read, a query on the annotation matrix will return all the labels that
+     * correspond to at least $label_fraction$ percent of the kmers.
+     * Then, returns the LCA of the taxid nodes where these frequent labels are mapping to.
+     *
+     * @param [input] sequence -> the given sequence to classify.
+     * @param [input] label_fraction -> threshold used for taxonomic classification.
+     * @return -> the classification result: a taxid node
+     */
+    TaxId assign_class_toplabels(const std::string &sequence, const double label_fraction) const;
```
karasikov, line 143:
```suggestion
    TaxId assign_class_toplabels(const std::string &sequence, double label_fraction) const;
```

metagraph/src/annotation/taxonomy/tax_classifier.hpp
```diff
@@ -0,0 +1,200 @@
...
+     * @param [input] tree -> the taxonomic tree stored as a list of children.
+     * @param [output] tree_linearization -> the linearization of the received tree.
+     */
+    void dfs_statistics(const TaxId node,
+                        const ChildrenList &tree,
+                        std::vector<TaxId> *tree_linearization);
+
+    TaxId find_lca(const std::vector<TaxId> &taxids) const;
+
+    std::vector<TaxId> get_lca_taxids_for_seq(const std::string_view &sequence, bool reversed) const;
```
karasikov, line 178: check which of these methods can be made `static`, and also consider moving them to the .cpp file as ordinary functions.

The same for other classes.

metagraph/src/annotation/taxonomy/tax_classifier.cpp
```diff
@@ -0,0 +1,300 @@
...
+    if (num_taxid_failed) {
+        logger->warn("During the tax_tree_filepath {} parsing, {} taxids were not found out of {} total evaluations.",
+                     tax_tree_filepath, num_taxid_failed, relevant_taxids.size());
+    }
+
+    // Construct the output tree.
+    for (const TaxId &taxid : relevant_taxids) {
+        if (taxid == root_node_) {
+            continue;
+        }
```
karasikov, line 221:
```suggestion
        if (taxid == root_node_)
            continue;

```

metagraph/src/annotation/taxonomy/tax_classifier.cpp
```diff
@@ -0,0 +1,300 @@
...
+            considered_relevant_taxids.insert(full_parents_list[taxid]);
+        }
+
+        // Check if the current taxid is the root.
+        if (taxid == full_parents_list[taxid]) {
+            root_node_ = taxid;
+        }
+    }
+    if (num_taxid_failed) {
+        logger->warn("During the tax_tree_filepath {} parsing, {} taxids were not found out of {} total evaluations.",
```
karasikov, line 213:
```suggestion
        logger->warn("During the tax_tree_filepath {} parsing, {} taxids were not found out of {} total evaluations",
```

metagraph/src/annotation/taxonomy/tax_classifier.cpp
```diff
@@ -0,0 +1,300 @@
...
+        // Store only the taxonomic nodes that exists in the annotation matrix.
+        for (const std::pair<std::string, TaxId> &pair : accversion_to_taxid_map_) {
+            relevant_taxids.push_back(pair.second);
+            considered_relevant_taxids.insert(pair.second);
+        }
+    } else {
+        // If 'this->accversion_to_taxid_map' is empty, store the entire taxonomic tree.
+        for (auto it : full_parents_list) {
+            relevant_taxids.push_back(it.first);
+            considered_relevant_taxids.insert(it.first);
```
karasikov, line 189: Rename `it` -> `pair`.

`it` would be an iterator in the container.

Or, better write the variables explicitly:
`for (const auto &[???, ???] : full_parents_list) {`

metagraph/src/annotation/taxonomy/tax_classifier.cpp
```diff
@@ -0,0 +1,300 @@
...
+        label_type_ = TAXID;
+        require_accversion_to_taxid_map = false;
+    } else {
+        logger->error("Error: Can't determine the type of the given label {}. "
+                      "Make sure the labels are in a recognized format.", sample_label);
+        exit(1);
+    }
+
+    Timer timer;
+    if (require_accversion_to_taxid_map) {
```
karasikov, line 122: Some code is redundant
```suggestion
    if (utils::starts_with(sample_label, "gi|")) {
        // e.g.   >gi|1070643132|ref|NC_031224.1| Arthrobacter phage Mudcat, complete genome
        label_type_ = GEN_BANK;
    } else if (utils::starts_with(utils::split_string(sample_label, ":")[1], "taxid|")) {
        // e.g.   >kraken:taxid|2016032|NC_047834.1 Alteromonas virus vB_AspP-H4/4, complete genome
        label_type_ = TAXID;
    } else {
        logger->error("Error: Can't determine the type of the given label {}. "
                      "Make sure the labels are in a recognized format.", sample_label);
        exit(1);
    }

    Timer timer;
    if (label_type_ == GEN_BANK) {
```

metagraph/src/annotation/taxonomy/tax_classifier.cpp
```diff
@@ -0,0 +1,300 @@
...
+        }
+        if (input_accessions.size() == 0 || input_accessions.count(parts[1])) {
+            accversion_to_taxid_map_[parts[1]] = std::stoul(parts[2]);
+        }
+    }
+}
+
+TaxonomyClsAnno::TaxonomyClsAnno(const graph::AnnotatedDBG &anno,
+                                 const double lca_coverage_rate,
+                                 const double kmers_discovery_rate,
```
karasikov, line 91: Don't pass simple types such as int/double/etc by const values. Usually, there is no difference and it doesn't help, so the more compact option is better.
```suggestion
                                 double lca_coverage_rate,
                                 double kmers_discovery_rate,
```

metagraph/src/annotation/taxonomy/tax_classifier.cpp
```diff
@@ -0,0 +1,300 @@
...
+        if (it == accversion_to_taxid_map_.end()) {
+            return false;
+        }
+        *taxid = it->second;
+        return true;
+    }
+
+    logger->error("Error: Could not get the taxid for label {}", label);
+    exit(1);
+}
```
karasikov, line 32: where is this called?

karasikov (changes requested):

metagraph/src/annotation/taxonomy/tax_classifier.cpp
karasikov: Also, I'm not sure `full_parents_list` is a good name... What does this thing store?

metagraph/src/annotation/taxonomy/tax_classifier.cpp
heracle: `get_taxid_from_label` will be used in the next PR. I implemented the entire `.hpp` file in this PR  and wanted to add a reasoning about why/how to use the `enum LabelType`

metagraph/src/annotation/taxonomy/tax_classifier.cpp
heracle: I have merged together the `full_parents_list` and `node_parent_` to delete the entire `full_parents_list` object. In this way, I have also reduced the number of `hopscotch_map` accesses. Thanks for the suggestion!

metagraph/src/annotation/taxonomy/tax_classifier.cpp
heracle: Regarding logger messages, I am not very sure in what context to add the period and when to remove it. Can you please list some small examples to understand how to write these in the future?

metagraph/src/annotation/taxonomy/tax_classifier.hpp
heracle: I tried to create an object field only for those variables that are used in both constructor and `assign_class`. All the methods are reading or writing at least some of the object fields, thus, if we want to remove them from the object and make static/ordinary functions we will need to send the fields that are used as arguments.

e.g, for `get_accession_version_from_label` and `get_taxid_from_label` we need to send `label_type_` and `accversion_to_taxid_map_`. But `accversion_to_taxid_map_` can be empty if `LabelType=TAXID`.

Or, in case of `read_tree` we are computing both `root_node_` and `node_parent_`. I can send both of them as pointer parameters if you think that this will make the code more clear. Let me know

metagraph/src/annotation/taxonomy/tax_classifier.cpp
```diff
@@ -0,0 +1,305 @@
...
+namespace mtg {
+namespace annot {
+
+using mtg::common::logger;
+
+bool TaxonomyBase::get_taxid_from_label(const std::string &label, TaxId *taxid) const {
+    if (label_type_ == TAXID) {
+        *taxid = std::stoul(utils::split_string(label, "|")[1]);
+        return true;
+    } else if (label_type_ == GEN_BANK) {
```
karasikov, line 20: use switch-case for enums

metagraph/tests/annotation/taxonomy/test_taxonomy.cpp
```diff
@@ -0,0 +1,90 @@
...
+        {5, 1},
+        {6, 1},
+        {7, 1},
+        {8, 1},
+    };
+
+    std::vector<uint32_t> linearization = {
+        0, 1, 4, 7, 4, 8, 4, 1, 5, 1, 0, 2, 0, 3, 6, 3, 0
+    };
+    std::vector<std::vector<uint32_t> > expected_rmq = {
```
karasikov, line 78:
```suggestion
    std::vector<std::vector<uint32_t>> expected_rmq = {
```

metagraph/src/annotation/taxonomy/tax_classifier.hpp
```diff
@@ -0,0 +1,201 @@
...
+                               TaxId *best_lca,
+                               uint32_t *best_lca_dist_to_root) const;
+
+    /**
+     * Get the list of LCA taxid for each kmer in a given sequences.
+     * The sequence can be given in forward or in reversed orientation.
+     */
+    virtual std::vector<TaxId> get_lca_taxids_for_seq(const std::string_view &sequence, bool reversed) const = 0;
+
+    LabelType label_type_ = LabelType::UNASSIGNED;
```
karasikov, line 78: Does it make sense to allow it to be unassigned?

metagraph/src/annotation/taxonomy/tax_classifier.hpp
```diff
@@ -0,0 +1,201 @@
...
+     *                                          considered as a valid solution.
+     * @param [modified] 'node_scores' -> the current score for each node in the tree.
+     * @param [modified] 'nodes_already_propagated' -> the set of nodes that were previously processed.
+     * @param [modified] 'best_lca' -> the current classification prediction (node that exceeds the
+     *                                 `desired_number_kmers` threshold and is placed as close as possible to the leaves).
+     * @param [modified] 'best_lca_dist_to_root' -> the distance to the root for the current classification prediction.
+     */
+    void update_scores_and_lca(TaxId start_node,
+                               const tsl::hopscotch_map<TaxId, uint64_t> &num_kmers_per_node,
+                               uint64_t desired_number_kmers,
```
karasikov, line 66: This is a minimum threshold, as far as I got from the description. "Desired" doesn't reflect that the value can be much larger.

metagraph/src/annotation/taxonomy/tax_classifier.hpp
```diff
@@ -0,0 +1,201 @@
...
+     * @param [input] 'desired_number_kmers' -> the threshold score that a node has to exceed in order to be
+     *                                          considered as a valid solution.
+     * @param [modified] 'node_scores' -> the current score for each node in the tree.
+     * @param [modified] 'nodes_already_propagated' -> the set of nodes that were previously processed.
+     * @param [modified] 'best_lca' -> the current classification prediction (node that exceeds the
+     *                                 `desired_number_kmers` threshold and is placed as close as possible to the leaves).
+     * @param [modified] 'best_lca_dist_to_root' -> the distance to the root for the current classification prediction.
+     */
+    void update_scores_and_lca(TaxId start_node,
+                               const tsl::hopscotch_map<TaxId, uint64_t> &num_kmers_per_node,
```
karasikov, line 65: I see you call nodes of the taxonomy tree "nodes". Would be good to distinguish them with the nodes of the de Bruijn graph (k-mers). One way is to always write "taxid" and not "node". Another way could be to write something like `tnode`.

metagraph/src/annotation/taxonomy/tax_classifier.hpp
```diff
@@ -0,0 +1,201 @@
...
+     * @param [input] 'num_kmers_per_node[taxid]' -> the number of kmers 'k' with taxonomic_map[k]=taxid.
+     * @param [input] 'desired_number_kmers' -> the threshold score that a node has to exceed in order to be
+     *                                          considered as a valid solution.
+     * @param [modified] 'node_scores' -> the current score for each node in the tree.
+     * @param [modified] 'nodes_already_propagated' -> the set of nodes that were previously processed.
+     * @param [modified] 'best_lca' -> the current classification prediction (node that exceeds the
+     *                                 `desired_number_kmers` threshold and is placed as close as possible to the leaves).
+     * @param [modified] 'best_lca_dist_to_root' -> the distance to the root for the current classification prediction.
+     */
+    void update_scores_and_lca(TaxId start_node,
```
karasikov, line 64: I don't see where this function is implemented or used. Remove it from this PR.
Also, check for this all other functions.

karasikov (changes requested):

metagraph/src/annotation/taxonomy/tax_classifier.hpp
heracle: sure, will remove them. I was just trying to present the complete structure of the objects: implement the entire `tax_classifier.hpp ` in this PR and not modify it in the next ones. But I see that it can be more clear to have the function header in the PR where it is implemented. I Will do the changes tomorrow, thanks!

metagraph/src/annotation/taxonomy/tax_classifier.hpp
heracle: you are right. Removed now

metagraph/src/annotation/taxonomy/tax_classifier.hpp
heracle: I have removed the `update_scores_and_lca` header from this PR, but I will address the review in the next PR, thanks!

metagraph/src/annotation/taxonomy/tax_classifier.cpp
```diff
@@ -0,0 +1,287 @@
...
+#include "common/logger.hpp"
+
+namespace mtg {
+namespace annot {
+
+using mtg::common::logger;
+
+bool TaxonomyBase::get_taxid_from_label(const std::string &label, TaxId *taxid) const {
+    switch (label_type_) {
+        case TAXID :
```
karasikov, line 18:
```suggestion
        case TAXID:
```

metagraph/src/annotation/taxonomy/tax_classifier.hpp
```diff
@@ -0,0 +1,136 @@
...
+    TaxonomyBase() {}
+    TaxonomyBase(double lca_coverage_rate, double kmers_discovery_rate)
+        : lca_coverage_rate_(lca_coverage_rate),
+          kmers_discovery_rate_(kmers_discovery_rate) {}
+    virtual ~TaxonomyBase() {}
+
+  protected:
+    std::string get_accession_version_from_label(const std::string &label) const;
+
+    bool get_taxid_from_label(const std::string &label, TaxId *taxid) const;
```
karasikov, line 34: write what it returns in a comment

karasikov: actually, this is never used. Remove

karasikov (changes requested):

metagraph/src/annotation/taxonomy/tax_classifier.hpp
heracle: removed, thanks

PR #352, opened 2021-08-05
building docker image and pushing to ghcr.io registry

izcram:
On commits/PR merges on master, build and push docker image to github packages/containers (ghcr.io).

Moving away from dockerhub as autobuilds are no longer part of the free plan. Alternatively, with building using github actions and pushing to dockerhub handling credentials properly would be a bit involved (and possibly not free either).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,6 +2,11 @@ name: MetaGraph CI
 
 on: [push]
 
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
+
 jobs:
 
   Linux:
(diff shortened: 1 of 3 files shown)


izcram:
test image https://github.com/ratschlab/metagraph/pkgs/container/metagraph/6218789 was built using https://github.com/ratschlab/metagraph/commit/4c0a95167fe1dd11a2c518851f2af1a57a69bded

.github/workflows/main.yml
```diff
@@ -173,6 +178,43 @@ jobs:
     - name: run integration tests
       run: cd metagraph/build && make check
 
+  Build-and-Push-Docker:
+    # adapted from https://docs.github.com/en/actions/guides/publishing-docker-images#publishing-images-to-github-packages
+    if: github.ref == 'refs/heads/master'
+    needs: [Linux]
+    runs-on: ubuntu-latest
```
karasikov, line 185: Can we fix an LTS version, so this doesn't break if something changes in the update?

karasikov (approved):

karasikov (changes requested):
So it's not `ratschlab/metagraph` anymore, right?
Should we also update the docs?

metagraph/docs/source/installation.rst
```diff
@@ -23,13 +23,15 @@ Docker container
 
 If docker is available on your system, you can immediately get started with::
 
-    docker run -v ${DATA_DIR_HOST}:/mnt ratschlab/metagraph \
+    docker run -v ${DATA_DIR_HOST}:/mnt ghcr.io/ratschlab/metagraph:latest \
```
karasikov, line 26: there is also one line here that probably needs to be updated
https://github.com/ratschlab/metagraph/blob/0c88d4b6c6b3128f8cc0e7036caaa89684b7f893/Dockerfile#L90

karasikov (approved):

metagraph/docs/source/installation.rst
izcram: yes good point!

PR #358, opened 2021-11-29
API support for count/coord indexes, parallel graph search, and intermediate query representation

thomastzhou:
New features:

- [x] Support for `--count-kmers` and `--query-coords` in API and server.
- [x] New default coordinate behaviour: continuous coordinate sequences are collapsed into a range. All coordinates can be shown using `--verbose-coords` query flag.
- [x] API level parallel requests with `MultiGraphClient`

Backend has been re-written to use an intermediate query representation instead always returning string output (and re-parsing string into JSON for server). From here, can also allow for more flexible server JSON responses.


thomastzhou:
I believe these failed checks are related to AppleClang 13 and not the changes, this is the first commit that has been checked with 13 (previous ones are using 12_4). 

On MacOS Monterey with AppleClang 13 even master branch fails to build. I built these with g++ 11 instead.

karasikov:
> I believe these failed checks are related to AppleClang 13 and not the changes, this is the first commit that has been checked with 13 (previous ones are using 12_4).
> 
> On MacOS Monterey with AppleClang 13 even master branch fails to build. I built these with g++ 11 instead.

Feel free to adapt [main.yml](https://github.com/ratschlab/metagraph/blob/master/.github/workflows/main.yml)

thomastzhou:
I've left issue #359 for now and outlined a way to squash this in the workflow. It depends on whether or not you want to support AppleClang 13 now or point any users to another compiler for now

metagraph/src/cli/query.hpp
```diff
@@ -6,6 +6,13 @@
 #include <memory>
 #include <string>
 #include <vector>
+#include <utility>
+#include <variant>
+#include <optional>
+#include <json/json.h>
+#include <sdsl/int_vector.hpp>
+#include "common/vector.hpp"
+#include "graph/alignment/aligner_alignment.hpp"
```
karasikov, line 15:
```suggestion
#include <optional>

#include <json/json.h>
#include <sdsl/int_vector.hpp>

#include "common/vector.hpp"
#include "graph/alignment/aligner_alignment.hpp"
```

metagraph/src/cli/query.hpp
```diff
@@ -45,38 +52,160 @@ construct_query_graph(const graph::AnnotatedDBG &anno_graph,
...
+ * as a typesafe union std::variant.
+ *
+ * Note that this is an intermediate storage class and does not (1) store the actual sequence
+ * nor (2) perform any of the search itself.
+ */
+class SeqSearchResult {
+
+public:
+
+    typedef std::string label;
```
karasikov, line 98:
```suggestion
class SeqSearchResult {
  public:
    typedef std::string label;
```

karasikov: also rename `label` -> `Label`

metagraph/src/cli/query.hpp
```diff
@@ -45,38 +52,160 @@ construct_query_graph(const graph::AnnotatedDBG &anno_graph,
...
+    typedef std::vector<label> label_vec;
+    typedef std::vector<std::pair<label, size_t>> label_count_vec;
+    typedef std::vector<std::pair<label, sdsl::bit_vector>> label_sig_vec;
+    typedef std::vector<std::pair<label, std::vector<size_t>>> label_quantile_vec;
+    typedef std::vector<std::pair<label, std::vector<SmallVector<uint64_t>>>> label_coord_vec;
```
karasikov, line 103:
```suggestion
    typedef std::vector<std::pair<label, std::vector<SmallVector<uint64_t>>>> LabelCoordVec;
```

metagraph/src/cli/query.hpp
```diff
@@ -45,38 +52,160 @@ construct_query_graph(const graph::AnnotatedDBG &anno_graph,
...
+
+    typedef std::variant<
+            label_vec,
+            label_count_vec,
+            label_sig_vec,
+            label_quantile_vec,
+            label_coord_vec
+            > result_type;
```
karasikov, line 111:
```suggestion
    typedef std::variant<label_vec,
                         label_count_vec,
                         label_sig_vec,
                         label_quantile_vec,
                         label_coord_vec> result_type;
```

metagraph/src/cli/query.hpp
```diff
@@ -45,38 +52,160 @@ construct_query_graph(const graph::AnnotatedDBG &anno_graph,
...
+                          bool suppress_unlabeled,
+                          bool expand_coords,
+                          const graph::AnnotatedDBG &anno_graph) const;
+
+private:
+
+    QuerySequence sequence;     // query sequence this result represents
+    result_type result;         // result vector of labels and additional info
+
+};
```
karasikov, line 161:
```suggestion

  private:
    QuerySequence sequence; // query sequence this result represents
    result_type result; // result vector of labels and additional info
};
```

metagraph/src/cli/query.hpp
```diff
@@ -45,38 +52,160 @@ construct_query_graph(const graph::AnnotatedDBG &anno_graph,
...
+
     const Config &config_;
     const graph::AnnotatedDBG &anno_graph_;
     std::unique_ptr<graph::align::DBGAlignerConfig> aligner_config_;
     ThreadPool &thread_pool_;
 
     void batched_query_fasta(mtg::seq_io::FastaParser &fasta_parser,
-                             const std::function<void(const std::string &)> &callback);
+                             const std::function<void(const SeqSearchResult &)> &callback);
+
```
karasikov, line 208:
```suggestion
                             const std::function<void(const SeqSearchResult &)> &callback);
```

metagraph/src/cli/query.cpp
```diff
@@ -2,6 +2,7 @@
 
 #include <ips4o.hpp>
 #include <tsl/ordered_set.h>
+#include <mutex>
```
karasikov, line 5:
```suggestion
#include <mutex>

#include <ips4o.hpp>
#include <tsl/ordered_set.h>
```

metagraph/src/cli/query.cpp
```diff
@@ -38,123 +39,348 @@ using mtg::graph::boss::BOSSConstructor;
...
+/**
+ * Given a vector of kmer label matched coordinates, collapse continuous ranges of coordinates
+ * to start-end tuples.
+ *
+ * @param coords    the vector of tuples as stored in a SeqSearchResult
+ *                  originally returned from AnnotatedDBG::get_kmer_coordinates
+ * @return vector of 'begin-end' range string representations
+ */
+std::vector<std::string> get_collapsed_coord_ranges(
+        const std::vector<SmallVector<uint64_t>> &tuples) {
```
karasikov, line 63:
```suggestion
std::vector<std::string>
get_collapsed_coord_ranges(const std::vector<SmallVector<uint64_t>> &tuples) {
```

metagraph/src/cli/query.hpp
```diff
@@ -45,38 +52,160 @@ construct_query_graph(const graph::AnnotatedDBG &anno_graph,
...
+    /**
+     * Construct an instance of SeqSearchResult by providing QuerySequence instance describing
+     * sequence and result from that query on an annotated DBG. Will move both into this instance.
+     *
+     * @param sequence  rvalue reference to QuerySequence (WILL BE MOVED)
+     * @param result    rvalue reference to result_type (WILL BE MOVED)
+     */
+    SeqSearchResult(QuerySequence&& sequence, result_type&& result)
+            : sequence(std::move(sequence)),
+            result(std::move(result)) {};
```
karasikov, line 122:
```suggestion
    SeqSearchResult(QuerySequence&& sequence, result_type&& result)
        : sequence(std::move(sequence)),
          result(std::move(result)) {}
```

karasikov (changes requested):
Please apply similar stylistic changes in the rest of the PR.

karasikov:
> I've left issue #359 and outlined a way to squash this in the workflow. It depends on whether or not you want to support AppleClang 13 now or point any users to another compiler for now

I fixed that. Please merge master into this branch to pull the fix.

metagraph/src/cli/config/config.cpp
```diff
@@ -181,11 +181,11 @@ Config::Config(int argc, char *argv[]) {
         } else if (!strcmp(argv[i], "--max-count-q")) {
             max_count_quantile = std::min(std::stod(get_value(i++)), 1.);
         } else if (!strcmp(argv[i], "--count-bins-q")) {
-            for (const auto &border : utils::split_string(get_value(i++), " ")) {
+            for (const auto &border: utils::split_string(get_value(i++), " ")) {
```
karasikov, line 184:
```suggestion
            for (const auto &border : utils::split_string(get_value(i++), " ")) {
```

karasikov (changes requested):
It still shows changes that have been merged to master. So this means the commits from master haven't been merged into this branch. Can you do that?
```
git fetch
git checkout tz/count_api
git merge origin/master
```

metagraph/src/cli/config/config.cpp
thomastzhou: When I merge it already says its up to date. Maybe because I rebased the commits from master into the branch before?

karasikov: yeah, weird...

metagraph/api/python/metagraph/client.py
```diff
@@ -48,22 +48,33 @@ def search(self, sequence: Union[str, Iterable[str]],
...
+
+        if align:
+            if len(alignments) == len(search_results):
+                # Zip best alignment results
+                for alignment, search_result in zip(alignments, search_results):
+                    if 'alignments' in alignment and len(alignment['alignments']) > 0:
+                        search_result['best_alignment'] = alignment['alignments'][0]
+                    else:
+                        search_result['best_alignment'] = {}
+
```
karasikov, line 76: Could you clarify what is going on here?

Why is this only for `len(alignments) == len(search_results)`? Can they have different length?

metagraph/src/cli/query.cpp
```diff
@@ -39,122 +41,336 @@ using mtg::graph::boss::BOSSConstructor;
...
+
+    // Keep track of the beginning of the range (just use a sentinel)
+    uint64_t begin = 0;
+    uint64_t current = 0;
+    bool in_range = false;
+
+    for (const auto &coords : tuples) {
+        if (coords.empty()) continue;
+
+        // TODO: Look for example where row_tuples are multiple values?
```
karasikov, line 65: What do you mean?

karasikov: oh, never mind. I see what you mean...

karasikov: You can think of the case where the indexed sequence is AAA...A and the query is AAA.
So essentially every k-mer gets all the coordinates.

The current implementation will only print one coordinate range, which is incorrect. So this should be fixed. But we should agree on the desired output format first.

metagraph/api/python/metagraph/client.py
thomastzhou: With search with alignment it should only retrieve the best alignment I believe? Although I agree now I will adjust this to make it more robust.

karasikov: If it must always return 1 alignment, this should be an assertion instead of a check

karasikov: ping

metagraph/api/python/metagraph/client.py
```diff
@@ -170,51 +204,135 @@ def ready(self) -> bool:
...
+            futures[name] = executor.submit(graph_client.align, sequence, min_exact_match,
+                                            max_alternative_alignments,
+                                            max_num_nodes_per_seq_char)
 
-        return result
+        print(f'Made {len(self.graphs)} requests with {num_processes} threads...')
+
+        # Shutdown executor but do not stop futures
+        executor.shutdown(wait=False)
+        return futures
```
karasikov, line 315: Do we really benefit from returning Futures instead of just always waiting for everything to finish and returning the actual dictionary?

PR #361, opened 2021-12-03
reliability improvements

karasikov:


metagraph/src/common/sorted_sets/sorted_set_disk_base.cpp
```diff
@@ -58,15 +58,23 @@ ChunkedWaitQueue<T>& SortedSetDiskBase<T>::data(bool free_buffer) {
...
-std::vector<std::string> SortedSetDiskBase<T>::files_to_merge() {
+std::vector<std::string> SortedSetDiskBase<T>::get_chunks(bool free_buffer) {
     std::unique_lock<std::mutex> exclusive_lock(mutex_);
     std::unique_lock<std::shared_timed_mutex> multi_insert_lock(multi_insert_mutex_);
     // write any residual data left
     if (!data_.empty()) {
         sort_and_dedupe();
-        dump_to_file(true /* is_done */);
+        dump_to_file();
     }
```
hmusta, line 68: Since the data is dumped regardless, is there a reason to have `free_buffer` as a parameter here? When would it be useful to set it to `false`?

hmusta (changes requested):

metagraph/src/common/sorted_sets/sorted_set_disk_base.cpp
hmusta: ah, I see, it's for forcing a deallocation...regardless, it doesn't look like true is ever passed to this method

karasikov: Yeah, it's made to support re-using the object more than once. It was supported in other similar methods, so I kept that.

PR #370, opened 2021-12-23
Output counts for aggregated columns

karasikov:
Also added support for printing statistics for counts, e.g.:
```
./metagraph stats -a annotation.column.annodbg --print-counts-hist --count-quantiles "0.0 0.5 1.0"
```

### DIFF
--- a/metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
+++ b/metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
@@ -266,7 +266,7 @@ void ColumnCompressed<Label>::serialize_counts(const std::string &filename) cons
     }
 
     logger->info("Num relation counts: {}", num_counts);
-    logger->info("Total relation count: {}", sum_counts);
+    logger->info("Relation counts sum: {}", sum_counts);
 }
 
 template <typename Label>
--- a/metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
+++ b/metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
@@ -234,23 +234,35 @@ void ColumnCompressed<Label>::serialize_counts(const std::string &filename) cons
 
     for (size_t j = 0; j < relation_counts_.size(); ++j) {
         if (!relation_counts_[j].size()) {
-            sdsl::int_vector<>(bitmatrix_[j]->num_set_bits(), 0, count_width_).serialize(out);
+            sdsl::int_vector<>(bitmatrix_[j]->num_set_bits(), 0, 1).serialize(out);
 
         } else {
             assert(relation_counts_[j].size() == bitmatrix_[j]->num_set_bits()
                 && "Binary relation matrix must not be changed while adding relation counts");
 
+            // pack counts
+            uint64_t max_count = 0;
             for (uint64_t v : relation_counts_[j]) {
                 num_counts++;
                 sum_counts += v;
+                max_count = std::max(max_count, v);
             }
 
-            relation_counts_[j].serialize(out);
+            const uint8_t packed_width = sdsl::bits::hi(max_count) + 1;
+            if (packed_width == relation_counts_[j].width()) {
+                relation_counts_[j].serialize(out);
+            } else {
+                // TODO: use int_vector_buffer<>
+                sdsl::int_vector<> packed_counts(relation_counts_[j].size(), 0, packed_width);
+                std::copy(relation_counts_[j].begin(), relation_counts_[j].end(),
+                          packed_counts.begin());
+                packed_counts.serialize(out);
+            }
         }
     }
 
     for (size_t j = relation_counts_.size(); j < bitmatrix_.size(); ++j) {
-        sdsl::int_vector<>(bitmatrix_[j]->num_set_bits(), 0, count_width_).serialize(out);
+        sdsl::int_vector<>(bitmatrix_[j]->num_set_bits(), 0, 1).serialize(out);
     }
 
     logger->info("Num relation counts: {}", num_counts);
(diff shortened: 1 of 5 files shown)


metagraph/src/cli/stats.cpp
```diff
@@ -226,6 +247,70 @@ void print_stats(const Annotator &annotation) {
...
+        std::cout << "===================== COUNTS STATS =====================" << std::endl;
+        if (!dynamic_cast<const annot::ColumnCompressed<> *>(&annotation)) {
+            logger->error("Printing statistics for counts is currently only supported for"
+                          " ColumnCompressed annotations");
+            exit(1);
+        }
+        // the annotation is not needed anymore -> free memory
+        anno_p.reset();
+
+        std::cout << fmt::format("Column-index\tLabel\tNum-counts");
```
hmusta, line 260: Maybe change the order to "Index NumCounts Label"? The label could be of any length, so it won't misalign the table columns if this happens.

metagraph/src/cli/stats.cpp
```diff
@@ -226,6 +247,70 @@ void print_stats(const Annotator &annotation) {
...
+                    for (auto c : counts) {
+                        counts_hist[c]++;
+                    }
+                    count_hist_v.assign(counts_hist.begin(), counts_hist.end());
+                }
+
+                ips4o::parallel::sort(count_hist_v.begin(), count_hist_v.end(),
+                                      utils::LessFirst(), get_num_threads());
+
+                std::cout << fmt::format("{}\t{}\t{}", j, label, counts.size());
```
hmusta, line 290: see above

metagraph/src/cli/stats.cpp
hmusta: Though I guess it'll be misaligned anyway since "Column-index" is a long header name

metagraph/src/cli/stats.cpp
```diff
@@ -177,7 +180,25 @@ void print_brwt_stats(const annot::binmat::BRWT& brwt) {
...
+        logger->error("Cannot load annotations from file '{}'", fname);
+        exit(1);
+    }
+
+    using RowDiffCol = annot::binmat::RowDiff<annot::binmat::ColumnMajor>;
+    if (auto *rd = dynamic_cast<const RowDiffCol *>(&annotation.get_matrix())) {
+        std::string anchors_fname = utils::make_suffix(config.infbase,
+                                                       annot::binmat::kRowDiffAnchorExt);
+        if (!config.infbase.empty() && std::filesystem::exists(anchors_fname)) {
+            const_cast<RowDiffCol *>(rd)->load_anchor(anchors_fname);
```
hmusta, line 198: does it explain in the manual that one needs to keep the anchor file after constructing the annotator in order for this to work?

metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
```diff
@@ -234,27 +234,39 @@ void ColumnCompressed<Label>::serialize_counts(const std::string &filename) cons
 
     for (size_t j = 0; j < relation_counts_.size(); ++j) {
         if (!relation_counts_[j].size()) {
-            sdsl::int_vector<>(bitmatrix_[j]->num_set_bits(), 0, count_width_).serialize(out);
+            sdsl::int_vector<>(bitmatrix_[j]->num_set_bits(), 0, 1).serialize(out);
```
hmusta, line 237: Explain with a comment that the serialization format is: bit vector, counts. It'll be easier to understand what's happening here

metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
```diff
@@ -234,27 +234,39 @@ void ColumnCompressed<Label>::serialize_counts(const std::string &filename) cons
...
-            relation_counts_[j].serialize(out);
+            const uint8_t packed_width = sdsl::bits::hi(max_count) + 1;
+            if (packed_width == relation_counts_[j].width()) {
+                relation_counts_[j].serialize(out);
+            } else {
+                // TODO: use int_vector_buffer<>
+                sdsl::int_vector<> packed_counts(relation_counts_[j].size(), 0, packed_width);
+                std::copy(relation_counts_[j].begin(), relation_counts_[j].end(),
+                          packed_counts.begin());
+                packed_counts.serialize(out);
```
hmusta, line 259: if you want to save a bit more space doing this, check out sdsl's implementation of `bit_compress`, they do this in-place:

https://github.com/simongog/sdsl-lite/blob/master/include/sdsl/util.hpp#L403

hmusta (changes requested):

metagraph/src/cli/stats.cpp
karasikov: Yeah, it's tab-separated, so will be easily parsable anyway

metagraph/src/cli/stats.cpp
karasikov: This format is actually never used in practice, the manuals say you're supposed to convert it to something else and not keep the columns separately from the anchor bitmap. I guess we could remove this, could be a thing for a next refactoring.

metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
karasikov: done

metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
karasikov: Yeah, I also wanted to do that first, but this method is `const`, so I can't change the width of the existing vector. That's why you see this workaround.

PR #372, opened 2022-01-06
hide some obscure flags from usage help messages in CLI

karasikov:


karasikov:
@hmusta @akahles what else should we hide to make the CLI look more straightforward?

PR #373, opened 2022-01-20
Alignment to labeled subgraphs and counting DBGs

hmusta:
Supercedes #336


metagraph/tests/graph/test_aligner.cpp
```diff
@@ -601,8 +603,13 @@ TYPED_TEST(DBGAlignerTest, align_noise_in_branching_point) {
...
 
     EXPECT_EQ(query.size() - k + 2, path.size());
-    EXPECT_EQ(reference_1 + "T", path.get_sequence());
-    EXPECT_EQ("4=1D7=", path.get_cigar().to_string());
+    if (!path.get_orientation()) {
+        EXPECT_EQ("AAAACTTTTTTT", path.get_sequence());
+        EXPECT_EQ("4=1D7=", path.get_cigar().to_string());
+    } else {
+        EXPECT_EQ("AAAAAAACTTTT", path.get_sequence());
+        EXPECT_EQ("7=1D4=", path.get_cigar().to_string());
```
karasikov, line 611: rc of `AAAACTTTTTTT` is `AAAAAAAGTTTT`. Why do you check for `AAAAAAACTTTT` ?

metagraph/src/cli/align.hpp
```diff
@@ -19,7 +18,7 @@ class Config;
 
 graph::align::DBGAlignerConfig initialize_aligner_config(const Config &config);
 
-template <class Graph = graph::DeBruijnGraph>
+template <class Graph, typename... AlignerArgs>
 std::unique_ptr<graph::align::IDBGAligner>
 build_aligner(const Graph &graph, const graph::align::DBGAlignerConfig &aligner_config);
```
karasikov, line 23: it will become simpler with `template <class Graph, typename Aligner>`, also the usage will be more straightforward because you'll see every time which aligner exactly is going to be used.

karasikov (changes requested):

metagraph/src/graph/annotated_dbg.cpp
```diff
@@ -25,19 +25,14 @@ typedef std::pair<Label, size_t> StringCountPair;
 
 AnnotatedSequenceGraph
 ::AnnotatedSequenceGraph(std::shared_ptr<SequenceGraph> graph,
-                         std::unique_ptr<Annotator>&& annotation,
+                         std::shared_ptr<Annotator> annotation,
```
karasikov, line 28: Why would anyone create multiple instances of AnnotatedDBG with the same (graph, annotation) pairs?
You can always create one and share it across.

karasikov (changes requested):

metagraph/src/cli/align.cpp
```diff
@@ -426,21 +414,29 @@ int align_to_graph(Config *config) {
...
+                    aln_graph = std::make_shared<CanonicalDBG>(aln_graph, 100'000);
 
-                if (auto *canonical = dynamic_cast<CanonicalDBG*>(aln_graph.get()))
-                    aln_graph = std::make_shared<CanonicalDBG>(*canonical);
+                std::shared_ptr<AnnotatedDBG> aln_anno_graph;
+                std::shared_ptr<IDBGAligner> aligner;
 
-                auto aligner = build_aligner(*aln_graph, aligner_config);
+                if (annotator) {
+                    aln_anno_graph = std::make_shared<AnnotatedDBG>(aln_graph, annotator);
```
karasikov, line 427: You're creating an annotated graph from a pair of potentially incompatible (graph, annotator) pair.

The annotator is created for `base_graph` and hence should be used with it.
Now you wrap `base_graph` into something else (`aln_graph`), so it's no longer compatible.
You are not supposed to create such cases.

I guess this is why you need `get_base_node` and stuff in downstream to recover access to the graph with which you use the aligner?

karasikov (changes requested):

metagraph/src/cli/align.cpp
```diff
@@ -426,21 +414,29 @@ int align_to_graph(Config *config) {
...
             auto process_batch = [&,graph](SeqBatch batch) {
-                // make a shared_ptr in a thread-safe way
-                std::shared_ptr<DeBruijnGraph> aln_graph(
-                    std::shared_ptr<DeBruijnGraph>{}, graph.get()
-                );
+                // make an aliased shared_ptr in a thread-safe way
+                auto aln_graph = std::shared_ptr<DeBruijnGraph>(std::shared_ptr<DeBruijnGraph>{}, graph.get());
+
+                if (aln_graph->get_mode() == DeBruijnGraph::PRIMARY)
+                    aln_graph = std::make_shared<CanonicalDBG>(aln_graph, 100'000);
```
karasikov, line 421: Isn't this already wrapped in line 337?

metagraph/src/cli/align.cpp
hmusta: It depends on how "compatibility" with AnnotatedDBG is defined. If it's based on simply the number of nodes in the graph, then yes, this breaks things.

However, if you base compatibility on the functionality of the following workflow:
1) run `map_to_nodes` on the sequence
2) extract labels for the returned nodes.

Then compatibility shouldn't be broken.

hmusta: The place where the inconsistency happens is that other operations like `map_to_nodes_sequentially` and `call_outgoing_kmers` do not map to/return canonical k-mers (because this would lead to false traversal cycles). So for these, there needs to be some interface for fetching labels on these calls.

metagraph/src/graph/annotated_dbg.cpp
hmusta: The idea behind this is that when pairing an annotator with a graph that has an underlying cache, the cache update operations have to lock every time, so this slows down performance.

If there's a single annotator which is shared across multiple wrappers of the same underlying graph, then each wrapper can be assigned to a separate thread and there's no contention.

metagraph/src/cli/align.cpp
hmusta: yes, that's the case. In a different branch, I have an if-statement before this which checks if the graph can be cached, then wraps the base graph if that's the case. So, by the time it gets here, you could be left with a primary graph, so this is meant to fix that case.

This statement is left over from that, so it should have been spliced out.

metagraph/src/cli/align.cpp
karasikov: It's already defined by `check_compatibility()`. For some reason it was missing in the constructor, so your tests passed. As you see, they don't anymore: 8e0adfd

In fact, this was checked in any method. So you just created that class and never called any method from it.

metagraph/src/graph/annotated_dbg.cpp
karasikov: That idea is not realized in this PR. You still used the global cache.

metagraph/src/cli/align.cpp
hmusta: `check_compatibility` is defined differently between `AnnotatedSequenceGraph` and `AnnotatedDBG`. In the latter, it checks the base graph, rather than the graph itself.

However, if something like `get_base_graph` existed in `SequenceGraph`, then they could be defined the same way.

hmusta: In the end, I didn't use the `AnnotatedDBG` interface because it would have added a bunch of overhead to the workflow.

Essentially, I need something that can take a bunch of graph nodes (which may or may not be from the wrapper), and map fetch a vector of label IDs for each one. I want to be able to do this in a way that minimizes remapping to the graph, and also doesn't involve a bunch of unnecessary conversion between labels and label IDs.

karasikov: I don't see a problem with them being defined differently.
What does it have to do with breaking the compatibility in this line?

metagraph/src/graph/annotated_dbg.cpp
hmusta: This is a vestige from the old version of `DBGSuccinct::CachedView`. Originally, that class handled all of the caching, but then this was changed and some caching duties went back to `CanonicalDBG`.

hmusta: So now, this branch needs to be changed so that the wrapping to `CanonicalDBG` happens separately for each batch of sequences.

metagraph/src/cli/align.cpp
karasikov: The check was changed in `9f4e8e5b73571d874dd93eedc53079f71615023c`. So now it's not even possible to add check_compatibility() to the constructor...

On the other hand, without that check, everything is compatible. Brilliant.

metagraph/src/graph/annotated_dbg.cpp
karasikov: I'm reviewing this PR, I don't look at any other branches. If there is something there necessary for this, merge them into this PR.

metagraph/tests/graph/test_aligner.cpp
karasikov: Sure, I just noticed this and was curious. Maybe add a comment or something explaining the test?

metagraph/src/cli/load/load_graph.hpp
```diff
@@ -17,14 +18,21 @@ template <class Graph>
 std::shared_ptr<Graph> load_critical_graph_from_file(const std::string &filename) {
     auto graph = std::make_shared<Graph>(2);
     if (!graph->load(filename)) {
-        mtg::common::logger->error("Cannot load graph from file '{}'", filename);
+        common::logger->error("Cannot load graph from file '{}'", filename);
         exit(1);
     }
+
+    if (graph->get_mode() == graph::DeBruijnGraph::PRIMARY && !(graph->get_k() % 2))
+        common::logger->warn("PRIMARY mode graphs of even order k are less efficient to query. Consider using an odd value of k.");
```
karasikov, line 26: This warning should be moved where this is queried and maybe to `build` as well. It makes more sense to print this in `build` than here.

metagraph/src/cli/load/load_graph.cpp
```diff
@@ -71,5 +71,16 @@ std::shared_ptr<DeBruijnGraph> load_critical_dbg(const std::string &filename) {
...
+std::shared_ptr<CanonicalDBG> primary_to_canonical(std::shared_ptr<DeBruijnGraph> graph,
+                                                   size_t cache_size) {
+    if (graph->get_mode() != DeBruijnGraph::PRIMARY) {
+        logger->error("Only primary mode graphs can be wrapped into canonical mode.");
+        exit(1);
+    }
+
+    logger->trace("Primary graph wrapped into canonical");
+    return std::make_shared<CanonicalDBG>(graph, cache_size);
+}
```
karasikov, line 83: Why not return `CanonicalDBG`? You can always move it to create a new shared pointer from it.

karasikov (changes requested):

metagraph/src/cli/load/load_graph.cpp
karasikov: So in such cases, it's always better to return the object itself and not a shared pointer. If you follow this rule, it will be easier to see further simplifications.

karasikov:
Sorry about this, I thought it shows the diff incorrectly as I remember we renamed `get_annotation()` to `get_annotator()` a while ago. Though, it turns out it was actually one of the commits in this PR, so it's fine.

metagraph/src/cli/load/load_graph.hpp
hmusta: I guess it's better to just have it in `build`, rather than to keep bugging the user about it with every query.

metagraph/src/graph/annotated_dbg.cpp
```diff
@@ -49,6 +50,11 @@ bool AnnotatedDBG::check_compatibility() const {
     // TODO: find a way to fix this thing to add check_compatibility() to the constructor
     // of AnnotatedSequenceGraph.
     // Maybe make it how it was before 9f4e8e5b73571d874dd93eedc53079f71615023c
+
+    // TODO: what if CanonicalDBG is not the highest level? find a better way to do this
+    if (const auto *canonical = dynamic_cast<const CanonicalDBG*>(&dbg_))
+        return canonical->get_graph().max_index() == annotator_->num_objects();
```
karasikov, line 56: CanonicalDBG should not be passed to this class. It should always be the base graph.

karasikov: Ah, never mind. Looks like we've been passing canonical_dbg to this since it was added.

metagraph/src/cli/load/load_graph.cpp
```diff
@@ -71,5 +71,15 @@ std::shared_ptr<DeBruijnGraph> load_critical_dbg(const std::string &filename) {
     exit(1);
 }
 
+CanonicalDBG primary_to_canonical(std::shared_ptr<DeBruijnGraph> graph, size_t cache_size) {
+    if (graph->get_mode() != DeBruijnGraph::PRIMARY) {
+        logger->error("Only primary mode graphs can be wrapped into canonical mode.");
+        exit(1);
+    }
```
karasikov, line 78: this should be an assertion, which could actually happen in the constructor of CanonicalDBG itself. After this, the function becomes 2 lines. Probably not needed...

metagraph/src/common/algorithms.hpp
```diff
@@ -66,37 +66,137 @@ namespace utils {
...
+                set_op(a2_begin->begin(), a2_begin->end(),
+                       b2_begin->begin(), b2_begin->end(),
+                       std::back_inserter(merged));
+                if (merged.size()) {
+                    new_size += merged.size();
+                    *out1 = *a1_begin;
+                    ++out1;
+                    *out2 = std::move(merged);
+                    ++out2;
+                }
```
karasikov, line 153: Hm, I don't quite get what's going on. If the ranges are disjoint, you skip it in the output completely and you never know which one was skipped.

karasikov: Hm, no, I'm totally confused by the description.

metagraph/src/graph/alignment/aligner_aggregator.hpp
```diff
@@ -22,9 +26,23 @@ class PriorityDeque : public boost::container::priority_deque<Type, Sequence, Co
...
+        AlignmentCompare base_cmp_;
+    };
+
     typedef Alignment::node_index node_index;
     typedef Alignment::score_t score_t;
-    typedef PriorityDeque<Alignment, std::vector<Alignment>, AlignmentCompare> PathQueue;
+    typedef annot::binmat::BinaryMatrix::Column Column;
+    typedef PriorityDeque<std::shared_ptr<Alignment>,
+                          std::vector<std::shared_ptr<Alignment>>,
+                          SharedPtrCmp> PathQueue;
```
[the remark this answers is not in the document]
karasikov: Depends on how much time `std::move` takes for the objects, but usually avoiding pointers and storing the objects with implemented `std::move` is significantly faster. I would suggest benchmarking this.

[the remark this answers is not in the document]
hmusta: The idea behind this is to prevent keeping several copies of the same alignment if there's a path which is shared among several labels.
`AlignmentAggregator` operations haven't been a bottleneck, so I didn't think that this needed any optimization.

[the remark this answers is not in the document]
karasikov: Isn't a priority queue critical to the performance of an alignment algorithm?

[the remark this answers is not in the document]
karasikov: Hm, ok. Why don't you just push all the alignments into a vector and sort them? Why do you need a queue here at all?

metagraph/src/graph/alignment/aligner_alignment.hpp
```diff
@@ -49,7 +58,13 @@ class Alignment {
...
+    // Append other to the end of the current alignment. In this process, alignment
+    // labels are intersected. If coordinates are present, then the append is only
+    // successful if at least one coordinate of other immediately proceeds the
+    // one of the coordinates in this. If this operation is unsuccessful, then
+    // *this == {} afterwards.
+    // Returns true if the label or coordinate set of this changed.
+    bool append(Alignment&& other);
 
     size_t size() const { return nodes_.size(); }
     bool empty() const { return nodes_.empty(); }
```
karasikov, line 70: An example of confusing interface. `empty()` may return `false` even if all queues for all columns are empty.

metagraph/src/graph/alignment/dbg_aligner.cpp
```diff
@@ -38,110 +39,462 @@ ::ISeedAndExtendAligner(const DeBruijnGraph &graph, const DBGAlignerConfig &conf
...
+            align_core(*seeder, *extender, add_alignment, get_min_path_score, false);
         }
 #else
-        align_core(this_query, *seeder, *extender, add_alignment, get_min_path_score);
+        align_core(*seeder, *extender, add_alignment, get_min_path_score, false);
 #endif
 
         num_explored_nodes += extender->num_explored_nodes();
+        num_extensions += extender->num_extensions();
+        size_t num_labels = std::max(size_t { 1 }, aggregator.num_labels() - 1);
```
karasikov, line 312: Why does this have such a weird definition?

metagraph/src/graph/alignment/aligner_aggregator.hpp
```diff
@@ -62,200 +98,181 @@ class AlignmentAggregator {
...
-        return;
-
-    if (config_.chain_alignments || path_queue_.size() < config_.num_alternative_paths) {
-        path_queue_.emplace(std::move(alignment));
-    } else if (!cmp_(alignment, path_queue_.minimum())) {
-        path_queue_.update(path_queue_.begin(), std::move(alignment));
-    }
-}
+inline bool AlignmentAggregator<AlignmentCompare>::add_alignment(Alignment&& alignment) {
+    auto packaged_alignment = std::make_shared<Alignment>(std::move(alignment));
```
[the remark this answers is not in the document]
karasikov: Actually, since you pass the ownership (eventually you return a vector of objects), this works in this case.

metagraph/src/graph/alignment/aligner_aggregator.hpp
```diff
@@ -4,258 +4,229 @@
...
-inline void AlignmentAggregator<AlignmentCompare>::add_alignment(Alignment&& alignment) {
-    if (std::find(path_queue_.begin(), path_queue_.end(), alignment) != path_queue_.end())
-        return;
-
-    if (config_.chain_alignments || path_queue_.size() < config_.num_alternative_paths) {
-        path_queue_.emplace(std::move(alignment));
-    } else if (!cmp_(alignment, path_queue_.minimum())) {
-        path_queue_.update(path_queue_.begin(), std::move(alignment));
+inline bool AlignmentAggregator<AlignmentCompare>::add_alignment(Alignment&& alignment) {
+    auto a = std::make_shared<Alignment>(std::move(alignment));
```
karasikov, line 71: Please double-check this function. I didn't change the logic but did some cleanups, so now it's easier to spot issues.
It seems strange to me. For instance, sometimes you check for the presence of an equal alignment, and sometimes you don't.

metagraph/src/graph/alignment/aligner_aggregator.hpp
```diff
@@ -4,258 +4,229 @@
...
-            // first trim front of the incoming alignment
-            size_t overlap = std::min(
-                static_cast<size_t>((chain.get_cigar().data().end() - 2)->second),
-                aln.trim_query_prefix(chain_end - it->get_query().data(), graph_, config_)
-            );
+template <class AlignmentCompare>
+inline auto AlignmentAggregator<AlignmentCompare>
+::get_max_path_score(Column label) const -> score_t {
+    auto find = path_queue_.find(label);
+    return find == path_queue_.end() ? config_.min_path_score
```
karasikov, line 202: Wouldn't it make more sense to return `std::numeric_limits<score_t>::min();`?

hmusta: hmm, I guess so. I added a check later on to the extender to set the upper bound to the max of this and `min_path_score`, so here it's redundant.

metagraph/src/graph/alignment/aligner_aggregator.hpp
```diff
@@ -4,258 +4,229 @@
...
-    score_t score = chain.get_score();
-
-    bool called = false;
-    for (auto it = begin; it != end; ++it) {
-        if (it->get_offset())
-            continue;
+template <class AlignmentCompare>
+inline auto AlignmentAggregator<AlignmentCompare>
+::get_min_path_score(const Vector<Column> &labels) const -> score_t {
+    score_t global_min = get_max_path_score() * config_.rel_score_cutoff;
```
karasikov, line 156: You're using `global_min` here but not in `get_min_path_score(Column label)`, so the results for `get_min_path_score(t)` and `get_min_path_score({ t })` might be different.

Am I missing something?

metagraph/src/graph/alignment/aligner_aggregator.hpp
karasikov: Can you also add some comments to explain what is going on on each step? I'm trying to reverse-engineer the logic, but I think it has bugs if I'm understanding it correctly.

metagraph/src/graph/alignment/dbg_aligner.cpp
hmusta: fixed

metagraph/src/graph/alignment/aligner_extender_methods.hpp
```diff
@@ -96,22 +126,51 @@ class DefaultColumnExtender : public SeedFilteringExtender {
                               char /* last char of node */,
                               ssize_t /* offset (distance from start of the first node) */,
                               ssize_t /* absolute index of maximal value*/,
-                              ssize_t /* trim (starting absolute index of array) */>;
+                              ssize_t /* trim (starting absolute index of array) */,
+                              size_t /* xdrop_cutoffs_ index */,
+                              size_t /* last fork index in table */,
+                              score_t /* traversal penalty */>;
```
karasikov, line 132: Anything untrivial like this should be a struct instead of an std::tuple. Especially since you're writing a constructor separately `alloc_column(size_t size, RestArgs... args)`

karasikov (changes requested):

metagraph/src/graph/alignment/dbg_aligner.hpp
```diff
@@ -50,11 +52,15 @@ class ISeedAndExtendAligner : public IDBGAligner {
...
   protected:
-    typedef AlignmentAggregator<AlignmentCompare> Aggregator;
+    typedef std::vector<std::tuple<std::shared_ptr<ISeeder>, std::vector<node_index>,
+                                   std::shared_ptr<ISeeder>, std::vector<node_index>>> BatchSeeders;
     const DeBruijnGraph &graph_;
+    DBGAlignerConfig config_;
 
     virtual std::shared_ptr<IExtender>
-    build_extender(std::string_view query, const Aggregator &aggregator) const = 0;
+    build_extender(std::string_view query, const DBGAlignerConfig &config) const = 0;
```
karasikov, line 61: I think it makes sense to pass Seeder and Extender as template parameters to this class, so you don't go via 
interfaces ISeeder and IExtender but call them directly.

metagraph/src/graph/alignment/alignment.cpp
```diff
@@ -796,6 +1226,9 @@ bool spell_path(const DeBruijnGraph &graph,
 }
 
 bool Alignment::is_valid(const DeBruijnGraph &graph, const DBGAlignerConfig *config) const {
+    if (empty())
+        return true;
+
     std::string path;
     if (!spell_path(graph, nodes_, path, offset_)) {
```
karasikov, line 1233: Why do you reconstruct the seq from the path if you could save it beforehand?

Can you add comments to describe what `spell_path` returns (when true and when false)?

hmusta: The whole point of this function is to check if the path stored in the alignment spells the stored sequence.

hmusta: ...and when an alignment is loaded from a json

karasikov: Also, wouldn't it be easier to just call map_to_nodes_sequentially instead of doing it the other way around?

metagraph/src/graph/alignment/aligner_alignment.hpp
```diff
@@ -244,8 +236,8 @@ class QueryAlignment {
     const std::vector<Alignment>& data() const { return alignments_; }
 
   private:
-    std::shared_ptr<const std::string> query_;
-    std::shared_ptr<const std::string> query_rc_;
+    std::string query_;
+    std::string query_rc_;
```
hmusta, line 239: Doing this leads to a lot of issues. Firstly, one has to either disable the copy constructor of AlignmentResults, or do some complicated pointer re-assignment when a copy happens.

Even when this is done, string optimizations in the compiler tend to screw things up...

metagraph/src/graph/alignment/alignment.cpp
hmusta: Since alignments which come from post-alignment chaining could have gaps (i.e., `nodes_` contains `npos`), this needs a special function to spell those kinds of paths.

I refactored this so that `spell_path` returns the spelling, but throws runtime errors if it encounters anything weird. `is_valid` then tries to catch these.

metagraph/src/graph/alignment/alignment.cpp
```diff
@@ -1164,17 +1164,17 @@ void Alignment::insert_gap_prefix(ssize_t gap_length,
     offset_ = graph.get_k() - 1;
 }
 
-
-bool spell_path(const DeBruijnGraph &graph,
-                const std::vector<DeBruijnGraph::node_index> &path,
-                std::string &seq,
-                size_t offset) {
+// Return the string spelled by the path. This path may have disconnects (if it came)
+// from a chain alignment), so this method inserts
```
karasikov, line 1168: throws?

metagraph/src/graph/alignment/aligner_alignment.hpp
karasikov: maybe at least change to std::unique_ptr?

metagraph/src/graph/alignment/alignment.cpp
karasikov: Yeah, but still you could just call map_to_nodes + some postprocessing. But probably this way is also fine.

metagraph/src/graph/alignment/alignment.cpp
hmusta: I forgot to complete the comment. Fixed now.

metagraph/src/graph/alignment/aligner_alignment.hpp
hmusta: If it's a unique_ptr, then this object will no longer be copyable. The extra overhead shouldn't be an issue.

karasikov: but you don't copy it (and you don't need to), you copy a shared pointer to it.

hmusta: I ended up finding a way to get it to work with `std::string` which involved reserving the string to be just big enough to disable short-string optimizations.

metagraph/src/graph/alignment/aligner_extender_methods.hpp
karasikov: up

metagraph/src/graph/alignment/dbg_aligner.hpp
hmusta: If it was designed this way, then one would have to either move the implementations to the hpp file, or instantiate this class with every combination of Seeder and Extender types. So, I thought doing it this way instead would be a good compromise.

metagraph/src/graph/alignment/aligner_chainer.hpp
```diff
@@ -0,0 +1,39 @@
...
+// until terminate() returns true.
+std::pair<size_t, size_t>
+call_seed_chains_both_strands(std::string_view forward,
+                              std::string_view reverse,
+                              const DeBruijnGraph &graph,
+                              const DBGAlignerConfig &config,
+                              std::vector<Alignment>&& fwd_seeds,
+                              std::vector<Alignment>&& bwd_seeds,
+                              const std::function<void(Chain&&, score_t)> &callback,
+                              const std::function<bool()> &terminate = []() { return false; });
```
karasikov, line 24: exit via exception?

metagraph/src/graph/alignment/aligner_chainer.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+typedef std::vector<Alignment> Chain;
+typedef Alignment::score_t score_t;
+
+// Given sets of forward and reverse-complement seeds, construct and call chains of seeds
+// until terminate() returns true.
+std::pair<size_t, size_t>
+call_seed_chains_both_strands(std::string_view forward,
+                              std::string_view reverse,
+                              const DeBruijnGraph &graph,
```
karasikov, line 19: It seems you only need the value of `k` and not the graph itself. Am I correct?

metagraph/src/graph/alignment/aligner_chainer.cpp
```diff
@@ -0,0 +1,491 @@
...
+        size_t num_nodes_bwd;
+        std::tie(dp_tables[1], num_seeds_bwd, num_nodes_bwd)
+            = chain_seeds(config, reverse, both_seeds[1]);
+        num_seeds += num_seeds_bwd;
+        num_nodes += num_nodes_bwd;
+    };
+
+    chain_fwd();
+    chain_bwd();
+
```
karasikov, line 84: Why do you create a function instead of simply just execute that code

metagraph/src/graph/alignment/aligner_chainer.cpp
```diff
@@ -0,0 +1,491 @@
...
+                coord_union.insert(coord_union.end(), a_c_begin, last_chain[i].label_coordinates.end());
+                coord_union.insert(coord_union.end(), b_c_begin, chain[i].label_coordinates.end());
+                std::swap(last_chain[i].label_columns, label_union);
+                std::swap(last_chain[i].label_coordinates, coord_union);
+            }
+        } else {
+            callback(std::move(last_chain), last_chain_score);
+            std::swap(last_chain, chain);
+            last_chain_score = chain_score;
+        }
```
karasikov, line 212:
```suggestion
        if (last_chain != chain) {
            callback(std::move(last_chain), last_chain_score);
            std::swap(last_chain, chain);
            last_chain_score = chain_score;
            continue;
        }

        // if this chain has the same seeds as the last one, merge their coordinate sets
        for (size_t i = 0; i < chain.size(); ++i) {
            Alignment::LabelSet label_union;
            Alignment::CoordinateSet coord_union;
            auto a_begin = last_chain[i].label_columns.begin();
            auto a_end = last_chain[i].label_columns.end();
            auto a_c_begin = last_chain[i].label_coordinates.begin();
            auto b_begin = chain[i].label_columns.begin();
            auto b_end = chain[i].label_columns.end();
            auto b_c_begin = chain[i].label_coordinates.begin();
            while (a_begin != a_end && b_begin != b_end) {
                if (*a_begin < *b_begin) {
                    label_union.push_back(*a_begin);
                    coord_union.push_back(*a_c_begin);
                    ++a_begin;
                    ++a_c_begin;
                } else if (*a_begin > *b_begin) {
                    label_union.push_back(*b_begin);
                    coord_union.push_back(*b_c_begin);
                    ++b_begin;
                    ++b_c_begin;
                } else {
                    label_union.push_back(*a_begin);
                    coord_union.emplace_back();
                    std::set_union(a_c_begin->begin(), a_c_begin->end(),
                                   b_c_begin->begin(), b_c_begin->end(),
                                   std::back_inserter(coord_union.back()));
                    ++a_begin;
                    ++a_c_begin;
                    ++b_begin;
                    ++b_c_begin;
                }
            }
            label_union.insert(label_union.end(), a_begin, a_end);
            label_union.insert(label_union.end(), b_begin, b_end);
            coord_union.insert(coord_union.end(), a_c_begin, last_chain[i].label_coordinates.end());
            coord_union.insert(coord_union.end(), b_c_begin, chain[i].label_coordinates.end());
            std::swap(last_chain[i].label_columns, label_union);
            std::swap(last_chain[i].label_coordinates, coord_union);
        }
```

karasikov (changes requested):

metagraph/src/graph/alignment/aligner_extender_methods.hpp
hmusta: done

metagraph/src/graph/alignment/aligner_chainer.hpp
hmusta: yea, it could also work here

metagraph/src/graph/alignment/aligner_config.hpp
```diff
@@ -66,16 +68,27 @@ class DBGAlignerConfig {
     score_t min_path_score = 0;
     score_t xdrop = std::numeric_limits<score_t>::max();
 
+    DeBruijnGraph::node_index terminal_node = DeBruijnGraph::npos;
```
karasikov, line 71: So you keep this in the config, but not the sequence. This is weird.

karasikov: I would recommend to keep sequence here as well and not pass it as a separate input parameter, or pass this terminal node as a separate parameter, similarly to `query`

hmusta: I'll add a comment describing this. The point is that I want to define an endpoint for an alignment by setting a node and a target traversal distance at which to stop.

metagraph/src/graph/alignment/aligner_chainer.hpp
hmusta: In the release-build code, yes. In debug builds, the graph is also used to check that seeds are still valid after merging them.

metagraph/src/graph/alignment/aligner_config.hpp
karasikov: I got that. The thing I don't get is why you want to keep this and the sequence itself separately.
I could understand if you keep alignment parameters in a struct separately from the objects (sequence + node) or just keep all these things in one struct and call it `AlingmentContext`, or something.

metagraph/src/graph/alignment/aligner_chainer.hpp
karasikov: Maybe you can check it outside this function?

hmusta: there are a few places where the graph is only used to access k. I've changed these.

metagraph/src/graph/alignment/aligner_config.hpp
hmusta: I changed this so now target_distance and target_node are passed as arguments to extend.

metagraph/src/graph/alignment/aligner_extender_methods.cpp
```diff
@@ -19,6 +20,12 @@ constexpr score_t ninf = DBGAlignerConfig::ninf;
 constexpr size_t kPadding = 5;
 
 
+DefaultColumnExtender DefaultColumnExtender::make(const IDBGAligner &aligner,
+                                                  const DBGAlignerConfig &config,
```
karasikov, line 24: Ideally, it should only take `const IDBGAligner &aligner` and extract the config from it.

metagraph/src/graph/alignment/aligner_labeled.hpp
```diff
@@ -219,68 +224,39 @@ class LabeledExtender : public DefaultColumnExtender {
...
-          : ILabeledAligner<AlignmentCompare>(std::forward<Args>(args)...) {
-        if (this->labeled_graph_.get_coordinate_matrix()
-                && std::is_same_v<Extender, LabeledExtender>) {
-            // do not use a global xdrop cutoff since we need separate cutoffs
-            // for each label
-            this->config_.global_xdrop = false;
-        }
-    }
+    typedef typename DBGAligner<Seeder, Extender, AlignmentCompare>::BatchSeeders BatchSeeders;
+    mutable AnnotationBuffer labeled_graph_;
```
karasikov, line 252: So, "annotation buffer" or "labeled graph"?

metagraph/src/graph/alignment/aligner_labeled.hpp
```diff
@@ -0,0 +1,289 @@
...
+    typedef AnnotationBuffer::LabelSet LabelSet;
+    typedef Alignment::node_index node_index;
+    typedef Alignment::Column Column;
+
+    virtual ~ILabeledAligner() {
+        common::logger->trace("Buffered {}/{} nodes and {} label combinations",
+                              labeled_graph_.num_nodes_buffered(),
+                              this->graph_.num_nodes(),
+                              labeled_graph_.num_label_sets());
+    }
```
karasikov, line 246: move to .cpp

hmusta:
> I tried to use `LabeledAligner<UniMEMSeeder>` instead of `LabeledAligner<SuffixSeeder<UniMEMSeeder>>` and it segfaulted. Any idea why? I thought it's supposed to work with any seeder.

It should work now. There was an issue with a commit in this branch where the result of `map_to_nodes` was being destroyed before it was used.

metagraph/src/graph/alignment/aligner_labeled.hpp
hmusta: I'll stick with `annotation_buffer`

metagraph/src/graph/alignment/alignment.cpp
```diff
@@ -489,11 +871,10 @@ Json::Value Alignment::path_json(size_t node_size, std::string_view label) const
 }
 
 Json::Value Alignment::to_json(std::string_view full_query,
-                               const DeBruijnGraph &graph,
+                               size_t node_size,
```
karasikov, line 874: Isn't node_size a property of the alignment, so it should be stored there?

metagraph/src/graph/alignment/alignment.cpp
```diff
@@ -489,11 +871,10 @@ Json::Value Alignment::path_json(size_t node_size, std::string_view label) const
...
                                bool is_secondary,
                                std::string_view read_name,
                                std::string_view label) const {
```
karasikov, line 877: Not clear what is `full_query` and how it relates to `query_`.

karasikov (changes requested):

metagraph/src/graph/alignment/alignment.cpp
hmusta: `full_query` can actually be reconstructed from the alignment, so I changed this function

metagraph/src/graph/alignment/alignment.cpp
hmusta: It's a property of the graph to which the alignment was done. If we stick to the assumption that the alignment is done against a fixed-order de Bruijn graph, then node_size can be inferred from the stored sequence, the path, and the alignment's first-node offset.

metagraph/src/graph/alignment/aligner_chainer.hpp
```diff
@@ -0,0 +1,38 @@
...
+typedef std::vector<Alignment> Chain;
+typedef Alignment::score_t score_t;
+
+// Given sets of forward and reverse-complement seeds, construct and call chains of seeds
+// until terminate() returns true.
+std::pair<size_t, size_t>
+call_seed_chains_both_strands(std::string_view forward,
+                              std::string_view reverse,
+                              size_t node_overlap,
+                              const DBGAlignerConfig &config,
```
karasikov, line 20: Shouldn't `node_overlap` be a member of `DBGAlignerConfig`?

hmusta: I mean, one can do that, but I don't see how that's semantically more correct than what's happening here.

metagraph/src/graph/alignment/aligner_extender_methods.cpp
hmusta: done. This basically involved moving a bunch of stuff out of DBGAlignerConfig and passing them as arguments to `Extender::extend` instead

metagraph/src/graph/alignment/aligner_chainer.cpp
```diff
@@ -0,0 +1,474 @@
...
+                               typename std::vector<Alignment>::iterator begin,
+                               typename std::vector<Alignment>::iterator end,
+                               std::vector<score_t> &best_score,
+                               const std::function<void(Alignment&&)> &callback) {
+    assert(begin <= end);
+    assert(chain.size());
+
+    const char *chain_begin = chain.get_query().data();
+    const char *chain_end = chain_begin + chain.get_query().size();
+    if (begin == end || chain_end == query.data() + query.size()) {
```
karasikov, line 381: What's the difference between `query` and `chain.get_query()`?

hmusta: I haven't gotten around to refactoring this yet, I want to rename `get_query` to `get_query_view`. `Alignment::get_query` returns the window in the query to which the path aligned. For precisely,
```c++
query.data() == chain.get_query().data() - chain.get_clipping();
query.data() + query.size() == chain.get_query().data() + chain.get_query().size() + chain.get_end_clipping()
```

metagraph/src/graph/alignment/aligner_chainer.cpp
```diff
@@ -0,0 +1,474 @@
...
+            Alignment next_chain(chain);
+            next_chain.trim_end_clipping();
+            bool modified = next_chain.append(std::move(aln));
+            assert(next_chain.get_score() == next_score);
+            if (next_chain.size()) {
+                called |= modified;
+                construct_alignment_chain<AlignmentCompare>(
+                    node_overlap, config, query, std::move(next_chain),
+                    it + 1, end, best_score, callback
+                );
```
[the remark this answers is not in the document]
hmusta: This can be rewritten in a way that doesn't involve recursion, but it's a bit more complicated. The `best_score` vector ensures that a lot of the calls get cut off early, but I suppose it's possible for the depth to be exceeded in a very bad case.

metagraph/src/graph/alignment/aligner_chainer.cpp
karasikov: maybe `get_path_window()`, or something?

karasikov: if you have an idea, please go ahead and rename

karasikov (approved):

PR #374, opened 2022-01-22
query expressions (counts) in labels for individual k-mers in query

karasikov:
* Support for querying abundances of individual k-mers
* Other refactorings and improvements
* Support for printing query results with CLI in JSON

Example:
```
./metagraph build -k 31 -o test_1 ../tests/data/transcripts_1000.fa
./metagraph annotate --anno-header -o test_1 -i test_1.dbg ../tests/data/transcripts_1000.fa --count-kmers
./metagraph transform_anno --anno-type int_brwt test_1.column.annodbg -v -o test_1
./metagraph query -i test_1.dbg  -a test_1.int_brwt.annodbg ../tests/data/transcripts_100.fa --query-counts --suppress-unlabeled | less
./metagraph query -i test_1.dbg  -a test_1.int_brwt.annodbg ../tests/data/transcripts_100.fa --query-counts --suppress-unlabeled --json | less
```

Output format: `POS_FIRST-POS_LAST=COUNT` for a run of k-mers in query at positions [POS_FIRST, POS_LAST] with abundance `COUNT`, or `POS=COUNT` for isolated k-mer matches. Skipped k-mers have count 0.

Example:
```
{
        "results" : 
        [seq_description" : "ENST00000450305.2..."
                {
                        "kmer_abundances" : 
                        [
                                "0-1626=1"
                        ],
                        "kmer_count" : 1627,
                        "sample" : "ENST00000456328.2..."
                }
        ],
        "seq_description" : "ENST00000456328.2..."
}
{
        "results" : 
        [
                {
                        "kmer_abundances" : 
                        [
                                "0-601=1"
                        ],
                        "kmer_count" : 602,
                        "sample" : "ENST00000450305.2..."
                },
                {
                        "kmer_abundances" : 
                        [
                                "0-17=1",
                                "48-151=1",
                                "258-383=1",
                                "414-601=1"
                        ],
                        "kmer_count" : 436,
                        "sample" : "ENST00000456328.2..."
                },
                {
                        "kmer_abundances" : 
                        [
                                "68-310=1",
                                "367-383=1",
                                "423-586=1"
                        ],
                        "kmer_count" : 424,
                        "sample" : "ENST00000624431.2..."
                }
        ],
        "seq_description" : "ENST00000450305.2..."
}
...
```

TODO: add docs

### DIFF
--- a/metagraph/docs/source/api.rst
+++ b/metagraph/docs/source/api.rst
@@ -17,17 +17,84 @@ Install MetaGraph API in Python::
 
 Sequence search
 ---------------
-The Python client has a ``search`` method which allows querying an index running on a server.
+The Python client has a ``search`` method for querying an index running on a server.
 The method accepts a single sequence or list of sequences represented with strings.
+
+.. py:function:: metagraph.client.GraphClient.search(self, ...)
+
+        :param      sequence:             Query sequence
+        :type       sequence:             Union[str, Iterable[str]]
+        :param      top_labels:           The maximum number of matched labels to retrieve [default: 100]
+        :type       top_labels:           int
+        :param      discovery_threshold:  The minimum fraction (between 0.0 and 1.0) of k-mers from the query required to match a label (occur in a sample) in order for that label to show up in the result [default: 0.0]
+        :type       discovery_threshold:  float
+        :param      with_signature:       Return the signature of k-mer matches
+        :type       with_signature:       bool
+        :param      abundance_sum:        Compute the sum of abundances for all k-mers matched
+        :type       abundance_sum:        bool
+        :param      query_counts:         Query k-mer counts
+        :type       query_counts:         bool
+        :param      query_coords:         Query k-mer coordinates
+        :type       query_coords:         bool
+        :param      align:                Align the query sequence to the joint graph and query labels for that alignment instead of the original sequence
+        :type       align:                bool
+        :param      align_params:         The parameters for alignment (see method align())
+        :type       align_params:         dictionary
+        :return:                          A data frame with query results
+        :rtype:                           pandas.DataFrame
+
+``GraphClient`` will return an instance of ``pandas.DataFrame`` storing the result.
+You can also use the lower level ``GraphClientJson`` to instead receive a JSON response.
+
+
+Search with alignment
+^^^^^^^^^^^^^^^^^^^^^
+It is possible to first align a sequence to the joint graph and use the aligned sequence to query the index.
+This can be done by setting ``align=True`` (default False).
+If the ``align`` flag is set, all the alignment options (explained in the :ref:`Sequence alignment <alignment>` section below) are accepted::
+
+    metasub.search(query, discovery_threshold=0.0, top_labels=200,
+                   align=True, min_exact_match=0.8, max_num_nodes_per_seq_char=10.0)
+
+Querying k-mer abundance and coordinates
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Metagraph supports k-mer count- and coordinate-aware indexes (see :ref:`Indexing k-mer counts <indexing counts>`
+and :ref:`Indexing k-mer coordinates <indexing coordinates>`).
+If you are querying such an index,
+you can also set ``abundance_sum=True``, ``query_counts=True``, or ``query_coords=True``. If you try to query an
+index which does not support either of these query types, the server will return an error.
+
+
+.. _alignment:
+
+Sequence alignment
+------------------
+The ``align`` method allows alignment of sequences to the the graph.
+The method accepts a single sequence or list of sequences represented with strings.
+Additionally, the method accepts the following keyword arguments:
+
+.. py:function:: metagraph.client.GraphClient.align(self, ...)
+
+        Align sequence(s) to the joint graph
+
+        :param      sequence:                    The query sequence
+        :type       sequence:                    Union[str, Iterable[str]]
+        :param      min_exact_match:             The minimum fraction (between 0.0 and 1.0) of matching nucleotides required to align sequence [default: 0]
+        :type       min_exact_match:             float
+        :param      max_alternative_alignments:  The number of different alignments to return [default: 1]
+        :type       max_alternative_alignments:  int
+        :param      max_num_nodes_per_seq_char:  The maximum number of nodes to consider per sequence character during extension [default: 10.0]
+        :type       max_num_nodes_per_seq_char:  float
+
+        :returns:   A data frame with alignments
+        :rtype:     pandas.DataFrame
+
+
+Examples
+--------
 
 Example of search in MetaSUB
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-For basic usage, the method accepts the following keyword arguments:
-
-- ``top_labels``: the maximum number of matched labels to retrieve [default: 100]
-- ``discovery_fraction``: a value between 0.0 and 1.0 specifying the minimum fraction of
-  matching k-mers needed to match a label [default: 0.0]
-
 ::
 
     from metagraph.client import GraphClient
@@ -37,7 +104,7 @@ For basic usage, the method accepts the following keyword arguments:
     lbls = metasub.column_labels()
 
     # >ENA|A14565|A14565.1 16S rRNA
-    query= 'TCGAACGGTAACAGGAAGAAGCTTGCTTCTTTGCTGACGAGTGGCGGACGGGTGAGTAAT\
+    query = 'TCGAACGGTAACAGGAAGAAGCTTGCTTCTTTGCTGACGAGTGGCGGACGGGTGAGTAAT\
             GTCTGGGAAACTGCCTGATGGAGGGGGATAACTACTGGAAACGGTAGCTAATACCGCATA\
             ACGTCGCAAGACCAAAGAGGGGGACCTTCGGGCCTCTTGCCATCGGATGTGCCCAGATGG\
             GATTAGCTAGTAGGTGGGGTAACGGCTCACCTAGGCGACGATCCCTAGCTGGTCTGAGAG\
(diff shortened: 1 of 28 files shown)


metagraph/docs/source/api.rst
```diff
@@ -80,7 +80,7 @@ If the ``align`` flag is set, all the alignment options (explained in the alignm
 Searching count and coordinate indexes
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 Metagraph supports k-mer count-aware and coordinate-aware indexes. If you are querying such an index,
-you can also set ``query_counts=True`` or ``query_coords=True``, respectively. If you try to query an
+you can also set ``weighted_with_counts=True`` or ``query_coords=True``, respectively. If you try to query an
```
hmusta, line 83: Maybe it's better to say "abundances" here instead of "weights", since that's what these are measuring. I guess "weights" only makes sense in the context of some objective function.

metagraph/src/cli/config/config.hpp
```diff
@@ -40,8 +40,9 @@ class Config {
     bool count_kmers = false;
     bool print_signature = false;
     bool query_presence = false;
+    bool query_counts = false;
     bool query_coords = false;
-    bool verbose_coords = false;
+    bool verbose_output = false;
```
hmusta, line 45: I think `verbose_coords` was more descriptive, `--verbose-output` may be confused with `--verbose`

metagraph/src/cli/config/config.cpp
```diff
@@ -1241,16 +1243,18 @@ if (advanced) {
...
             fprintf(stderr, "\n");
             fprintf(stderr, "\t   --count-labels \t\tcount labels for k-mers from querying sequences [off]\n");
             fprintf(stderr, "\t   --count-kmers \t\tweight k-mers with their annotated counts (requires count annotation) [off]\n");
             fprintf(stderr, "\t   --count-quantiles [FLOAT ...] \tk-mer count quantiles to compute for each label [off]\n"
                             "\t                                 \t\tExample: --count-quantiles '0 0.33 0.5 0.66 1'\n"
                             "\t                                 \t\t(0 corresponds to MIN, 1 corresponds to MAX)\n");
+            fprintf(stderr, "\t   --query-counts \t\tquery k-mer counts (requires count or coord annotation) [off]\n");
             fprintf(stderr, "\t   --query-coords \t\tquery k-mer coordinates (requires coord annotation) [off]\n");
-            fprintf(stderr, "\t   --verbose-coords \t\tdo not collapse continuous matched coord ranges in result (for query-coords) [off]");
+            fprintf(stderr, "\t   --verbose-output \t\tdo not collapse continuous coord or count ranges (for query-coords and query-counts) [off]\n");
```
hmusta, line 1255:
```suggestion
            fprintf(stderr, "\t   --verbose-output \t\tdo not collapse contiguous coord or count ranges (for query-coords and query-counts) [off]\n");
```

metagraph/src/graph/annotated_dbg.hpp
```diff
@@ -80,6 +80,8 @@ class AnnotatedDBG : public AnnotatedSequenceGraph {
 
     bool check_compatibility() const;
 
+    std::vector<node_index> sequence_to_path(std::string_view sequence) const;
```
hmusta, line 83: Since this calls `map_to_nodes`, it doesn't exactly map to paths. How about `sequence_to_labeled_nodes`?

metagraph/src/graph/annotated_dbg.hpp
```diff
@@ -134,6 +136,18 @@ class AnnotatedDBG : public AnnotatedSequenceGraph {
...
                               const std::vector<double> &count_quantiles) const;
 
+    std::vector<std::pair<Label, std::vector<size_t>>>
+    get_kmer_counts(std::string_view sequence,
+                    size_t num_top_labels,
+                    double discovery_fraction,
+                    double presence_fraction) const;
+
+    std::vector<std::pair<Label, std::vector<size_t>>>
+    get_kmer_counts(const std::vector<node_index> &path,
```
hmusta, line 146: since this may not be a path, how about `nodes`?

metagraph/src/cli/server.cpp
```diff
@@ -67,7 +67,7 @@ std::string process_search_request(const std::string &received_message,
     config.fast = json.get("fast", config.fast).asBool();
     config.print_signature = json.get("with_signature", config.print_signature).asBool();
     config.query_coords = json.get("query_coords", config.query_coords).asBool();
-    config.count_kmers = json.get("query_counts", config.count_kmers).asBool();
+    config.count_kmers = json.get("weighted_with_counts", config.count_kmers).asBool();
```
hmusta, line 70: see above

metagraph/src/cli/server.cpp
```diff
@@ -128,7 +128,7 @@ std::string process_search_request(const std::string &received_message,
     // Create full JSON object
     Json::Value search_response(Json::arrayValue);
     for (const auto &seq_result : search_results) {
-        search_response.append(seq_result.to_json(config.verbose_coords, anno_graph));
+        search_response.append(seq_result.to_json(config.verbose_output, anno_graph));
```
hmusta, line 131: see above

metagraph/src/cli/query.cpp
```diff
@@ -166,96 +166,99 @@ Json::Value get_label_as_json(const std::string &label) {
...
                     coord_array.append(range);
                 }
             }
-
-
-            label_obj[KMER_COORDINATE_FIELD] = coord_array;
-            root["results"].append(label_obj);
         }
-    } else {
-        assert(false);
```
hmusta, line 257: why not keep this as a kind of future-proofing?

metagraph/src/cli/query.cpp
```diff
@@ -166,96 +166,99 @@ Json::Value get_label_as_json(const std::string &label) {
...
+                }
+            }
         }
-    } else if (std::holds_alternative<LabelCoordVec>(result)) {
+    } else {
         // Kmer coordinates
-        for (const auto &[label, tuples] : std::get<LabelCoordVec>(result)) {
-            Json::Value label_obj = get_label_as_json(label);
-            Json::Value coord_array = Json::arrayValue;
+        for (const auto &[label, tuples] : std::get<LabelCoordVec>(result_)) {
```
hmusta, line 247: there should at least be an assertion here that `std::get<LabelCoordVec>(result_)` returns a non-`nullptr`, otherwise this'll segfault if we add a new type of annotation later

metagraph/src/cli/query.cpp
```diff
@@ -264,65 +267,86 @@ Json::Value SeqSearchResult::to_json(bool verbose_coords,
...
+                        last.second = counts[i];
+                    }
+                }
+            }
         }
-    } else if (std::holds_alternative<LabelCoordVec>(result)) {
+    } else {
         // Kmer coordinates
-        for (const auto &[label, tuples] : std::get<LabelCoordVec>(result)) {
+        for (const auto &[label, tuples] : std::get<LabelCoordVec>(result_)) {
```
hmusta, line 339: again, if we add a new annotation type without remembering to change this, this'll segfault

hmusta (changes requested):

metagraph/docs/source/api.rst
karasikov: The thing is that this mode actually does not return the counts/abundances. It returns the sum of all abundances for all k-mers from the query matched to the label. So, it's a weighted sum, similar to the simple query with --count-labels, where each match is weighted with the k-mer's abundance.

karasikov: The other mode (added in this PR) for querying individual abundances/counts is triggered by `--query-counts`, which was previously triggering this weighted sum thing.

karasikov: Maybe we should rename this to `--abundance-sum` or something like that? What's your suggestion?

metagraph/src/cli/config/config.hpp
karasikov: Now it's both `verbose_coords` and `verbose_counts`, so I had to rename it. Not a major thing anyway.

metagraph/src/cli/config/config.cpp
karasikov: Isn't 'continuous' more appropriate in the context of counts?

metagraph/src/cli/query.cpp
karasikov: `std::get<LabelCoordVec>` returns a reference or throws an exception.

metagraph/src/cli/query.cpp
karasikov: no, it will throw, so you'll notice

metagraph/src/graph/annotated_dbg.hpp
karasikov: renamed to `sequence_to_lnodes`

metagraph/src/cli/config/config.cpp
hmusta: "Contiguous" would be more correct for coordinates, but I agree that it doesn't work for counts. For counts, we could say "runs" to refer to ranges of matching values.

metagraph/docs/source/api.rst
hmusta: I like that, it's more descriptive.

metagraph/src/cli/config/config.cpp
karasikov: I think it's better to deliver the intuition right than to be extremely accurate.

Among these two, continuous seems to be a better option to me.

metagraph/src/graph/annotated_dbg.hpp
```diff
@@ -129,33 +127,33 @@ class AnnotatedDBG : public AnnotatedSequenceGraph {
                    size_t min_count = 0,
                    bool with_kmer_counts = false) const;
 
-    std::vector<std::pair<Label, std::vector<size_t>>>
+    std::vector<std::tuple<Label, size_t, std::vector<size_t>>>
```
hmusta, line 130: maybe add comments to describe the relationship between `size_t` and `std::vector<size_t>` represent

metagraph/integration_tests/test_api.py
```diff
@@ -248,39 +247,46 @@ def setUpClass(cls):
         cls.graph_client = MultiGraphClient()
         cls.graph_client.add_graph(cls.host, cls.port, cls.graph_name)
 
+        cls.sample_query_expected_rows = 98 if cls.mode == 'basic' else 99
```
hmusta, line 250: Add a comment describing why this value depends on the case. Shouldn't the number of lines of output equal the number of lines in the query (and thus not depend on the graph mode)?

hmusta (approved):

metagraph/integration_tests/test_api.py
karasikov: the canonical graph has more k-mers, so it finds more k-mer matches

PR #383, opened 2022-02-04
Fix asserts when aligning an empty sequence

### LINKED ISSUE #369: Assertion fails when aligning empty sequence
**Error:** When aligning an empty sequence, an assertion fails and the metagraph process crashes. This is problematic for metagraph server_query processes, as aligning with `client.align("")` to a public graph API will crash the underlying server process.

The following assertion error is given:
```
Assertion failed: (config_.get_row(query_.front())[query_.front()] == partial_sum_[1]), function ExactSeeder, file aligner_seeder_methods.cpp, line 33.
[1]    7984 abort      ./metagraph align --json -i t100.dbg test_align.fq
```

**Steps to reproduce with CLI:**
1. Generate a FastQ file representing an empty sequence
```
@TestSeq

```
2. `./metagraph align --json -i <graph> <fastq_file>`

**Steps to reproduce with python API:**
1. `./metagraph server_query -i <graph> -a <anno> --port 5555`
2. In python:
```
from metagraph.client import *

graph = GraphClientJson('127.0.0.1', 5555, api_path='')
graph.align("")
```

hmusta:
Fixes #369


karasikov:
merge?

PR #384, opened 2022-02-07
Support outputting alignments in GAF format

hmusta:
This format reports the node IDs of alignment paths.

See ["The Graph Alignment Format (GAF)"](https://github.com/lh3/gfatools/blob/master/doc/rGFA.md#the-graph-alignment-format-gaf) and ["Sequence Alignment/Map Optional Fields Specification"](https://samtools.github.io/hts-specs/SAMtags.pdf)

### DIFF
--- a/metagraph/src/cli/align.cpp
+++ b/metagraph/src/cli/align.cpp
@@ -287,18 +287,7 @@ std::string format_alignment(std::string_view header,
                              const DeBruijnGraph &graph,
                              const Config &config) {
     std::string sout;
-    if (!config.output_json) {
-        sout += fmt::format("{}\t{}", header, paths.get_query());
-        if (paths.empty()) {
-            sout += fmt::format("\t*\t*\t{}\t*\t*\t*", config.alignment_min_path_score);
-        } else {
-            for (const auto &path : paths.data()) {
-                sout += fmt::format("\t{}", path);
-            }
-        }
-
-        sout += "\n";
-    } else {
+    if (config.output_json) {
         Json::StreamWriterBuilder builder;
         builder["indentation"] = "";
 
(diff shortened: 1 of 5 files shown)


metagraph/src/graph/alignment/aligner_cigar.cpp
```diff
@@ -93,6 +93,46 @@ std::string Cigar::to_string() const {
...
+            case MATCH: {
+                match_count += num;
+                ref_it += num;
+            } break;
+            case MISMATCH: {
+                if (match_count) {
+                    md_string += std::to_string(match_count);
+                    match_count = 0;
+                }
+                md_string += std::string(ref_it, ref_it + num);
```
karasikov, line 115: `string_view` would be enough. Why do you create a new string?

metagraph/src/cli/align.cpp
```diff
@@ -319,6 +308,58 @@ std::string format_alignment(std::string_view header,
 
             sout += fmt::format("{}\n", Json::writeString(builder, json_line));
         }
+    } else if (config.output_gaf) {
+        std::string base = fmt::format("{}\t{}\t", header, paths.get_query().size());
+        for (size_t i = 0; i < paths.size(); ++i) {
+            const auto &path = paths[i];
```
karasikov, line 314:
```suggestion
        for (const auto &path : paths) {
```

karasikov (changes requested):
Please add integration / unit tests.

metagraph/src/cli/align.cpp
hmusta: I had it set up this way because a bug in GCC-8 was causing the compilation to crash because of this kind of for-loop. Let's see if it happens this time.

PR #386, opened 2022-02-18
Added NodeRC for indexing reverse complements of prefixes/suffixes in PRIMARY graphs

hmusta:
Given a node, this GraphExtension maps it to the BOSS nodes corresponding to the reverse complements of its k-1 prefix and k-1 suffix, acting as a helper for traversal in CanonicalDBG. As long as such nodes only exist for <10% of the graph, this index will be smaller than the small representation of the graph.

The benchmarks show that this speeds up traversal in CanonicalDBG by about 2.3x


metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -256,101 +173,65 @@ ::call_outgoing_kmers(node_index node, const OutgoingEdgeCallback &callback) con
...
+
+        const DBGSuccinct *dbg_succ = get_dbg_succ(*graph_);
+        const boss::BOSS *boss = dbg_succ ? &dbg_succ->get_boss() : nullptr;
+        const auto cache = get_extension<NodeFirstCache>();
 
-            boss::BOSS::TAlphabet c = boss.get_W(adjacent_edge) % boss.alph_size;
+        get_extension<NodeRC<>>()->call_incoming_nodes_from_rc(node, [&](node_index next) {
+            char c;
+            if (cache) {
+                c = cache->get_first_char(next);
```
karasikov, line 199: So is it helpful to have both cache and NodeRC? Are they doing different things?

metagraph/src/graph/graph_extensions/node_rc.hpp
```diff
@@ -0,0 +1,61 @@
...
+    // Construct a NodeRC index
+    NodeRC(const DBGSuccinct &graph);
+
+    void set_graph(const DeBruijnGraph &graph) {
+        assert(!rc_.size() || is_compatible(graph));
+        graph_ = &graph;
+    }
+
+    void call_outgoing_nodes_from_rc(node_index node, const std::function<void(node_index)> &callback) const;
+    void call_incoming_nodes_from_rc(node_index node, const std::function<void(node_index)> &callback) const;
```
karasikov, line 37:
```suggestion
    void call_outgoing_from_rc(node_index node, const std::function<void(node_index)> &callback) const;
    void call_incoming_from_rc(node_index node, const std::function<void(node_index)> &callback) const;
```

metagraph/src/graph/graph_extensions/node_rc.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+    // Indicates whether a reverse complement prefix/suffix exists for each graph node.
+    Indicator rc_;
+
+    // Vector storing the indices of the reverse complements of the node prefix and
+    // suffix (if they exist) in an interleaved form.
+    // mapping_.size() == rc_->num_set_bits() * 2
+    Mapping mapping_;
+
+    static constexpr auto kRCExtension = ".rc";
```
karasikov, line 55:
```suggestion
    static constexpr auto kRCExtension = ".rc_adj";
```

metagraph/src/graph/graph_extensions/node_rc.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+class DBGSuccinct;
+
+// Maps each node in a PRIMARY-mode DeBruijnGraph to nodes adjacent to its reverse
+// complement. When using the index construct constructor on DBGSuccinct, or when
+// loading such an index, this stores a map from each DBGSuccinct to the BOSS nodes
+// corresponding to the reverse complements of its k-1 prefix and k-1 suffix.
+template <class Indicator = bit_vector_smart,
+          class Mapping = sdsl::dac_vector_dp<sdsl::rrr_vector<>>>
+class NodeRC : public SequenceGraph::GraphExtension {
```
karasikov, line 21: I don't see how having these flexible template types helps. You're not going to change them. Pick the best option and always use that.

metagraph/src/graph/graph_extensions/node_rc.cpp
```diff
@@ -0,0 +1,304 @@
...
+                                               bool verbose) const {
+    if (!rc_.size()) {
+        if (const auto *dbg = dynamic_cast<const DeBruijnGraph*>(&graph))
+            return *dbg == *graph_;
+
+        if (verbose)
+            std::cerr << "ERROR: only compatible with DeBruijnGraph" << std::endl;
+
+        return false;
+    }
```
karasikov, line 281: I don't see much sense in this check. If `rc_` size is zero, it's not going to be used, so it shouldn't matter what this function returns.

metagraph/src/graph/graph_extensions/node_rc.cpp
```diff
@@ -0,0 +1,304 @@
...
+    } catch (...) {
+        std::cerr << "ERROR: Cannot load graph RC from file " << rc_filename << std::endl;
+        return false;
+    }
+}
+
+template <class Indicator, class Mapping>
+void NodeRC<Indicator, Mapping>::serialize(const std::string &filename_base) const {
+    if (!rc_.size())
+        std::cerr << "WARNING: NodeRC was initialized with set_graph, so nothing to serialize." << std::endl;
```
karasikov, line 261: use logger?

metagraph/src/graph/graph_extensions/node_rc.cpp
karasikov:
```suggestion
    if (!rc_.size())
        return false;
```

metagraph/src/graph/graph_extensions/node_rc.cpp
```diff
@@ -0,0 +1,304 @@
...
+                if (found) {
+                    assert(mapping[map_i].first == path[i]);
+                    assert(mapping[map_i].second.first);
+                    assert(!mapping[map_i].second.second);
+                    mapping[map_i].second.second = edge;
+                } else {
+                    builder->add_one(path[i]);
+                    mapping.emplace_back(path[i], std::make_pair(0, edge));
+                }
+            }
```
karasikov, line 75: add more comments

metagraph/src/cli/config/config.cpp
```diff
@@ -340,6 +340,8 @@ Config::Config(int argc, char *argv[]) {
             enumerate_out_sequences = true;
         } else if (!strcmp(argv[i], "--to-gfa")) {
             to_gfa = true;
+        } else if (!strcmp(argv[i], "--noderc")) {
```
karasikov, line 343: `--adj-rc`?

karasikov (changes requested):

metagraph/src/graph/representation/canonical_dbg.cpp
hmusta: Sure. In CanonicalDBG, having the cache is still useful for traversing backwards in the graph, whereas this index is useful for quickly determining if a k-mer orientation switch can be done.

metagraph/src/graph/graph_extensions/node_rc.hpp
hmusta: I still want to keep this flexibility in case I decide to play around with different compression schemes, but it'll probably be useful to rename this to something like TNodeRC and make NodeRC a typedef with these types

metagraph/src/graph/graph_extensions/node_rc.cpp
hmusta: I guess the only use of this function in this case is to check to make sure the correct pointer was set, so this can be simplified

metagraph/src/graph/graph_extensions/node_rc.cpp
```diff
@@ -0,0 +1,304 @@
...
+
+    common::logger->trace("Generating reverse complement indicator vector");
+    auto initialization_data = builder->get_initialization_data();
+    assert(initialization_data.num_set_bits == mapping.size());
+    rc_ = Indicator(initialization_data.call_ones,
+                    initialization_data.size,
+                    initialization_data.num_set_bits);
+    builder.reset();
+
+    common::logger->trace("Constructing reverse complement map");
```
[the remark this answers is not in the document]
hmusta: done

metagraph/src/graph/graph_extensions/node_rc.cpp
hmusta: I cleaned this up a lot, and added comments

metagraph/src/cli/align.cpp
```diff
@@ -300,8 +301,13 @@ int align_to_graph(Config *config) {
     // For graphs which still feature a mask, this speeds up mapping and allows
     // for dummy nodes to be matched by suffix seeding
     auto dbg_succ = std::dynamic_pointer_cast<DBGSuccinct>(graph);
-    if (dbg_succ)
+    if (dbg_succ) {
         dbg_succ->reset_mask();
+        if (auto node_rc = dbg_succ->load_extension<NodeRC>(config->infbase)) {
+            logger->trace("Loaded graph RC file");
```
karasikov, line 307:
```suggestion
            logger->trace("Loaded the adj-rc index (adjacent to reverse-complement nodes)");
```

metagraph/src/cli/config/config.cpp
```diff
@@ -1088,6 +1090,7 @@ if (advanced) {
             fprintf(stderr, "\t   --state [STR] \tchange state of succinct graph: small / dynamic / stat / fast [stat]\n");
             fprintf(stderr, "\t   --to-adj-list \twrite adjacency list to file [off]\n");
             fprintf(stderr, "\t   --to-fasta \t\textract sequences from graph and dump to compressed FASTA file [off]\n");
+            fprintf(stderr, "\t   --adj-rc \t\tgenerate a reverse-complement index for the graph (only for primary succinct graphs) [off]\n");
```
karasikov, line 1093:
```suggestion
            fprintf(stderr, "\t   --adj-rc \t\tconstruct an index of adjacent to reverse-complement nodes (only for primary succinct graphs) [off]\n");
```

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -37,6 +37,14 @@ CanonicalDBG::CanonicalDBG(std::shared_ptr<const DeBruijnGraph> graph, size_t ca
         if (graph_->alphabet()[i] == boss::BOSS::kSentinel)
             has_sentinel_ = true;
     }
+
+    if (NodeRC *node_rc = graph_->get_extension_threadsafe<NodeRC>()) {
+        add_extension(std::shared_ptr<NodeRC>(std::shared_ptr<NodeRC>{}, node_rc));
+    } else {
+        auto ext = std::make_shared<NodeRC>();
+        ext->set_graph(*graph_);
+        add_extension(ext);
```
karasikov, line 46: At what point does it initialize the index?

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -256,101 +173,65 @@ ::call_outgoing_kmers(node_index node, const OutgoingEdgeCallback &callback) con
...
+         *         TGGCTT      AAGCCA
+         */
+
+        // for each n, check for nAGCCA. If found, define and store the index for
+        // TGGCTrc(n) as index(nAGCCA) + offset_
+        assert(get_extension<NodeRC>());
+
+        const DBGSuccinct *dbg_succ = get_dbg_succ(*graph_);
+        const boss::BOSS *boss = dbg_succ ? &dbg_succ->get_boss() : nullptr;
+        const auto cache = get_extension<NodeFirstCache>();
```
karasikov, line 194: Not thread-safe. All const methods in all graph classes must be thread-safe.

karasikov: Please also check other places.

karasikov: Actually, can we make get_extension() return a const reference instead of a pointer?

karasikov (changes requested):

metagraph/src/graph/graph_extensions/node_rc.hpp
```diff
@@ -0,0 +1,62 @@
...
+    // Use set_graph to set the graph pointer after using the default constructor
+    NodeRC() : graph_(nullptr) {};
+
+    // Construct a NodeRC index
+    NodeRC(const DBGSuccinct &graph);
+
+    void set_graph(const DeBruijnGraph &graph) {
+        assert(!rc_.size() || is_compatible(graph));
+        graph_ = &graph;
+    }
```
karasikov, line 32:
```suggestion
    // If |construct_index| is false, the index will not be initialized. Method load() must be called.
    NodeRC(const DBGSuccinct &graph, bool construct_index = false);
```

metagraph/src/graph/graph_extensions/node_rc.cpp
```diff
@@ -0,0 +1,275 @@
...
+}
+
+bool NodeRC::is_compatible(const SequenceGraph &graph, bool verbose) const {
+    if (!rc_.size()) {
+        if (const auto *dbg = dynamic_cast<const DeBruijnGraph*>(&graph)) {
+            if (dbg == graph_)
+                return true;
+
+            if (verbose)
+                logger->error("Stored graph pointer does not match");
```
karasikov, line 246: If it's an error, it has to be reported in every case. So `logger->error` should never be inside `if-verbose`

```suggestion
            logger->trace("Stored graph pointer does not match");
```

metagraph/src/graph/representation/canonical_dbg.cpp
hmusta: For this specific case, NodeFirstCache is added as an extension to `CanonicalDBG`, so it happens to never be used in a thread-unsafe way in this code.

For cases where thread safety is needed, I added the method `ExtensionSubtype* get_extension_threadsafe() const` already and I use it in `CanonicalDBG`'s constructor

metagraph/src/graph/graph_extensions/node_rc.cpp
karasikov: Also please fix this in other cases below

metagraph/src/graph/graph_extensions/node_rc.hpp
```diff
@@ -0,0 +1,62 @@
...
+    bool is_compatible(const SequenceGraph &graph, bool verbose = true) const;
+
+  private:
+    const DeBruijnGraph *graph_;
+
+    typedef bit_vector_smart Indicator;
+    typedef sdsl::dac_vector_dp<sdsl::rrr_vector<>> Mapping;
+
+    // Indicates whether a reverse complement prefix/suffix exists for each graph node.
+    Indicator rc_;
```
karasikov, line 49: What's the size of this bitmap?

metagraph/src/graph/representation/canonical_dbg.cpp
hmusta: `NodeRC`'s main purpose is to support the `call_outgoing_from_rc` and `call_incoming_from_rc` methods, so having an index loaded is not a strict requirement for it to function. If no index is available, it simply uses its stored graph pointer to compute those functions.

karasikov (changes requested):

metagraph/src/graph/representation/canonical_dbg.cpp
karasikov: So is it uninitialized then? Why do you add it?

metagraph/src/graph/representation/canonical_dbg.cpp
karasikov: `get_extension` is not thread-safe, so you can't call the const method `append_prev_rc_nodes` simultaneously from multiple threads, which should always be allowed.

karasikov (changes requested):

metagraph/src/graph/representation/canonical_dbg.cpp
hmusta: Calling traversal operations with `CanonicalDBG` from multiple threads is not ideal because of the caches, so I've always created a new `CanonicalDBG` for each thread.

But I guess it's good to use the thread-safe versions here as well, just in case.

metagraph/src/graph/representation/canonical_dbg.cpp
hmusta: The point is that I wanted all of the "call adjacent nodes to the reverse complements" code to be in the same place, so I moved it out of CanonicalDBG and into NodeRC.

So, NodeRC can perform the same task for all kinds of DeBruijnGraphs, even if they don't have an index computed and if they are not DBGSuccinct.

karasikov: Ah, I see. This wasn't obvious from the interface.

metagraph/src/graph/graph_extensions/node_rc.hpp
hmusta: I've incorporated this, but with a more accurate comment

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -251,108 +166,132 @@ ::call_outgoing_kmers(node_index node, const OutgoingEdgeCallback &callback) con
...
+    bool found = false;
+    try {
+        adjacent_outgoing_nodes(node, [&](node_index) {
+            if (found)
+                throw std::bad_function_call();
 
-            is_palindrome_cache_.Put(prev, true);
+            found = true;
         });
+    } catch (const std::bad_function_call&) { return true; }
```
karasikov, line 280:
```suggestion
    } catch (const std::bad_function_call&) {
        return true;
    }
```

metagraph/src/graph/graph_extensions/node_rc.hpp
```diff
@@ -0,0 +1,54 @@
...
+    bool load(const std::string &filename_base);
+    void serialize(const std::string &filename_base) const;
+
+    bool is_compatible(const SequenceGraph &graph, bool verbose = true) const;
+
+  private:
+    const DeBruijnGraph *graph_;
+
+    typedef bit_vector_smart Indicator;
+    typedef sdsl::dac_vector_dp<sdsl::rrr_vector<>> Mapping;
```
karasikov, line 37: Usually `sdsl::dac_vector_dp<sdsl::rrr_vector<>>` is not significantly smaller than `sdsl::dac_vector_dp<>`. Did you compare?

metagraph/src/graph/graph_extensions/node_rc.hpp
```diff
@@ -0,0 +1,54 @@
...
+// Maps each node in a PRIMARY-mode DeBruijnGraph to nodes adjacent to its reverse
+// complement. When using the index construct constructor on DBGSuccinct, or when
+// loading such an index, this stores a map from each DBGSuccinct to the BOSS nodes
+// corresponding to the reverse complements of its k-1 prefix and k-1 suffix.
+class NodeRC : public SequenceGraph::GraphExtension {
+  public:
+    using node_index = SequenceGraph::node_index;
+
+    // If construct_index is false, then an index must be loaded to take advantage
+    // of the index for call_*_from_rc calls
```
karasikov, line 22:
```suggestion
    // If construct_index is true, a fully pre-computed index will be constructed.
    // Otherwise, call_*_from_rc will invoke on-the-fly computations on the graph,
    // unless the index is loaded with load().
```

metagraph/src/graph/graph_extensions/node_rc.hpp
karasikov: On the other hand, if it's small, it's probably accessed so rarely that it doesn't contribute to the speed much, so we can use rrr too. In any case, what's the index size compared to the graph size?

metagraph/src/graph/graph_extensions/node_rc.cpp
```diff
@@ -0,0 +1,274 @@
+#include "node_rc.hpp"
+
+#include "graph/representation/succinct/dbg_succinct.hpp"
+#include "common/seq_tools/reverse_complement.hpp"
+#include "common/utils/template_utils.hpp"
+
+#include <ips4o.hpp>
+#include <mutex>
```
karasikov, line 8:
```suggestion
#include <mutex>

#include <ips4o.hpp>

#include "graph/representation/succinct/dbg_succinct.hpp"
#include "common/seq_tools/reverse_complement.hpp"
#include "common/utils/template_utils.hpp"
```

metagraph/src/graph/graph_extensions/node_rc.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+    const auto &alphabet = graph_->alphabet();
+    for (size_t c = 0; c < alphabet.size(); ++c) {
+        if (alphabet[c] != boss::BOSS::kSentinel) {
+            rev_seq.back() = complement(alphabet[c]);
+            node_index prev = graph_->kmer_to_node(rev_seq);
+            if (prev != DeBruijnGraph::npos)
+                callback(prev);
+        }
+    }
```
karasikov, line 150: better do inside -`else`. It's a relatively small chunk compared to the one under `if`

metagraph/src/graph/graph_extensions/node_rc.cpp
```diff
@@ -0,0 +1,274 @@
...
+        std::ifstream instream(rc_filename, std::ios::binary);
+        if (!instream.good())
+            return false;
+
+        rc_.load(instream);
+        mapping_.load(instream);
+        return true;
+
+    } catch (...) {
+        logger->error("Cannot load graph RC from file {}", rc_filename);
```
karasikov, line 230: Move the error message to where you check the returned value of this `load()`

karasikov (approved):

metagraph/src/graph/graph_extensions/node_rc.hpp
hmusta: It depends on how many near-reverse complements exist in the graph. For the BIGSI graph, for example, the index takes up 8GB, compared to 15GB and 28GB for the small and normal representations of the graph.

hmusta: Comparing with and without RRR, the execution time is the same, but RRR is 1.02x smaller, so it doesn't hurt

karasikov: I'm surprised by the overall index size, you said in the meeting it's around 50% of the graph, which means it's about 1 bit per edge. So what's the density of the indicator vector?

hmusta: The `Mapping` array has an element for 2.75% of the graph indices, and this is what takes up most of the space.

Here's a comparison of the sizes for the 750 subset of BIGSI:
`bit_vector_smart` indicator alone: 17MB

full index:
`sdsl::dac_vector_dp<sdsl::rrr_vector<>>`: 84MB
`sdsl::vlc_vector<>`: 97MB
`sdsl::int_vector<>`: 272MB
`sdsl::int_vector<>` packed: 137MB

karasikov: Can you compare this for a larger graph? Say, 30 GB?

karasikov: Just take the SRA-Microbe graph, for instance.

PR #387, opened 2022-02-20
NodeFirstCache graph extension for speeding up backwards traversal in DBGSuccinct

hmusta:
A simpler implementation of #334


metagraph/src/cli/align.cpp
```diff
@@ -405,8 +406,12 @@ int align_to_graph(Config *config) {
                 // Wrap it in CanonicalDBG if needed. This way, each thread gets its
                 // own wrapper (and more importantly, its own local cache).
                 if (aln_graph->get_mode() == DeBruijnGraph::PRIMARY) {
-                    aln_graph = std::make_shared<CanonicalDBG>(aln_graph);
+                    aln_graph = std::make_shared<CanonicalDBG>(aln_graph, 0);
                     logger->trace("Primary graph wrapped into canonical");
+                    // If backwards traversal on DBGSuccinct will be needed, then
+                    // add a cache to speed it up.
+                    if (dbg_succ && aln_graph->get_mode() != DeBruijnGraph::CANONICAL)
+                        aln_graph->add_extension(std::make_shared<NodeFirstCache>(*dbg_succ));
```
karasikov, line 414: How do these two caches interact (in `CanonicalDBG` and `NodeFirstCache`)?
Would it make sense to share the same storage of cached elements?

metagraph/src/graph/graph_extensions/node_first_cache.hpp
```diff
@@ -0,0 +1,45 @@
...
+    void serialize(const std::string &) const { throw std::runtime_error("Not implemented"); }
+
+    bool is_compatible(const SequenceGraph &graph, bool verbose = true) const;
+
+  private:
+    const DBGSuccinct *dbg_succ_;
+
+    using edge_index = boss::BOSS::edge_index;
+    mutable caches::fixed_sized_cache<edge_index, std::pair<edge_index, edge_index>,
+                                      caches::LRUCachePolicy<node_index>> first_cache_;
```
karasikov, line 36: Having a comment explaining what's in the cache would be nice

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -398,29 +406,38 @@ ::call_incoming_kmers(node_index node, const IncomingEdgeCallback &callback) con
...
             --max_num_edges_left;
         }
-    });
+    };
+
+    if (const auto cache = get_extension<NodeFirstCache>()) {
+        cache->call_incoming_kmers(node, incoming_kmer_callback);
+    } else {
+        graph_->call_incoming_kmers(node, incoming_kmer_callback);
+    }
```
karasikov, line 434: I think there is an unnecessary degree of freedom: `get_extension<NodeFirstCache>` vs. `use_parent_child_cache_`. Ideally we should have a single parameter for both. See another question above.

karasikov (changes requested):

metagraph/src/graph/representation/canonical_dbg.cpp
hmusta: I was doing this for now as part of my testing. Ideally, `caches::fixed_sized_cache` should just not cache anything if the size is set to 0 (the default behaviour is to cache everything).

metagraph/src/cli/align.cpp
hmusta: From benchmarking, I've found that having a `NodeRC` index or (inclusive) a `NodeFirstCache` leads to better performance and the caches in `CanonicalDBG` actually worsen performance in these cases.

The reason why I want to keep these separate is that `NodeFirstCache` is used by both `CanonicalDBG` and `RCDBG`.

metagraph/src/graph/graph_extensions/node_first_cache.hpp
hmusta: done

metagraph/src/graph/representation/canonical_dbg.cpp
hmusta: I created a new PR in the caches library to support empty caches, so I can update this accordingly.

metagraph/src/cli/align.cpp
karasikov: So if caching in `CanonicalDBG` doesn't help, why do we keep it?

hmusta: My plan now is to get rid of it, I wanted to run the benchmarks first.

hmusta: anyway, I got rid of it

metagraph/src/graph/representation/canonical_dbg.cpp
hmusta: I got rid of this variable now.

karasikov (approved):
Looks good, but I would check again before merging to master.

PR #389, opened 2022-02-24
Fixes to seed labeling and filtering

hmusta:


metagraph/src/graph/alignment/alignment.hpp
```diff
@@ -261,6 +261,78 @@ std::string spell_path(const DeBruijnGraph &graph,
...
+                ++b_begin;
+                ++out;
+            }
+        }
+    }
+
+    int64_t offset_;
+};
+
+struct CoordDifference {
```
karasikov, line 286: `SetDifference`?

metagraph/src/graph/alignment/alignment.hpp
```diff
@@ -261,6 +261,78 @@ std::string spell_path(const DeBruijnGraph &graph,
...
+    CoordDifference(int64_t offset = 0) : offset_(offset) {}
```
karasikov, line 287: offset -> delta

metagraph/src/graph/alignment/alignment.hpp
```diff
@@ -261,6 +261,78 @@ std::string spell_path(const DeBruijnGraph &graph,
...
+                *out = *a_begin;
+                ++a_begin;
+                ++b_begin;
+                ++out;
+            }
+        }
+    }
+
+    int64_t offset_;
+};
```
karasikov, line 284: why not a single function?

```
void set_difference(It1 a_begin, It1 a_end, It2 b_begin, It2 b_end, Out out, T delta = 0) {
```

karasikov (changes requested):

karasikov (approved):
otherwise, looks good

metagraph/src/graph/alignment/alignment.hpp
hmusta: I'd rather not call it simply `SetDifference` because it's not exactly doing that. How about `SetDifferenceWithOffset`?

karasikov: `WithOffset` is redundant. It may well be without any offset.

PR #393, opened 2022-03-22
Various alignment bug fixes

hmusta:


hmusta:
> Does it also fix the bug where the alignment score is larger than the maximum possible score (e.g., 109/105)?

No. That's actually a bug that needs to be fixed in the web interface repo. I'll open an issue there.

metagraph/src/graph/alignment/aligner_extender_methods.cpp
```diff
@@ -427,6 +428,10 @@ std::vector<Alignment> DefaultColumnExtender::extend(score_t min_path_score,
     table.clear();
     prev_starts.clear();
 
+    // prevent overflow
+    if (config_.xdrop > std::numeric_limits<score_t>::max() - added_xdrop)
+        added_xdrop = std::numeric_limits<score_t>::max() - config_.xdrop;
```
karasikov, line 433: Rather than changing the value, wouldn't it be better to do the checks below more carefully so you don't let the overflow happen?

hmusta: The point is that I want the xdrop parameter to be the saturated addition of `config_.xdrop` and `added_xdrop`. So another way I could do it is to just define a new variable as such called `xdrop` and use that everywhere instead.

hmusta: This allows the user to disable the xdrop heuristic by setting the xdrop cutoff to max_int.

hmusta:
I'm closing this. I realized that these changes don't actually fix the bug I was targeting, and that more of the changes in #392 are required.

PR #394, opened 2022-03-22
Fix chaining corner cases

hmusta:
- Two adjacent seeds in a chain now have a well-defined coordinate distance
- Better handle cases where seeds in a chain cannot connect via extension
- Support alignment of queries with non-alphabet characters
- Detect overflows when handling the xdrop parameter
- Correctly handle label changes in unitigs with more than two nodes

Broken off from #392, overrides #393


karasikov (approved):
I just skimmed through and it looks ok to me. Do you want me to check some sections in detail?

hmusta:
> I just skimmed through and it looks ok to me. Do you want me to check some sections in detail?

Not in particular. Thanks!

PR #395, opened 2022-03-24
added inverse_select to wavelet_tree to support more efficient BOSS traversal

hmusta:
TODO
- [x] Update DYNAMIC module after merging karasikov/DYNAMIC#2
- [x] Use new inverse_select method from dynamic wavelet tree

### DIFF
--- a/metagraph/src/common/vectors/wavelet_tree.cpp
+++ b/metagraph/src/common/vectors/wavelet_tree.cpp
@@ -188,6 +188,15 @@ TAlphabet wavelet_tree_sdsl_fast<t_wt_sdsl>::operator[](uint64_t i) const {
 }
 
 template <class t_wt_sdsl>
+std::pair<uint64_t, TAlphabet>
+wavelet_tree_sdsl_fast<t_wt_sdsl>::inverse_select(uint64_t i) const {
+    assert(i < size());
+    auto val = wwt_.inverse_select(i);
+    ++val.first;
+    return val;
+}
+
+template <class t_wt_sdsl>
 uint64_t wavelet_tree_sdsl_fast<t_wt_sdsl>::next(uint64_t i, TAlphabet c) const {
     assert(i < size());
     assert(c < (1llu << logsigma()));
@@ -248,6 +257,13 @@ uint64_t wavelet_tree_dyn::select(TAlphabet c, uint64_t i) const {
 TAlphabet wavelet_tree_dyn::operator[](uint64_t i) const {
     assert(i < size());
     return dwt_.at(i);
+}
+
+std::pair<uint64_t, TAlphabet>
+wavelet_tree_dyn::inverse_select(uint64_t i) const {
+    assert(i < size());
+    TAlphabet val = dwt_.at(i);
+    return std::make_pair(rank(val, i), val);
 }
 
 uint64_t wavelet_tree_dyn::next(uint64_t i, TAlphabet c) const {
@@ -343,6 +359,15 @@ template <class t_wt_sdsl>
 TAlphabet wavelet_tree_sdsl<t_wt_sdsl>::operator[](uint64_t i) const {
     assert(i < size());
     return wwt_[i];
+}
+
+template <class t_wt_sdsl>
+std::pair<uint64_t, TAlphabet>
+wavelet_tree_sdsl<t_wt_sdsl>::inverse_select(uint64_t i) const {
+    assert(i < size());
+    auto val = wwt_.inverse_select(i);
+    ++val.first;
+    return val;
 }
 
 template <class t_wt_sdsl>
@@ -456,6 +481,13 @@ template <class t_bv>
 TAlphabet partite_vector<t_bv>::operator[](uint64_t i) const {
     assert(i < size());
     return int_vector_[i];
+}
+
+template <class t_bv>
+std::pair<uint64_t, TAlphabet> partite_vector<t_bv>::inverse_select(uint64_t i) const {
+    assert(i < size());
+    TAlphabet val = int_vector_[i];
+    return std::make_pair(rank(val, i), val);
 }
 
 template <class t_bv>
--- a/metagraph/src/common/vectors/wavelet_tree.cpp
+++ b/metagraph/src/common/vectors/wavelet_tree.cpp
@@ -262,8 +262,9 @@ TAlphabet wavelet_tree_dyn::operator[](uint64_t i) const {
 std::pair<uint64_t, TAlphabet>
 wavelet_tree_dyn::inverse_select(uint64_t i) const {
     assert(i < size());
-    TAlphabet val = dwt_.at(i);
-    return std::make_pair(rank(val, i), val);
+    auto val = dwt_.inverse_select(i);
+    ++val.first;
+    return val;
 }
 
 uint64_t wavelet_tree_dyn::next(uint64_t i, TAlphabet c) const {
--- a/metagraph/src/common/vectors/wavelet_tree.hpp
+++ b/metagraph/src/common/vectors/wavelet_tree.hpp
@@ -22,6 +22,7 @@ class wavelet_tree {
     virtual uint64_t rank(TAlphabet c, uint64_t i) const = 0;
     virtual uint64_t select(TAlphabet c, uint64_t i) const = 0;
     virtual TAlphabet operator[](uint64_t i) const = 0;
+    virtual std::pair<uint64_t, TAlphabet> inverse_select(uint64_t i) const = 0;
 
     // get the position of the next value |c| in subvector [i, ...]
     virtual uint64_t next(uint64_t i, TAlphabet c) const = 0;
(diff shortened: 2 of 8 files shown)


metagraph/src/common/vectors/wavelet_tree.hpp
```diff
@@ -75,6 +76,7 @@ class wavelet_tree_sdsl_fast : public wavelet_tree_sdsl_augmented {
     uint64_t rank(TAlphabet c, uint64_t i) const;
     uint64_t select(TAlphabet c, uint64_t i) const;
     TAlphabet operator[](uint64_t i) const;
+    std::pair<uint64_t, TAlphabet> inverse_select(uint64_t i) const;
```
karasikov, line 79: Add unit tests?

metagraph/src/graph/representation/succinct/boss.cpp
```diff
@@ -639,6 +639,27 @@ edge_index BOSS::fwd(edge_index i, TAlphabet c) const {
...
+    CHECK_INDEX(i);
+
+    auto [rank, c] = W_->inverse_select(i);
+    assert(i == 1 || c != kSentinelCode);
+    if (c == kSentinelCode) {
+        --rank;
+    } else if (c >= alph_size) {
+        c %= alph_size;
+        rank = rank_W(i, c);
+    }
```
karasikov, line 656: `inverse_select` + `rank` (which happens now in the second branch) is actually more expensive than `access` + `rank` (performed previously). Did you benchmark this for all types?

In fact, for some types (e.g., `wavelet_tree_dyn`) this PR implements `inverse_select` via the same `access` + `rank`, so the `fwd()` calls became strictly slower now ([`access` + `rank` or `access` + 2*`rank`] vs. [`access` + `rank`]).

karasikov, line 656: `inverse_select` + `rank` might actually be more expensive than access + rank. Did you benchmark this for all types?

karasikov (changes requested):
Any benchmarks and tests?

metagraph/src/graph/representation/succinct/boss.cpp
hmusta: I've benchmarked this using alignment to a RowDiff-annotated microbial graph and this PR speeds it up about 1.2x.

The motivation is that the second branch occurs less frequently, and hence, its cost gets amortized.

hmusta: But yes, a more systematic benchmark is needed

hmusta: Using the benchmark code on the same graph:
```
-------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations
-------------------------------------------------------------------
BM_BOSS_get_W_and_fwd          2.90 us         2.89 us       241361
BM_BOSS_no_get_W_and_fwd       2.48 us         2.47 us       281466
```

metagraph/src/common/vectors/wavelet_tree.hpp
hmusta: done

metagraph/src/graph/representation/succinct/boss.cpp
hmusta: On the full microbe graph, it seems to be only 4% faster
small
```
-------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations
-------------------------------------------------------------------
BM_BOSS_get_W_and_fwd          6.31 us         6.31 us       105374
BM_BOSS_no_get_W_and_fwd       6.06 us         6.06 us       117376
```
stat
```
-------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations
-------------------------------------------------------------------
BM_BOSS_get_W_and_fwd         0.773 us        0.773 us       898763
BM_BOSS_no_get_W_and_fwd      0.744 us        0.744 us       950109
```

karasikov:
Hm, this is weird. Shouldn't these changes make it slower for `dyn`?

hmusta:
> Hm, this is weird. Shouldn't these changes make it slower for `dyn`?

Hmm...it could be that since the two calls now immediately follow one another, the CPU's branch predictor is working better when traversing down the tree.

Regardless, I can probably get better performance by just implementing `inverse_select` in DYNAMIC, it doesn't look hard to do.

metagraph/src/graph/representation/succinct/boss.cpp
hmusta: Using the new implementation of `inverse_select` in DYNAMIC on the microbe 750 graph
```
-------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations
-------------------------------------------------------------------
BM_BOSS_get_W_and_fwd          4.14 us         4.13 us       168212
BM_BOSS_no_get_W_and_fwd       3.62 us         3.61 us       198161
```

karasikov (approved):

karasikov:
don't forget to update the external library

metagraph/src/common/vectors/wavelet_tree.cpp
```diff
@@ -250,6 +259,13 @@ TAlphabet wavelet_tree_dyn::operator[](uint64_t i) const {
     return dwt_.at(i);
 }
 
+std::pair<uint64_t, TAlphabet>
+wavelet_tree_dyn::inverse_select(uint64_t i) const {
+    assert(i < size());
+    TAlphabet val = dwt_.at(i);
+    return std::make_pair(rank(val, i), val);
```
karasikov, line 266: fix this before merge

metagraph/src/common/vectors/wavelet_tree.cpp
```diff
@@ -458,6 +483,13 @@ TAlphabet partite_vector<t_bv>::operator[](uint64_t i) const {
     return int_vector_[i];
 }
 
+template <class t_bv>
+std::pair<uint64_t, TAlphabet> partite_vector<t_bv>::inverse_select(uint64_t i) const {
+    assert(i < size());
+    TAlphabet val = int_vector_[i];
+    return std::make_pair(rank(val, i), val);
```
karasikov, line 490: so for this type, the old implementation makes more sense. Can you check it inside `fwd(edge)` and avoid `inverse_select` for `partite_vector`?

karasikov (changes requested):

metagraph/src/common/vectors/wavelet_tree.hpp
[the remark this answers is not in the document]
hmusta: I extended `reference_based_test` to check this.

metagraph/src/common/vectors/wavelet_tree.cpp
hmusta: done

metagraph/src/common/vectors/wavelet_tree.cpp
hmusta: The extra dynamic_cast for this check slows things down.
```
volatile auto t = dynamic_cast<const partite_vector<>*>(W_);
    std::ignore = t;
    // if (!dynamic_cast<const partite_vector<>*>(W_)) {
        auto [rank, c] = W_->inverse_select(i);
        assert(i == 1 || c != kSentinelCode);
        if (c == kSentinelCode) {
            --rank;
        } else if (c >= alph_size) {
            // this branch is rare on graphs with low average in-degree
            c %= alph_size;
            rank = rank_W(i, c);
        }
        // get the index of the target node
        node_index target_node = NF_[c] + rank;
        // get the index of the representative edge with that target node
        return select_last(target_node);
    // }

    // return fwd(i, get_W(i) % alph_size);
```
gets me this on the stat representation
```
-------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations
-------------------------------------------------------------------
BM_BOSS_get_W_and_fwd         0.338 us        0.337 us      2094306
BM_BOSS_no_get_W_and_fwd      0.532 us        0.531 us      1154902
```

whereas if I don't do the cast at all
```
-------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations
-------------------------------------------------------------------
BM_BOSS_get_W_and_fwd         0.386 us        0.385 us      1904289
BM_BOSS_no_get_W_and_fwd      0.379 us        0.378 us      1827670
```

And I'm not sure if it's even worth doing all this, since this PR doesn't seem to be negatively affecting fwd on partite_vector anyway
```
-------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations
-------------------------------------------------------------------
BM_BOSS_get_W_and_fwd         0.443 us        0.442 us      1511411
BM_BOSS_no_get_W_and_fwd      0.420 us        0.419 us      1681473
```

hmusta: An alternative would be to do the dynamic cast one in the BOSS constructor and loader and keep using that.

PR #398, opened 2022-03-28
Optimize coordinate handling during alignment

hmusta:
This can help prevent buffering a large number of annotation to RAM.

TODO:
- [x] Skip seeds that have too many coordinates
- [x] Only store coordinates in a cache. Fetch them when needed.
- [ ] When an element gets popped off the DP table, also pop its annotation from the buffer if the annotation is unique (later?)


karasikov (changes requested):
Can you add some tests? In the last commits, you fixed something non-trivial but nevertheless, all the tests were passed...

PR #406, opened 2022-06-01
refactored processing with buffers

karasikov:


metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -1003,32 +931,24 @@ void convert_batch_to_row_diff(const std::string &pred_succ_fprefix,
                     uint64_t succ_value = is_anchor ? 0 : get_value(source_col, source_idx, j, *succ);
                     if (succ_value != curr_value) {
                         // no reduction, we must keep the bit
-                        auto &v = set_rows_fwd[source_idx][j];
                         if constexpr(with_values) {
-                            v.emplace_back(row_idx, curr_value - succ_value);
+                            set_rows[source_idx][j].emplace_back(row_idx, curr_value - succ_value);
```
hmusta, line 935: Before you had `set_rows_fwd` and `set_rows_bwd`, but now it looks like these have been consolidated into a single `set_rows` vector. are they now interleaved or something? I wanted to double-check if this is what you intended.

metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -1343,48 +1259,67 @@ void convert_batch_to_row_diff_coord(const std::string &pred_succ_fprefix,
...
+                tmp_dir = tmp_path/fmt::format("{}/col_{}_{}", s / 100, s, j);
+                fs::create_directories(tmp_dir);
+                uint64_t original_nbits = sources[s].get_matrix().data()[j]->num_set_bits();
+                buffer_size = std::min(buf_size_bytes / sizeof(T), original_nbits);
+            }
 
-            // make sure the first chunk exists even if empty
-            dump_chunk_to_disk({}, s, j, 0);
-            set_rows_fwd[s][j].reserve(std::min(buf_size, original_nbits));
+            set_rows[s].emplace_back(1, buffer_size, tmp_dir, chunks_open_per_thread);
```
hmusta, line 1322: see above comment

metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.hpp
```diff
@@ -166,7 +167,8 @@ class ColumnCompressed : public MultiLabelEncoded<Label> {
     uint8_t count_width_;
     uint64_t max_count_;
     std::vector<sdsl::int_vector<>> relation_counts_;
-    std::vector<Vector<std::pair<Index, uint64_t>>> coords_;
+    std::vector<common::SortedVector<std::pair<Index, uint64_t>>> coords_;
```
hmusta, line 170: maybe add a comment here to state that this may be stored either completely in RAM, or in a disk buffer depending on input parameters (at least, this is what I gathered from reading ahead)

metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
```diff
@@ -253,11 +268,16 @@ void ColumnCompressed<Label>::serialize_counts(const std::string &filename) cons
...
+                size_t offset = out.tellp();
+                out.close();
+                {
+                    sdsl::int_vector_buffer<> packed_counts(counts_fname, std::ios::out, 1024*1024,
+                                                            packed_width, false, offset);
+                    for (size_t c : relation_counts_[j]) {
+                        packed_counts.push_back(c);
+                    }
+                }
+                out = std::ofstream(counts_fname, std::ios::binary | std::ios::app);
```
hmusta, line 280: I know it's hard to do this differently, but do you know if this repeated opening and closing causes much overhead?

hmusta (changes requested):

metagraph/src/common/sorted_vector.hpp
```diff
@@ -0,0 +1,135 @@
...
+        if (buffer_.size() == buffer_.capacity())
+            flush();
+    }
+
+    void push_back(const T &value) {
+        buffer_.push_back(value);
+        total_size_++;
+        if (buffer_.size() == buffer_.capacity())
+            flush();
+    }
```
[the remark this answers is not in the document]
karasikov: There is some difference between push_back and emplace_back. Keeping them separate might help detecting some bugs. Maybe a very unlikely thing for this class, but still. It's not. log of code overhead.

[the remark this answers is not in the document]
karasikov: https://stackoverflow.com/questions/10890653/why-would-i-ever-use-push-back-instead-of-emplace-back

metagraph/src/annotation/row_diff_builder.cpp
karasikov: Previously, one of them was storing positions that are already sorted, and the sorting was applied only to the other buffer. But now I figured it's not a big difference if we just sort everything together so now I push everything into the same buffer.

metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
karasikov: It shouldn't be large, maybe a few minutes in total even for a million columns, but I'll add a timer and will print logging messages. The next time we run this thing, so we'll see how big the overhead is in logs.

PR #411, opened 2022-07-10
detect max number of open files and limit num chunks merged

karasikov:


metagraph/src/common/elias_fano/elias_fano_merger.hpp
```diff
@@ -193,15 +193,15 @@ void merge_files(std::vector<std::string> sources,
...
+    assert(max_sources_open >= 3);
     while (sources.size() > max_sources_open) {
         // chunk 0 may be special (e.g. storing only fwd bits) and hence
         // never pre-merged
         std::vector<std::string> new_chunks = { sources.at(0) };
         size_t i = 1;
         while (i < sources.size()) {
             std::vector<std::string> to_merge;
-            while (i < sources.size() && to_merge.size() < max_sources_open) {
+            while (i < sources.size() && to_merge.size() + 1 < max_sources_open) {
```
hmusta, line 204: I'm guessing the +1 is because of the output file? Please add a comment to explain this

metagraph/src/common/unix_tools.cpp
```diff
@@ -111,6 +115,30 @@ size_t get_curr_RSS()
...
 
+size_t get_max_files_open() {
+    struct rlimit lim;
+    getrlimit(RLIMIT_NOFILE, &lim);
+    rlim_t max_files = lim.rlim_cur;
+    return max_files;
+}
+
+int get_num_fds() {
+    FILE *p = popen(fmt::format("lsof -p {} | tail -n +2 | wc -l", getpid()).c_str(), "r");
```
hmusta, line 126: Isn't this unsafe since it's getting this value from an external process?

Maybe add a TODO to get this value from `lsof.h` or some POSIX API calls

metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
```diff
@@ -669,25 +711,36 @@ BOSS::Chunk construct_boss_chunk_disk(KmerCollector &kmer_collector,
...
     for (size_t i = 0; i < chunk_fnames.size(); ++i) {
         chunks_split[i] = split<T_REAL>(k, chunk_fnames[i]);
     }
 
     // for a DNA alphabet, this will contain 16 chunks, split by kmer[0] and kmer[1]
-    std::vector<std::string> real_F_W(std::pow(KmerExtractor2Bit().alphabet.size(), 2));
+    std::vector<std::string> real_F_W(A2);
+
+    n_threads = check_fd_and_adjust_threads(std::min(num_threads, real_F_W.size()),
+                                            4 * (2 + utils::is_pair_v<T>));
```
hmusta, line 734: is the `4` here supposed to be the alphabet size?

metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
```diff
@@ -404,6 +404,27 @@ void skip_same_suffix(const KMER &el, Decoder &decoder, size_t suf) {
     }
 }
 
+int check_fd_and_adjust_threads(size_t num_threads, size_t fd_per_thread) {
```
hmusta, line 407: why not move this to `common`? This could potentially be useful in other parts of the code.

hmusta (changes requested):

hmusta:
What is the maximum number of threads used during the unit and integration tests? I'm thinking it would be useful to have one test with a large number of threads to test out construction in that scenario.

metagraph/src/common/unix_tools.cpp
karasikov: Ah, good that you found that. When I quickly checked it on the internet, I couldn't find a way to get around invoking `lsof` in a separate process.

metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
karasikov: no, it's 2 + 2 (2 chunks with 2 files per each)

metagraph/src/graph/representation/succinct/boss_chunk_construct.cpp
karasikov: makes sense

metagraph/src/common/unix_tools.cpp
karasikov: I checked it out but it doesn't look straightforward. Not sure that makes sense.

metagraph/src/common/elias_fano/elias_fano_merger.hpp
karasikov: done

metagraph/src/common/unix_tools.cpp
hmusta: ok, then let's add a TODO comment here so we can find it later

hmusta (approved):

metagraph/src/common/unix_tools.cpp
karasikov: Not sure that would be a good thing to spend time on... The current approach should work fine.

PR #417, opened 2022-07-23
compress node ranges index

karasikov:
Add compression of the node ranges index.

Now allows indexing all 10-mers with ~1 MB and all 15-mers using ~400 MB.

### DIFF
--- a/metagraph/src/graph/representation/succinct/boss.cpp
+++ b/metagraph/src/graph/representation/succinct/boss.cpp
@@ -1663,7 +1663,7 @@ BOSS::erase_redundant_dummy_edges(sdsl::bit_vector *source_dummy_edges,
                                   size_t num_threads,
                                   bool verbose) {
     // reset the index of suffix ranges
-    index_suffix_ranges(0, 0);
+    index_suffix_ranges(0);
 
     sdsl::bit_vector redundant_dummy_edges_mask(W_->size(), false);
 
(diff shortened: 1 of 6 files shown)


metagraph/src/graph/representation/succinct/boss.cpp
```diff
@@ -3147,76 +3138,54 @@ void BOSS::call_kmers(Call<edge_index, const std::string&> callback) const {
...
                 edge_index rl_next = rl;
                 edge_index ru_next = ru;
                 if (!tighten_range(&rl_next, &ru_next, c))
                     continue;
 
-                assert(idx < num_suffixes);
-                narrowed.emplace_back(num_suffixes * (c - 1) + idx,
-                                      rl_next, ru_next);
+                narrowed[2 * i + 2 * (N * (c - 1))] = rl_next;
+                narrowed[2 * i + 2 * (N * (c - 1)) + 1] = ru_next + 1;
```
hmusta, line 3174: If `tighten_range` in a previous iteration fails, doesn't this lead to extra `rank_W` calls being made since `tighten_range` is called on `(1, 1)`?

hmusta (changes requested):

metagraph/src/graph/representation/succinct/boss.cpp
karasikov: if it fails, `(rl, ru) = (1, 0)`, so `tighten_range` will skip the computation and will return false.

hmusta: but if `tighten_range` fails, it doesn't modify the values:
https://github.com/ratschlab/metagraph/blob/6a2f4e024bdaa0e8cc549898ee7cad673375c3ec/metagraph/src/graph/representation/succinct/boss.hpp#L674

karasikov: It does ru = ... - 1 in one of the lines above, so ru becomes < rl.

PR #419, opened 2022-08-04
Compilation fixes

hmusta:


hmusta:
This issue only seems to come up when running the main `Makefile`, and not when creating a `build` directory and running `cmake`. So, I have a feeling that running everything through `bash -c` call in the `Makefile` is not passing all environment variables.

hmusta:
This was actually not completed. I tried this out and simply setting `-l` doesn't work.

Also, we need to change the workflow so that it does the compilation using the Makefile instead of cmake

karasikov:
why do we need this Makefile? What's wrong with cmake?

karasikov:
So, anyway. This bug is on the compiler side, so I suggest we just require a more recent version

PR #420, opened 2022-08-04
Row sparse disk

marekkokot:
Implementation of row_sparse_disk and row_diff_sparse_disk annotations.
It seems I am overestimating memory requirements, which results in using less memory than specified with `--mem-cap-gb`


karasikov:
Please apply [.clang-format](https://github.com/ratschlab/metagraph/blob/master/metagraph/.clang-format)

marekkokot:
Thanks for pointing this. I configured my environment to use .clang-format file and reformated my changes. If something is wrong let me know.

metagraph/src/cli/load/load_annotation.hpp
```diff
@@ -20,7 +20,8 @@ initialize_annotation(Config::AnnotationType anno_type,
                       const std::string &swap_dir = "",
                       double memory_available_gb = 1,
                       uint8_t count_width = 8,
-                      size_t max_chunks_open = 2000);
+                      size_t max_chunks_open = 2000,
+                      size_t row_sparse_disk_buff_size = 10'000);
```
hmusta, line 24: why not make this generic and just call it `disk_buffer_size`?

metagraph/src/cli/config/config.hpp
```diff
@@ -91,6 +91,7 @@ class Config {
     unsigned int genome_binsize_anno = 1000;
     unsigned int arity_brwt = 2;
     unsigned int relax_arity_brwt = 10;
+    unsigned long long row_sparse_disk_buff_size = 10'000; // in B
```
hmusta, line 94: why not make this more generic and call it `disk_buffer_size`?

metagraph/src/cli/config/config.cpp
```diff
@@ -393,6 +393,8 @@ Config::Config(int argc, char *argv[]) {
             arity_brwt = atoi(get_value(i++));
         } else if (!strcmp(argv[i], "--relax-arity")) {
             relax_arity_brwt = atoi(get_value(i++));
+        } else if (!strcmp(argv[i], "--row-sparse-disk-buff-size")) {
```
hmusta, line 396: please add appropriate help message lines for this new flag so it show up when one types `--help`

metagraph/src/cli/load/load_annotation.cpp
```diff
@@ -110,6 +117,20 @@ initialize_annotation(Config::AnnotationType anno_type,
             annotation.reset(new annot::RowSparseAnnotator());
             break;
         }
+        case Config::RowSparseDisk: {
+            auto x = new annot::RowSparseDiskAnnotator();
+            const_cast<mtg::annot::binmat::RowSparseDisk &>(x->get_matrix())
+                    .set_buff_size(row_sparse_disk_buff_size);
```
hmusta, line 123: see comment about constructor

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.hpp
```diff
@@ -0,0 +1,108 @@
...
+namespace binmat {
+
+
+class RowSparseDisk : public BinaryMatrix {
+  public:
+    RowSparseDisk() {}
+    RowSparseDisk(const std::function<void(const RowCallback &)> &call_rows,
+                  uint64_t num_columns,
+                  uint64_t num_rows,
+                  uint64_t num_relations);
```
hmusta, line 21: instead of having to do the `const_cast` later to set the buffer size, why not initialize the buffer size here?

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.hpp
```diff
@@ -0,0 +1,108 @@
...
+                .num_relations_impl();
+    }
+
+    static void serialize(const std::function<void(binmat::BinaryMatrix::RowCallback)> &call_rows,
+                          const std::string &filename,
+                          uint64_t num_cols,
+                          uint64_t num_set_bits,
+                          uint64_t num_rows);
+
+    void set_buff_size(uint64_t _buff_size) {
```
hmusta, line 64: for these kinds of methods, we haven't been using preceding `_`s in variable names.

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.hpp
```diff
@@ -0,0 +1,108 @@
...
+    class Impl {
+        const bit_vector_small &boundary_;
+        sdsl::int_vector_buffer<> set_bits_;
+
+      public:
+        Impl(const bit_vector_small &boundary_,
+             const std::string &filename,
+             uint64_t offset,
+             uint64_t buff_size)
+            : boundary_(boundary_),
```
hmusta, line 84:
```suggestion
        Impl(const bit_vector_small &boundary,
             const std::string &filename,
             uint64_t offset,
             uint64_t buff_size)
            : boundary_(boundary),
```
it makes it more clear which is a class member and which is a constructor argument

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.hpp
```diff
@@ -0,0 +1,108 @@
...
+    const bit_vector_small get_boundary() const { return boundary_; }
+
+  private:
+    // For the multithreading to work properly the idea is to open int_vector_buffer<> per
+    // each public method call that needs to work with int_vector_buffer<> The actual
+    // implementation is wrapped in the class below to assure it will not call any methods
+    // of RowSparseDisk class which could result in multiple opens of int_vector_buffer
+    class Impl {
+        const bit_vector_small &boundary_;
+        sdsl::int_vector_buffer<> set_bits_;
```
hmusta, line 77: just as a style note, we've been putting private members after the `public` declarations, rather than at the top

metagraph/src/common/ifstream_with_name_and_offset.hpp
```diff
@@ -0,0 +1,25 @@
...
+
+  public:
+    explicit IfstreamWithNameAndOffset(const std::string &fname, std::ios_base::openmode mode)
+        : std::ifstream(fname, mode), fname(fname) {}
+
+    const std::string &GetFName() const { return fname; }
+
+    void SetOffset(uint64_t offset) { this->offset = offset; }
+
+    uint64_t GetOffset() const { return offset; }
```
hmusta, line 21: we've been using snake_case for class method names

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.cpp
```diff
@@ -0,0 +1,142 @@
...
+                           | std::ios::in);
+    boundary_start = outstream.tellp();
+    boundary.serialize(outstream);
+    outstream.seekp(boundary_start_pos, ios_base::beg);
+    outstream.write(reinterpret_cast<char *>(&boundary_start), sizeof(std::streampos));
+}
+
+} // namespace binmat
+} // namespace annot
+} // namespace mtg
```
hmusta, line 142: please add the newline after this

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.cpp
```diff
@@ -0,0 +1,142 @@
...
+BinaryMatrix::SetBitPositions RowSparseDisk::Impl::get_row(Row row) const {
+    assert(boundary_[boundary_.size() - 1] == 1);
+    uint64_t start_idx = row == 0 ? 0 : boundary_.select1(row) + 1;
+    uint64_t end_idx = boundary_.next1(start_idx);
+    SetBitPositions result;
+    result.reserve(end_idx - start_idx);
+    // In each row, the first value in `set_bits_` stores the first set bit,
+    // and all next ones store deltas pos[i] - pos[i-1].
+    uint64_t last = 0;
+    for (uint64_t i = start_idx; i != end_idx; ++i) {
```
hmusta, line 40: why not `i < end_idx`?

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.cpp
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    return rows;
+}
+
+bool RowSparseDisk::load(std::istream &f) {
+    auto _f = dynamic_cast<mtg::common::IfstreamWithNameAndOffset *>(&f);
+    assert(_f);
+    try {
+        f.read(reinterpret_cast<char *>(&num_columns_), sizeof(uint64_t));
```
hmusta, line 61: please use the methods in `serialization.hpp` for this kind of IO

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+
+void RowSparseDisk::serialize(const std::function<void(binmat::BinaryMatrix::RowCallback)> &call_rows,
+                              const std::string &filename,
+                              uint64_t num_cols,
+                              uint64_t num_set_bits,
+                              uint64_t num_rows) {
+    std::ofstream outstream(filename,
+                            std::ios::binary | std::ios::ate
+                                    | std::ios::in); // std::ios_base::in needed ...
```
hmusta, line 98: Could you elaborate on why in the comment? Is this is a bug in sdsl-lite?

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+    if (!outstream.good())
+        throw std::ofstream::failure("Cannot write to file " + filename);
+
+    outstream.write(reinterpret_cast<char *>(&num_cols), sizeof(uint64_t));
```
hmusta, line 103: use `serialization.hpp`

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.cpp
```diff
@@ -0,0 +1,142 @@
...
+        });
+        outbuf.close();
+    };
+
+    bit_vector_small boundary(call_bits, num_rows + num_set_bits, num_rows);
+
+    outstream.open(filename,
+                   std::ios::in | std::ios::out | std::ios::binary | std::ios::ate
+                           | std::ios::in);
+    boundary_start = outstream.tellp();
```
hmusta, line 134: why does this need all of these flags?

hmusta (changes requested):
Here's my first batch of comments. I still need to look at annotation_converters.cpp

metagraph/src/common/ifstream_with_name_and_offset.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+    const std::string& GetFName() const {
+        return fname;
+    }
+
+    void SetOffset(uint64_t offset) {
+        this->offset = offset;
+    }
+
+    uint64_t GetOffset() const {
```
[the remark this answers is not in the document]
marekkokot: Fixed

metagraph/src/cli/config/config.hpp
marekkokot: Answered above

metagraph/src/cli/config/config.cpp
marekkokot: Done

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.hpp
marekkokot: Fixed

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.hpp
marekkokot: Fixed

metagraph/src/common/ifstream_with_name_and_offset.hpp
marekkokot: Fixed

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.cpp
marekkokot: Fixed

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.cpp
marekkokot: this was copied from `row_sparse`, does it makes any difference? I can change that but then maybe it should also be changed in `row_sparse`?

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.cpp
marekkokot: Fixed

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.cpp
marekkokot: Ou, actually in the first open it is probably not needed.  Please read also my answer below.

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.cpp
marekkokot: I needed to close the file before using `int_vector_buffer` because it will open and write to the same file. I will need to store the boundary vector, which position in the file is before `int_vector_buffer` content. When I reopen the file (after closing it via `int_vector_buffer`) I want to append the boundary vector (thus ios::ate). The problem is I cannot go back to the place where the boundary position is to be stored when opening with ios::ate unless I also specify ios::in.
Check this: https://stackoverflow.com/questions/47014463/ofstream-open-modes-ate-vs-app This is kinda confusing and counterintuitive that std::in must be used while it is out stream... I think std::out could be skipped but I decided to leave it here to be more verbose.
If you have any suggestions on how to do it better let me know!

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.cpp
marekkokot: Fixed

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.cpp
marekkokot: Hmm, nie I think it is needed because before this I am writing to the same file (annotation mapping and row-diff stuff)

metagraph/src/cli/config/config.hpp
```diff
@@ -91,6 +91,7 @@ class Config {
     unsigned int genome_binsize_anno = 1000;
     unsigned int arity_brwt = 2;
     unsigned int relax_arity_brwt = 10;
+    unsigned long long disk_buffer_size = 10'000; // in B
```
karasikov, line 94: Is this the buffer for an int_vector_buffer? They say modern SSDs' page size is 8192 bytes. So maybe set this to a multiple of that, like 16k? Also, what about hardcode this and not expose as a parameter? I guess we aren't going to change this ever, are we?

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.hpp
```diff
@@ -0,0 +1,107 @@
...
+      private:
+        const bit_vector_small &boundary_;
+        sdsl::int_vector_buffer<> set_bits_;
+    };
+
+    struct {
+        std::string filename;
+        uint64_t offset;
+        uint64_t buff_size = 10'000;
+    } int_vector_buffer_params;
```
karasikov, line 97:
```suggestion
    struct int_vector_buffer_params {
        std::string filename;
        uint64_t offset;
        uint64_t buff_size = 10'000;
    };
```

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.cpp
```diff
@@ -0,0 +1,139 @@
...
+        outbuf.close();
+    };
+
+    bit_vector_small boundary(call_bits, num_rows + num_set_bits, num_rows);
+
+    outstream.open(filename, std::ios::in | std::ios::out | std::ios::binary | std::ios::ate);
+    boundary_start = outstream.tellp();
+    boundary.serialize(outstream);
+    outstream.seekp(boundary_start_pos, ios_base::beg);
+    outstream.write(reinterpret_cast<char *>(&boundary_start), sizeof(std::streampos));
```
karasikov, line 134: Since you have to load `boundary` anyway, why not serialize them before the other vector? Then you won't need to do another seek in load. In fact, you won't need to do any seek in load...

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+
+void RowSparseDisk::serialize(const std::function<void(binmat::BinaryMatrix::RowCallback)> &call_rows,
+                              const std::string &filename,
+                              uint64_t num_cols,
+                              uint64_t num_set_bits,
+                              uint64_t num_rows) {
+    std::ofstream outstream(filename,
+                            std::ios::binary | std::ios::ate
+                                    | std::ios::in); // std::ios_base::in needed ...
```
karasikov, line 97: These flags are not necessary at this point when the file is empty.
```suggestion
    std::ofstream outstream(filename, std::ios::binary); // std::ios_base::in needed ...
```
Or, is the file not empty? Then add a clarifying comment would be nice...

karasikov (changes requested):

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.hpp
marekkokot: This wouldn't compile, I mean I will need to create an instance of this struct. My idea was to just have an instance of an unnamed type since the type is not needed elsewhere. The idea was to just have a single instance keeping `int_vector_buffer` parameters together. Anyway, if this is somehow against metagraph's code style I will change it into:
```
struct int_vector_buffer_params_t {
        std::string filename;
        uint64_t offset;
        uint64_t buff_size = 10'000;
    };
int_vector_buffer_params_t int_vector_buffer_params;
```
Let me know.

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.cpp
marekkokot: Yeah, but the boundary is built in memory while writing rows data to disk via `int_vector_buffer`. If my understanding is correct it would require iterating rows twice to first build the boundary, store it and then iterate again to store rows' content. 
I assumed we don't want to pay the cost of the second iteration.
One option would be to leave the space for a boundary before rows content, yet I think we don't know (in advance) how many bytes are needed to store the boundary (to prepare an appropriate "hole" in the file), or is there a way to determine it?
Maybe there is another way that I just can't see. Any ideas?

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.cpp
marekkokot: The file is not empty, labels are serialized before. I have added clarifying comments.

metagraph/src/cli/config/config.hpp
marekkokot: Yes, this is for `int_vector_buffer`. I exposed this as a parameter because of the kinda unexpected behaviour (in terms of performance) depending on this value. I thought it may be convenient to leave this as a CLI parameter to be able to easily verify if changing this may speed up execution time.
Let me know if you want me to change it to the hardcoded. A lower number of CLI parameters is less confusing for the end-user.

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.hpp
karasikov: Why would it not compile?

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.cpp
karasikov: Yeah, I see the point. Would be impossible without a temp file. I guess it's better to keep it as is then. Though would be good to have a clarifying comment

metagraph/src/cli/config/config.hpp
karasikov: Ok, then let's figure out a name for this parameter.

This is the size of the buffer in int_vector_buffer. I guess it makes sense to use the same value for all int_vector_buffer instances that are used in the same way, that is, doing random access on them.
What about `RA_ivbuffer_size`?

metagraph/src/annotation/representation/annotation_matrix/annotation_matrix.cpp
```diff
@@ -61,7 +61,8 @@ ::serialize(const std::string &filename) const {
 
 template <class BinaryMatrixType, typename Label>
 bool StaticBinRelAnnotator<BinaryMatrixType, Label>::load(const std::string &filename) {
-    std::ifstream instream(make_suffix(filename, kExtension), std::ios::binary);
+    mtg::common::IfstreamWithNameAndOffset instream(make_suffix(filename, kExtension),
+                                                    std::ios::binary);
```
karasikov, line 65:
```suggestion
    common::IfstreamWithNameAndOffset instream(make_suffix(filename, kExtension),
                                               std::ios::binary);
```

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.hpp
```diff
@@ -0,0 +1,107 @@
...
+
+    bit_vector_small boundary_;
+    uint64_t num_columns_ = 0;
+    uint64_t num_rows_ = 0;
+};
+
+
```
karasikov, line 104:
```suggestion
```

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.hpp
```diff
@@ -0,0 +1,107 @@
...
+        uint64_t num_relations_impl() const { return set_bits_.size(); }
+        bool get(Row row, Column column) const;
+        BinaryMatrix::SetBitPositions get_row(Row row) const;
+        std::vector<SetBitPositions> get_rows(const std::vector<Row> &row_ids) const;
+        std::vector<BinaryMatrix::Row> get_column(uint64_t num_rows, Column column) const;
+
+      private:
+        const bit_vector_small &boundary_;
+        sdsl::int_vector_buffer<> set_bits_;
```
karasikov, line 90: Would `col_ids_` be more descriptive than `set_bits_`?

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.hpp
```diff
@@ -0,0 +1,107 @@
...
+    const bit_vector_small get_boundary() const { return boundary_; }
+
+  private:
+    // For the multithreading to work properly the idea is to open int_vector_buffer<> per
+    // each public method call that needs to work with int_vector_buffer<> The actual
+    // implementation is wrapped in the class below to assure it will not call any methods
+    // of RowSparseDisk class which could result in multiple opens of int_vector_buffer
+    class Impl {
+      public:
+        Impl(const bit_vector_small &boundary,
```
karasikov, line 75: Maybe, call it `View` or something like that?

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.hpp
```diff
@@ -0,0 +1,107 @@
...
+    SetBitPositions get_row(Row row) const override {
+        return Impl(boundary_, int_vector_buffer_params.filename,
+                    int_vector_buffer_params.offset, int_vector_buffer_params.buff_size)
+                .get_row(row);
+    }
+    std::vector<Row> get_column(Column column) const override {
+        return Impl(boundary_, int_vector_buffer_params.filename,
+                    int_vector_buffer_params.offset, int_vector_buffer_params.buff_size)
+                .get_column(num_rows(), column);
+    }
```
karasikov, line 38: Maybe also print a warning saying that this is an extremely inefficient operation and one should consider converting to a column-major format?

karasikov (changes requested):

metagraph/src/cli/query.cpp
```diff
@@ -1178,7 +1178,7 @@ int query_graph(Config *config) {
                                               config->suppress_unlabeled,
                                               config->verbose_output
                                                 || !(config->query_counts || config->query_coords),
-                                              anno_graph) << "\n";
+                                              anno_graph) + "\n";
```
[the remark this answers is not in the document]
marekkokot: Fixed

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.cpp
```diff
@@ -0,0 +1,141 @@
...
+        return false;
+    }
+
+    return true;
+}
+
+void RowSparseDisk::serialize(std::ostream & /*f*/) const {
+    throw std::runtime_error(
+            "RowSparseDisk::serialize not supproted, use static variant instead");
+}
```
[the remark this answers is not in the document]
marekkokot: Well, this is a good question and in fact at the beginning of my work I wanted to do this like other annotators, i.e. building in the constructor and then using serialize.
Maybe it would be possible but as it is disk-based annotation it would require that constructor write to disk. I mean the constructor would do something similar to what currently is done in static serialize, i.e. it would write int_vector_buffer with rows to the disk.
Then non-static serialize would rewrite it from disk to full annotation. In my opinion, not a good choice since in practice it would mean writing row content on disk twice.
I guess the alternative would be to store rows in memory when this loading ctor is used, but it doesn't seem to make sence to have a ctor that loads data to memory for a class that is disk-based.
Do you see another, better, way to do it more consistently or are we leaving it as is?
Maybe there are also some other, more implementation-related, issues to be considered here, but I think the one mentioned is quite important.

Hmm, now I see that you probably meant having both static and non-static variants. I guess that would be possible. I will try to add this.

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.hpp
marekkokot: Fixed

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.cpp
marekkokot: I have added some comments to the code, I hope it clarifies things

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.cpp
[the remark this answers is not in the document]
marekkokot: I implemented a non-static variant. The int_vector_buffer part of the annotation is just binary rewritten (that way I don't need to read the data with one int_vector_buffer and store it with another one, which would require to have a file name, which is not available in the non-static serialize and would require similar trick like in load (i.e. IfstreamWithName). Let me know if that's fine.

metagraph/src/common/ifstream_with_name_and_offset.hpp
```diff
@@ -0,0 +1,25 @@
...
+
+#include <fstream>
+#include <string>
+
+namespace mtg {
+namespace common {
+
+class IfstreamWithNameAndOffset : public std::ifstream {
+    std::string fname;
+    uint64_t offset {};
```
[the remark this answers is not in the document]
marekkokot: Fixed

metagraph/src/common/ifstream_with_name_and_offset.hpp
```diff
@@ -0,0 +1,25 @@
...
+
+  public:
+    explicit IfstreamWithNameAndOffset(const std::string &fname, std::ios_base::openmode mode)
```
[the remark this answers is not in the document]
marekkokot: You are right, I removed `explicit`.

metagraph/src/common/ifstream_with_name_and_offset.hpp
```diff
@@ -0,0 +1,25 @@
...
+        : std::ifstream(fname, mode), fname(fname) {}
+
+    const std::string &get_name() const { return fname; }
+
+    void set_offset(uint64_t offset) { this->offset = offset; }
+
+    uint64_t get_offset() const { return offset; }
```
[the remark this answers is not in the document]
marekkokot: That is true, I removed it and renamed the class and source file. It just stores file without offset.

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.hpp
marekkokot: Maybe, I think they are both similary descriptive. So I guess it's up to you. I used `set_bits_` because it is used in `row_sparse`. If you think `col_ids_` is better then maybe change also in `row_sparse`?

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.cpp
```diff
@@ -0,0 +1,141 @@
...
+        auto boundary_start = load_number(f);
+
+        _f->set_offset(static_cast<uint64_t>(f.tellg()));
+
+        int_vector_buffer_params.filename = _f->get_name();
+        int_vector_buffer_params.offset = _f->get_offset();
+
+        f.seekg(-sizeof(std::streampos), ios_base::end);
+
+        f.seekg(boundary_start, ios_base::beg);
```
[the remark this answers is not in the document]
marekkokot: Yeah, the first one is not needed, thanks! Fixed.

metagraph/src/annotation/annotation_converters.cpp
```diff
@@ -742,6 +743,445 @@ convert_to_RbBRWT<RbBRWTAnnotator>(const std::vector<std::string> &annotation_fi
...
+
+        size_t current_boundary_cost = 0;
+        for (; i < files.size(); ++i) {
+            uint64_t file_size = fs::file_size(files[i]);
+
+            if (file_size > mem_bytes) {
+                logger->warn("Not enough memory to process {}, requires {} MB, skipped",
+                             files[i], file_size / 1e6);
+                continue;
+            }
```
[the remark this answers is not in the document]
marekkokot: Thanks. I refactored `convert_to_row_sparse_disk` and `convert_row_diff_to_row_diff_sparse_disk` to a single, more general function, as they were almost identical.

metagraph/src/annotation/annotation_converters.cpp
```diff
@@ -742,6 +743,445 @@ convert_to_RbBRWT<RbBRWTAnnotator>(const std::vector<std::string> &annotation_fi
...
+    // compute global offsets (partial sums)
+    std::partial_sum(offsets.begin(), offsets.end(), offsets.begin());
+
+    ProgressBar progress_bar(num_rows, "Merge rows", std::cerr, !common::get_verbose());
+
+    RowSparseDisk::serialize(
+            [&](BinaryMatrix::RowCallback write_row) {
+// maybe better if single thread reads single annotation?
+#pragma omp parallel for ordered num_threads(num_threads) schedule(dynamic)
+                for (uint64_t i = 0; i < num_rows; i += kNumRowsInBlock) {
```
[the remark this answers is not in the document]
marekkokot: Thanks, my autoformat tool is doing it wrong, I will need to configure it properly.

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.hpp
marekkokot: Agree, but maybe the same should be added to `row_sparse` and possibly other row-major formats? I added warn for this annotation.

metagraph/src/annotation/binary_matrix/row_sparse_disk/row_sparse_disk.hpp
marekkokot: Done

metagraph/src/cli/config/config.hpp
marekkokot: Ok, I will rename it into `RA_ivbuffer_size` and make it an "advanced" parameter.
As I see `int_vector_buffer` is used in many places in the code. I guess most or all except row_disk annotations are not random access. Am I right? Are you aware of any random access use cases beyond `row_sparse_disk`?

metagraph/src/common/ifstream_with_name.hpp
```diff
@@ -0,0 +1,21 @@
...
+
+namespace mtg {
+namespace common {
+
+class IfstreamWithName : public std::ifstream {
+  public:
+    IfstreamWithName(const std::string &fname, std::ios_base::openmode mode)
+        : std::ifstream(fname, mode), fname_(fname) {}
+
+    const std::string &get_name() const { return fname_; }
```
karasikov, line 14:
```suggestion
    const std::string& get_name() const { return fname_; }
```

metagraph/src/annotation/annotation_converters.cpp
```diff
@@ -1195,12 +1175,19 @@ void convert_to_row_diff<IntRowDiffDiskAnnotator>(
 
     std::mutex mtx;
 
+    size_t max_val = 0;
+
     ColumnCompressed<>::load_columns_and_values(
             files,
             [&](size_t column_index, const std::string &label, std::unique_ptr<bit_vector> &&column,
                 sdsl::int_vector<> &&values) {
+                    size_t local_max_val = *std::max_element(values.begin(), values.end());
```
karasikov, line 1184: Can `values` be empty?

marekkokot: Oh, nice catch, thanks! Fixed with 514b256e267bef93828e40ee62a6c281c2e74f9f

metagraph/src/annotation/annotation_converters.cpp
```diff
@@ -751,6 +786,602 @@ convert_to_RbBRWT<RbBRWTAnnotator>(const std::vector<std::string> &annotation_fi
...
+                if (num_rows != column->size()) {
+                    logger->error("Annotations have different number of rows");
+                    std::exit(1);
+                }
+
+                num_set_bits += column->num_set_bits();
+                if (columns.size() <= column_index) {
+                    columns.resize(column_index + 1);
+                    col_names.resize(column_index + 1);
+                }
```
karasikov, line 816: This would cause a race condition. I wonder why this is never caught in unit tests. Do we test parallel conversion?

metagraph/src/annotation/annotation_converters.cpp
```diff
@@ -751,6 +786,602 @@ convert_to_RbBRWT<RbBRWTAnnotator>(const std::vector<std::string> &annotation_fi
...
+            uint64_t file_size = fs::file_size(files[i]);
+
+            if (file_size > mem_bytes) {
+                logger->warn("Not enough memory to process {}, requires {} MB, skipped",
+                             files[i], file_size / 1e6);
+                continue;
+            }
+
+            size_t new_predicted_boundary_cost
+                    = bit_vector_small::predict_size(num_rows * (file_batch.size() + 1)
```
karasikov, line 1052: Why ` + 1`?

metagraph/src/annotation/annotation_converters.cpp
karasikov: Ah, I see a lock at the top now. All good.

karasikov (approved):

PR #424, opened 2022-08-08
Support streaming of annotation columns in differential assembly

hmusta:


metagraph/src/annotation/representation/column_compressed/column_compressed_lazy.hpp
```diff
@@ -0,0 +1,51 @@
...
+    bool has_label(Index, const Label &) const override final { throw std::runtime_error("has_label not implemented"); }
+    bool has_labels(Index, const VLabels &) const override final { throw std::runtime_error("has_labels not implemented"); }
+    void insert_rows(const std::vector<Index> &) override final { throw std::runtime_error("insert_rows not implemented"); }
+    void rename_labels(const tsl::hopscotch_map<Label, Label> &) override final { throw std::runtime_error("rename_labels not implemented"); }
+    size_t num_labels() const override final { throw std::runtime_error("num_labels not implemented"); }
+    uint64_t num_relations() const override final { throw std::runtime_error("num_relations not implemented"); }
+    void call_objects(const Label &, std::function<void(Index)>) const override final { throw std::runtime_error("call_objects not implemented"); }
+    std::string file_extension() const override final { throw std::runtime_error("file_extension not implemented"); }
+    const binmat::BinaryMatrix& get_matrix() const override final { throw std::runtime_error("get_matrix not implemented"); }
```
karasikov, line 41: Why do you make this an instance of `MultiLabelEncoded` if almost none of its methods are implemented?

karasikov (changes requested):

metagraph/src/annotation/representation/column_compressed/column_compressed_lazy.hpp
hmusta: I'm using this with `AnnotatedDBG` (needed by the differential assembly code), which needs `MultiLabelEncoded`. This would have needed a much bigger refactor otherwise.

hmusta: What would have been nice is if the current interface for MultiLabelEncoded had a function for calling each (column label, column bit_vector) pair.

karasikov: No, this should definitely not be MultiLabelEncoded if it only implements 5% of its methods.

karasikov: You can introduce a higher level abstraction, which has a couple of methods, and use that. This refactoring should be minor.

hmusta: I added a new level simply called `Annotation`

PR #426, opened 2022-08-09
Column streaming for differential assembly

hmusta:


karasikov:
Is this ready for review?

hmusta:
> Is this ready for review?

Yes

metagraph/src/graph/annotated_dbg.hpp
```diff
@@ -61,7 +71,8 @@ class AnnotatedSequenceGraph {
 
   protected:
     std::shared_ptr<SequenceGraph> graph_;
-    std::unique_ptr<Annotator> annotator_;
+    std::unique_ptr<Annotation> annotation_;
+    const std::shared_ptr<Annotator> annotator_;
```
karasikov, line 75: Will it work if you pass lazy annotator to AnnotatedDBG? This points to a design issue.
Maybe it's better to do without AnnotatedDBG in differential assembly? Better pass graph and annotation separately. Then you can pass lazy annotation in any type you want.

karasikov (changes requested):

hmusta:
Superseded by #487

PR #430, opened 2022-09-08
Added unit tests for RowDiff

hmusta:


hmusta:
Some of the `AnnotatedDBG` unit tests seem to not be working with RowDiff annotators. @karasikov could you please take a look and see if the conversion is being done correctly in `build_anno_graph`?

metagraph/tests/annotation/test_annotated_dbg_helpers.cpp
```diff
@@ -7,65 +7,184 @@
...
+        const std::string fork_succ_file = graph_fname + annot::binmat::kRowDiffForkSuccExt;
+        if (!std::filesystem::exists(anchors_file)) {
+            logger->error("Anchor bitmap {} does not exist.", anchors_file);
+            std::exit(1);
+        }
+        if (!std::filesystem::exists(fork_succ_file)) {
+            logger->error("Fork successor bitmap {} does not exist", fork_succ_file);
+            std::exit(1);
+        }
+
```
karasikov, line 117: I would look closely and double-check this function. The row-diff converters are tested pretty well, so I doubt there is any bug there.

hmusta: Do you have any ideas about what could be wrong? I've made a few tweaks (explicitly setting the `row_count` and `row_reduction` file extensions, increasing the amount of available RAM, tweaking the max path length, setting separate output and swap dirs), but none of these seem to fix things.

karasikov: Does this graph have a mask for dummy k-mers?

hmusta: It does, but removing the mask doesn't seem to fix the issue.

metagraph/tests/annotation/test_annotated_dbg_helpers.cpp
```diff
@@ -1,71 +1,228 @@
...
+                                static_cast<RowDiffStage>(1), out_path + ".row_reduction", false, coordinates);
+            convert_to_row_diff(files, graph_fname, 100000000, 5, tmp_dir, swap_dir,
+                                static_cast<RowDiffStage>(2), out_path, false, coordinates);
+        }
+
+        auto annotator = std::make_unique<ColumnCompressed<>>(0);
+        if (!annotator->merge_load(files)) {
+            logger->error("Cannot load annotations");
+            exit(1);
+        }
```
karasikov, line 139: So you converted it to row-diff but then load again the same untransformed columns... Am I missing something?

hmusta: I'm only using this for loading. Below, I'm wrapping the underlying columns in a RowDiff matrix.

karasikov: So you wrap the original untransformed columns as row-diff?

karasikov: For plain row-diff without coordinates, it writes to `.row_diff` columns. So you can't load them as `ColumnCompressed<>`. See how it's done in the main converter.

hmusta: Ahh, I see, that's confusing... I had previously done it this way for annotations with coordinates, so I thought it would also work the same way without coordinates.

hmusta: That sounds like it should fix the issue, thanks!

PR #433, opened 2023-01-11
sd_vector_disk

karasikov:
SD vector with the vector of the least significant bits of set bits' positions stored on disk.
Takes only $\approx 2$ bits of RAM per set bit. The remaining $\lceil \log_2 n\rceil - \lceil \log_2 m\rceil$ bits are stored on disk.

Run benchmarks:
```
./benchmarks --benchmark_filter=BM_bv_query_sequential_sd_vector_disk_acce*
```

Results:

m_low in int_vector_buffer:
```
BM_bv_query_random_sd_vector_disk_access_every_nth_bit_set<10000>       5.26 us         5.17 us       100000 RAM=1.35269G
BM_bv_query_random_sd_vector_disk_access_every_nth_bit_set<3000>        4.25 us         4.22 us       160777 RAM=1.45588G
BM_bv_query_random_sd_vector_disk_access_every_nth_bit_set<1000>        53.0 us         18.2 us        32213 RAM=1.64712G
BM_bv_query_random_sd_vector_disk_access_every_nth_bit_set<500>         17.8 us         10.4 us       101077 RAM=1.43773G
BM_bv_query_random_sd_vector_disk_access_every_nth_bit_set<200>          376 us         51.3 us        11791 RAM=2.58662G
```

m_low in int_vector_mapper:
```
BM_bv_query_random_sd_vector_disk_access_every_nth_bit_set<10000>      0.235 us        0.233 us      2828471 RAM=1.5224G
BM_bv_query_random_sd_vector_disk_access_every_nth_bit_set<3000>       0.450 us        0.445 us      1135598 RAM=1.91674G
BM_bv_query_random_sd_vector_disk_access_every_nth_bit_set<1000>        1.17 us         1.17 us       545515 RAM=2.73401G
BM_bv_query_random_sd_vector_disk_access_every_nth_bit_set<500>         68.7 us         10.7 us        66802 RAM=2.24338G
^C

```

m_low and m_high in int_vector_mapper:
```
BM_bv_query_random_sd_vector_disk_access_every_nth_bit_set<10000>      0.276 us        0.274 us      2332459 RAM=1.52114G
BM_bv_query_random_sd_vector_disk_access_every_nth_bit_set<3000>       0.413 us        0.412 us      1216758 RAM=1.92005G
BM_bv_query_random_sd_vector_disk_access_every_nth_bit_set<1000>        1.11 us         1.11 us       521963 RAM=2.69745G
BM_bv_query_random_sd_vector_disk_access_every_nth_bit_set<500>         28.5 us         6.71 us       256609 RAM=2.83147G
BM_bv_query_random_sd_vector_disk_access_every_nth_bit_set<200>          197 us         32.7 us        26774 RAM=2.71836G
```

all with mmap:
```
BM_bv_query_random_sd_vector_mmap_access_every_nth_bit_set<10000>      0.285 us        0.284 us      2025029 RAM=1.52162G
BM_bv_query_random_sd_vector_mmap_access_every_nth_bit_set<3000>       0.396 us        0.394 us      1432762 RAM=1.92811G
BM_bv_query_random_sd_vector_mmap_access_every_nth_bit_set<1000>        1.68 us         1.67 us       334188 RAM=2.52352G
BM_bv_query_random_sd_vector_mmap_access_every_nth_bit_set<500>         4.12 us         4.09 us       134115 RAM=2.61806G
BM_bv_query_random_sd_vector_mmap_access_every_nth_bit_set<200>        430 us         57.3 us         8541 RAM=954.814M
```

all in RAM:
```
BM_bv_query_random_sd_vector_access_every_nth_bit_set<10000>      0.231 us        0.227 us      2777150 RAM=1.51786G
BM_bv_query_random_sd_vector_access_every_nth_bit_set<3000>       0.268 us        0.267 us      2635572 RAM=1.91844G
BM_bv_query_random_sd_vector_access_every_nth_bit_set<1000>       0.342 us        0.339 us      2106042 RAM=2.89771G
BM_bv_query_random_sd_vector_access_every_nth_bit_set<500>        0.373 us        0.371 us      1886208 RAM=4.23003G
BM_bv_query_random_sd_vector_access_every_nth_bit_set<200>        0.581 us        0.566 us      1703989 RAM=6.93669G
```

The times fluctuate quite a bit, it would be good to also benchmark it on the cluster.


metagraph/tests/test_bit_vector.cpp
```diff
@@ -573,6 +574,113 @@ TEST(bit_vector_sd, CheckIfInverts) {
...
+void compare(const sdsl::sd_vector<> &first, const sdsl::sd_vector_disk<> &second) {
+    ASSERT_EQ(first.size(), second.size());
+
+    sdsl::sd_vector<>::select_1_type first_slct(&first);
+    sdsl::sd_vector<>::rank_1_type first_rank(&first);
+
+    sdsl::sd_vector_disk<>::select_1_type second_slct(&second);
+    sdsl::sd_vector_disk<>::rank_1_type second_rank(&second);
+
+    uint64_t m = first_rank(first.size());
```
hmusta, line 593: maybe also add a test here to make sure that `first_rank(first.size()) == second_rank(second.size())`?

metagraph/src/common/vectors/sd_vector_disk.hpp
```diff
@@ -0,0 +1,530 @@
...
+#include <sdsl/sd_vector.hpp>
+#include <sdsl/int_vector.hpp>
+#include <sdsl/int_vector_buffer.hpp>
+#include <sdsl/select_support_mcl.hpp>
+#include <sdsl/util.hpp>
+#include <sdsl/iterators.hpp>
+
+// Based on https://github.com/simongog/sdsl-lite/blob/master/include/sdsl/sd_vector.hpp
+
+// FYI: This implementation does NOT support multithreading
```
hmusta, line 15: can you write `TODO` here instead of `FYI` so it'll be easier to find parts of the code that can be improved in the future?

metagraph/src/common/vectors/sd_vector_disk.hpp
```diff
@@ -0,0 +1,530 @@
...
+         class t_select_1     = typename t_hi_bit_vector::select_1_type,
+         class t_select_0     = typename t_hi_bit_vector::select_0_type>
+class rank_support_sd_disk;  // in sd_vector
+
+// forward declaration needed for friend declaration
+template<uint8_t t_b          = 1,
+         class t_hi_bit_vector= bit_vector,
+         class t_select_1     = typename t_hi_bit_vector::select_1_type,
+         class t_select_0     = typename t_hi_bit_vector::select_0_type>
+class select_support_sd_disk;  // in sd_vector
```
hmusta, line 31: how much would it break things if we make the storage type of `m_low` and `m_high` template parameters in `sdsl`? we don't have to do this here, but keep this in mind for the future.

metagraph/src/common/vectors/sd_vector_disk.hpp
```diff
@@ -0,0 +1,530 @@
...
+                throw std::runtime_error("sd_vector_builder: requested capacity is larger than vector size.");
+
+            size_type logm = bits::hi(m_capacity) + 1;
+            const size_type logn = bits::hi(m_size) + 1;
+            if(logm == logn)
+            {
+                --logm; // to ensure logn-logm > 0
+                assert(logn - logm > 0);
+            }
+            m_wl = logn - logm;
```
hmusta, line 83: does this class still work if compressing a vector where all bits are set?

hmusta (changes requested):

metagraph/src/common/vectors/sd_vector_disk.hpp
karasikov: Yeah, it should. And it's in the tests, isn't it?

metagraph/src/common/vectors/sd_vector_disk.hpp
karasikov: `m_high` is already a template (`t_hi_bit_vector`). Making `m_low` a template doesn't make much sense, there is too much code specific to `int_vector_buffer`. So a good way would be to make a parent class and move all shared code there but for this, we would need to rewrite sd_vector in sdsl.

metagraph/src/common/vectors/sd_vector_disk.hpp
karasikov: It's not clear what's the best way to support multithreading. Memory mapping, probably. It naturally supports multithreading and may be even faster.

metagraph/src/common/vectors/sd_vector_disk.hpp
hmusta: I couldn't find a test like that in sdsl. The tests I found fill the vectors randomly, e.g.,
https://github.com/simongog/sdsl-lite/blob/master/test/bit_vector_test.cpp

We do have this in metagraph:
https://github.com/ratschlab/metagraph/blob/master/metagraph/tests/test_bit_vector.cpp#L195

But the problem is that it's tested with `bit_vector_sd`, which I believe defaults to storing the bit-flipped version of the vector if the density is high enough. So again, I'm not sure if this gets tested.

karasikov: There are specific tests for sd_vector_disk. BTW, it's implemented in metagraph, not sdsl. Hence tested here.
Also, note that it's never inverted.
https://github.com/ratschlab/metagraph/pull/433/files#diff-44088ac3b4ca1eb7a80c3d8cdc12044299432149e4d3eb23b6c6cda0aea22ba6R637

hmusta: ok, looks good

hmusta: thanks for the clarification

karasikov:
I figured out how to easily load all vectors with memory mapping (will implement in another PR), so I'm only gonna keep the constructor for sd_vector with streaming here.

PR #434, opened 2023-01-22
support memory mapping

karasikov:
Added support for loading `sdsl::int_vector<*>` with memory mapping. So practically all bit/int-vectors and other data structures in metagraph become memory mapped, including the main graph representations as well as almost all annotation representations.

OFF by default. To turn on, add the `--mmap` flag.

### DIFF
--- a/metagraph/src/graph/representation/succinct/dbg_succinct.cpp
+++ b/metagraph/src/graph/representation/succinct/dbg_succinct.cpp
@@ -671,7 +671,6 @@ bool DBGSuccinct::load(const std::string &filename) {
     auto prefix = utils::remove_suffix(filename, kExtension);
 
     std::unique_ptr<std::ifstream> in = utils::open_ifstream(prefix + kDummyMaskExtension);
-
     if (!in->good())
         return true;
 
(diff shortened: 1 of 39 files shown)


hmusta:
Some parts that could still use mmap:
- [ ] ColumnCompressed: *load* methods
- [ ] RowCompressed: *load* methods
- [ ] StaticBinRelAnnotator: merge_load_row_diff, load_coords
- [ ] Elias Fano methods?
- [ ] RowDiff: load_* and builder
- [ ] Chunk-based graph builders?
- [ ] Node weights
- [ ] BOSS: load(std::string) (do we still use this?), chunk loading
- [ ] BRWT builders

metagraph/src/graph/representation/succinct/dbg_succinct.cpp
```diff
@@ -648,14 +649,15 @@ bool DBGSuccinct::load_without_mask(const std::string &filename) {
...
-        if (!boss_graph_->load(instream))
+        if (!boss_graph_->load(*in))
             return false;
 
-        mode_ = static_cast<Mode>(load_number(instream));
+        mode_ = static_cast<Mode>(load_number(*in));
 
-        if (!boss_graph_->load_suffix_ranges(instream))
+        if (!boss_graph_->load_suffix_ranges(*in))
             logger->warn("No index for node ranges could be loaded");
```
hmusta, line 661: Could you please double-check this? I'm noticing that this warning is being printed for all graphs

hmusta (changes requested):

metagraph/src/graph/representation/succinct/dbg_succinct.cpp
karasikov: This is a different warning...

hmusta: false alarm

PR #445, opened 2023-04-06
switch CI to ubuntu-20.04 LTS, include HTSlib as a submodule

hmusta:


metagraph/CMakeLists.txt
```diff
@@ -331,8 +334,18 @@ add_library(mersenne_twister STATIC
 add_library(sdust STATIC
   external-libraries/sdust/sdust.c
 )
+
 target_compile_options(sdust PRIVATE -Wall -Wc++-compat -Wno-error)
 
+include(ExternalProject)
+ExternalProject_Add(HTSLIB_Project
+  SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/external-libraries/htslib
+  CONFIGURE_COMMAND autoreconf -i ${CMAKE_CURRENT_SOURCE_DIR}/external-libraries/htslib && ${CMAKE_CURRENT_SOURCE_DIR}/external-libraries/htslib/configure --prefix=${CMAKE_CURRENT_SOURCE_DIR}/external-libraries/htslib --disable-libcurl --disable-bz2 --disable-lzma
```
karasikov, line 343: Why no bz2?

hmusta: It was having linking errors at some point, but I'll try it again now to see if it works.

hmusta: So it seems like it fails to link if I enable BZ2. According to HTSlib's documentation, bz2 is only used for loading cram files, so I think it should be fine to omit this.

.github/workflows/main.yml
```diff
@@ -139,7 +129,17 @@ jobs:
...
+    - name: run unit tests
+      if: matrix.build_static == 'OFF'
+      run: |
+        export LD_LIBRARY_PATH="/usr/local/lib/:$LD_LIBRARY_PATH"
+        cd metagraph/build && ./unit_tests
 
+    - name: run integration tests
+      run: |
+        export LD_LIBRARY_PATH="/usr/local/lib/:$LD_LIBRARY_PATH"
+        cd metagraph/build && make check
```
karasikov, line 142: why did you have to move these steps?

hmusta: It helps with debugging. If any of the test segfault and I can't reproduce the issue locally, this allows me to download the binary and run that instead.

PR #447, opened 2023-05-11
Replace CMake config file for KMC

hmusta:


karasikov:
Should we close this?

hmusta:
> Should we close this?

Sure, we can clean this up later, it's a low priority

PR #453, opened 2023-06-05
Less stringent alignment and seed filtering

hmusta:
- No default limit on the total number of output alignments
- If a seed has a filtered-out label, report the seed as an alignment without extending it
- Annotate dummy node seeds with labels of their nearest annotated nodes
- Introduce alignment chaining from MetaGraph-MLA

TODO
- [x] Merge reported seeds into MEMs (since it outputs too much)
- [x] When merging seeds with coordinates, ensure that the coordinates are consistent
- [x] Figure out why some extensions are not happening
- [x] Alignment is too slow and timing out, figure out why
- [x] Chain together filtered-out seeds to cheaply get longer alignments


metagraph/src/graph/alignment/aligner_seeder_methods.cpp
```diff
@@ -166,7 +166,8 @@ void SuffixSeeder<BaseSeeder>::generate_seeds() {
     }
 
     const DBGSuccinct &dbg_succ = get_base_dbg_succ(&this->graph_);
-    assert(!dbg_succ.get_mask());
+    if (dbg_succ.get_mask())
+        logger->warn("Graph has a dummy k-mer mask. Seeds containing dummy k-mers will be missed.");
```
karasikov, line 167: Seems like it's better to have the mask missing? If so, can't you just ignore it?

hmusta: Technically I can, but this would overcomplicate the code because I would have to have some way to know whether an index stored in a seed represents a BOSS edge or a DBGSuccinct node.

PR #454, opened 2023-06-12
deprecate adj-rc index, optimize and fix alignment to CanonicalDBG

karasikov:


metagraph/src/graph/graph_extensions/node_rc.cpp
```diff
@@ -229,6 +230,60 @@ void NodeRC::call_incoming_from_rc(node_index node,
...
+    } else {
+        adjacent_incoming_from_rc(node, [&](node_index prev) {
+            char c = graph_->get_node_sequence(prev)[0];
+            if (c == boss::BOSS::kSentinel)
+                return;
+
+            callback(prev, complement(c));
+        });
+    }
+}
```
karasikov, line 307: I suggest reimplementing this. That is, remove the calls to `adjacent_incoming_from_rc`.
Also, use `NodeFirstCache::call_incoming_kmers`?

metagraph/src/graph/graph_extensions/node_rc.cpp
```diff
@@ -58,13 +58,29 @@ NodeRC::NodeRC(const DeBruijnGraph &graph, bool construct_index) : graph_(&graph
         auto it = rc_edges.begin();
         for (size_t i = 0; i < path.size(); ++i, ++it) {
             if (*it) {
-                std::lock_guard<std::mutex> lock(mu);
-                rc_nodes_prefix.emplace_back(path[i]);
+                bool found = false;
+                boss.call_outgoing(*it, [&](edge_index adjacent_edge) {
+                    found |= (dbg_succ->boss_to_kmer_index(adjacent_edge) != DeBruijnGraph::npos);
```
karasikov, line 63: This will only work with an edge mask in dbg_succinc. You actually need to check if the last character is not dummy

metagraph/src/graph/graph_extensions/node_rc.cpp
```diff
@@ -215,15 +252,69 @@ void NodeRC::call_incoming_from_rc(node_index node,
...
+    assert(graph_);
+
+    if (const auto *dbg_succ_ = dynamic_cast<const DBGSuccinct*>(graph_)) {
+        const auto &boss = dbg_succ_->get_boss();
+        auto first_cache = dbg_succ_->get_extension<NodeFirstCache>();
+        adjacent_incoming_from_rc(node, [&](node_index prev) {
+            char c = first_cache
+                ? first_cache->get_first_char(prev)
+                : boss.decode(boss.get_minus_k_value(
+                    dbg_succ_->kmer_to_boss_index(prev), boss.get_k() - 1).first);
```
karasikov, line 291: `get_first_char` could take a child hint. Really, better call `NodeFirstCache::call_incoming_kmers`.

metagraph/src/graph/graph_extensions/node_rc.cpp
```diff
@@ -74,7 +75,10 @@ NodeRC::NodeRC(const DeBruijnGraph &graph, bool construct_index) : graph_(&graph
                     boss.bwd(*(it + 1)),
                     boss.get_node_last_value(*(it + 1)),
                     [&](edge_index incoming_boss_edge) {
-                        found |= (dbg_succ->boss_to_kmer_index(incoming_boss_edge) != DeBruijnGraph::npos);
+                        // TODO: make this more efficient?
+                        found |= (boss.get_minus_k_value(incoming_boss_edge,
```
karasikov, line 79: yeah, I guess it's better to precompute the mask and check with that as in traversal functions

metagraph/src/graph/graph_extensions/node_rc.cpp
```diff
@@ -32,6 +32,14 @@ NodeRC::NodeRC(const DeBruijnGraph &graph, bool construct_index) : graph_(&graph
         return;
 
     const BOSS &boss = dbg_succ->get_boss();
+    const auto *mask = dbg_succ->get_mask();
+    std::unique_ptr<bit_vector_stat> computed_mask;
+    if (!mask) {
+        auto sdsl_mask = boss.mark_all_dummy_edges(get_num_threads());
+        sdsl_mask.flip();
+        computed_mask = std::make_unique<bit_vector_stat>(std::move(sdsl_mask));
+        mask = computed_mask.get();
```
karasikov, line 41: No need to set it, just use the sdsl bitmap.
Actually if you do set it, it changes the node indexes

hmusta: this isn't setting it, actually, it's only using what's computed by `mark_all_dummy_edges`.

If it were setting the mask, this would not compile

metagraph/src/graph/graph_extensions/node_rc.cpp
```diff
@@ -31,10 +32,19 @@ NodeRC::NodeRC(const DeBruijnGraph &graph, bool construct_index) : graph_(&graph
         return;
 
     const BOSS &boss = dbg_succ->get_boss();
+    const auto *mask = dbg_succ->get_mask();
+    std::unique_ptr<bit_vector_stat> computed_mask;
+    if (!mask) {
+        auto sdsl_mask = boss.mark_all_dummy_edges(get_num_threads());
+        sdsl_mask.flip();
+        computed_mask = std::make_unique<bit_vector_stat>(std::move(sdsl_mask));
```
karasikov, line 40: So bit_vector_stat computes rank and select support which are unnecessary here.
Better use sdsl::bit_vector

hmusta: do you think it's good now? The only unnecessary overhead is that it now computes the number of set bits

karasikov: but why not use the sdsl bitmap?

karasikov: never mind, it's good

metagraph/src/graph/graph_extensions/node_rc.cpp
hmusta: done

metagraph/src/graph/representation/canonical_dbg.hpp
```diff
@@ -141,6 +142,10 @@ class CanonicalDBG : public DBGWrapper<DeBruijnGraph> {
 
     std::array<size_t, 256> alphabet_encoder_;
 
+    // a thread-safe 0-size NodeFirstCache to use as a fallback if another NodeFirstCache
+    // extension is not available
+    mutable std::unique_ptr<NodeFirstCache> fallback_cache_;
```
karasikov, line 147: Why not create a new cache for each CanonicalDBG wrapper?

metagraph/src/graph/graph_extensions/node_rc.cpp
hmusta: done

metagraph/src/graph/representation/canonical_dbg.hpp
hmusta: In the ideal case (see align.cpp), one CanonicalDBG wrapper and one NodeFirstCache exists for each thread. However, if a CanonicalDBG is being used by multiple threads, then we don't create a NodeFirstCache for it, so this acts as a fallback to do the same operations cache-free.

karasikov: You could create multiple CanonicalDBG wrappers, one for each thread, and each of the wrappers would have its cache in it.

hmusta: not really. the caches are accessed very often, so having a shared cache for each thread creates too much contention

karasikov: Hm, I'd think it's way less than fwd/bwd calls that it replaces. Does it really affect the performance that much?

hmusta: there are two things to consider here:
1) when I tested it last time, the CPU usage saturated at around 4000%, so it was pretty severe
2) the cache is most effective locally. 10 threads aligning different read batches are very unlikely to be accessing the same part of the graph within a reasonable timeframe of each other, so it doesn't make much sense to have a shared cache

karasikov: If you create a cache for each thread, every new query will start with an empty cache. Though, the queries might be actually similar... What do you think?

hmusta: I tried this out on a machine with 128 threads. It looks like saturates at 16 threads when having a shared cache, but we don't know when it saturates for the separate caches (I assume that it slows down at 128 threads because of hyperthreading)

```
Nthreads	Separate caches							Shared cache
128		Elapsed (wall clock) time (h:mm:ss or m:ss): 0:41.13		Elapsed (wall clock) time (h:mm:ss or m:ss): 0:56.94
64		Elapsed (wall clock) time (h:mm:ss or m:ss): 0:37.88		Elapsed (wall clock) time (h:mm:ss or m:ss): 0:56.27
32		Elapsed (wall clock) time (h:mm:ss or m:ss): 0:38.86		Elapsed (wall clock) time (h:mm:ss or m:ss): 0:51.65
16		Elapsed (wall clock) time (h:mm:ss or m:ss): 0:46.97		Elapsed (wall clock) time (h:mm:ss or m:ss): 0:49.63
8		Elapsed (wall clock) time (h:mm:ss or m:ss): 1:16.69		Elapsed (wall clock) time (h:mm:ss or m:ss): 1:11.30
4		Elapsed (wall clock) time (h:mm:ss or m:ss): 2:19.48		Elapsed (wall clock) time (h:mm:ss or m:ss): 2:03.94
2		Elapsed (wall clock) time (h:mm:ss or m:ss): 4:21.65		Elapsed (wall clock) time (h:mm:ss or m:ss): 3:46.17
1		Elapsed (wall clock) time (h:mm:ss or m:ss): 9:05.37		Elapsed (wall clock) time (h:mm:ss or m:ss): 7:37.99
```

karasikov: Hard to say. Would be interesting to compare, single cache vs multiple for 64 threads

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -577,5 +564,171 @@ void CanonicalDBG::reverse_complement(std::string &seq,
...
+}
+
+void CanonicalDBG
+::call_incoming_from_rc(node_index node,
+                        const std::function<void(node_index, char)> &callback,
+                        const std::string &spelling_hint) const {
+    if (get_dbg_succ(*graph_)) {
+        auto *cache = get_extension_threadsafe<NodeFirstCache>();
+        if (!cache)
+            cache = fallback_cache_.get();
```
karasikov, line 715: Add method `get_cache` to remove code duplication

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -577,5 +564,171 @@ void CanonicalDBG::reverse_complement(std::string &seq,
...
+        }, spelling_hint);
+
+        return;
+    }
+
+    adjacent_outgoing_from_rc(node, [&](node_index next) {
+        char c = graph_->get_node_sequence(next).back();
+        if (c != boss::BOSS::kSentinel)
+            callback(next, c);
+    }, spelling_hint);
```
karasikov, line 705: I think it would be easier to read if it was in `} else {`. Same in other functions.

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -577,5 +564,171 @@ void CanonicalDBG::reverse_complement(std::string &seq,
...
+            if (c != boss::BOSS::kSentinel)
+                callback(next, c);
+        }, spelling_hint);
+
+        return;
+    }
+
+    adjacent_outgoing_from_rc(node, [&](node_index next) {
+        char c = graph_->get_node_sequence(next).back();
+        if (c != boss::BOSS::kSentinel)
```
karasikov, line 703: if it's not SuccinctDBG, it doesn't make sense to compare to `boss::BOSS::kSentinel`. Does this check even make sense?

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -577,5 +564,171 @@ void CanonicalDBG::reverse_complement(std::string &seq,
...
+
+    const auto &alphabet = graph_->alphabet();
+    for (size_t c = 0; c < alphabet.size(); ++c) {
+        if (alphabet[c] != boss::BOSS::kSentinel) {
+            rev_seq[0] = complement(alphabet[c]);
+            node_index next = graph_->kmer_to_node(rev_seq);
+            if (next != DeBruijnGraph::npos)
+                callback(next, 0);
+        }
+    }
```
karasikov, line 683: Do in an `else` block. Otherwise it's easy to miss `return` above and assume that it's the ending of the function.
In general, try no to do `return` in super large `if` blocks like here.

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -577,5 +564,171 @@ void CanonicalDBG::reverse_complement(std::string &seq,
...
+            rev_seq.back() = complement(alphabet[c]);
+            node_index prev = graph_->kmer_to_node(rev_seq);
+            if (prev != DeBruijnGraph::npos)
+                callback(prev);
+        }
+    }
+}
+
+void CanonicalDBG
+::adjacent_incoming_from_rc(node_index node,
```
karasikov, line 623: How this is "incoming" I still don't get... This is confusing

karasikov: Add the comment above `// ATGGCT -> TGGCT* -> *AGCCA`
Without this comment, it's impossible to guess what this function does

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -577,5 +564,171 @@ void CanonicalDBG::reverse_complement(std::string &seq,
...
+                if (next == DeBruijnGraph::npos)
+                    return;
+
+                if (is_dummy) {
+                    char c = (cache ? cache : fallback_cache_.get())->get_first_char(
+                        incoming_boss_edge, rc_edge
+                    );
+                    if (c == boss::BOSS::kSentinel)
+                        return;
+                }
```
karasikov, line 660: I don't get the logic here. Can you add some comments with examples?

metagraph/src/graph/representation/canonical_dbg.cpp
hmusta: I'll make this an assert instead, this was done for futureproofing in case we ever incorporate dummy k-mers into the other representations.

metagraph/src/graph/representation/canonical_dbg.cpp
hmusta: I added some extra comments

hmusta: Also, I renamed these functions so the header looks like this
```
    // Find incoming nodes that are on the reverse complement strand of node
    void adjacent_incoming_rc_strand(node_index node,
                                     const std::string &spelling_hint) const;
    void call_incoming_rc_strand(node_index node,
                                 const std::function<void(node_index, char)> &callback,
                                 const std::string &spelling_hint) const;

    // Find outgoing nodes that are on the reverse complement strand of node
    void adjacent_outgoing_rc_strand(node_index node,
                                     const std::function<void(node_index, uint64_t)> &callback,
                                     const std::string &spelling_hint) const;
    void call_outgoing_rc_strand(node_index node,
                                 const std::function<void(node_index, char)> &callback,
                                 const std::string &spelling_hint) const;
```

hmusta: So now `*_incoming_rc_strand` is called by `call_incoming_kmers` and `*_outgoing_rc_strand` is called by `call_outgoing_kmers`.

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -721,13 +715,13 @@ ::call_incoming_from_rc(node_index node,
         }, spelling_hint);
 
         return;
```
karasikov, line 725: no need for return

metagraph/src/graph/representation/canonical_dbg.hpp
```diff
@@ -112,6 +141,28 @@ class CanonicalDBG : public DBGWrapper<DeBruijnGraph> {
...
+                                 const std::function<void(node_index, char)> &callback,
+                                 const std::string &spelling_hint) const;
+
+    // Find outgoing nodes that are on the reverse complement strand of node
+    void adjacent_outgoing_rc_strand(node_index node,
+                                     const std::function<void(node_index, uint64_t)> &callback,
+                                     const std::string &spelling_hint) const;
+    void call_outgoing_rc_strand(node_index node,
+                                 const std::function<void(node_index, char)> &callback,
+                                 const std::string &spelling_hint) const;
```
karasikov, line 163: There are so many of these functions... Do we really need all of them?

hmusta: Having a separate `adjacent_outgoing_nodes` (which the extra functions are meant to support) is used by #453 to find the descendants of dummy k-mers when annotating dummy k-mer seeds.

metagraph/src/graph/representation/canonical_dbg.hpp
```diff
@@ -57,17 +58,45 @@ class CanonicalDBG : public DBGWrapper<DeBruijnGraph> {
...
+
+
+    // Given a node index, call the target nodes of all edges outgoing from it.
+    virtual void adjacent_outgoing_nodes_hint(node_index node,
+                                              const std::function<void(node_index)> &callback,
+                                              const std::string &spelling_hint) const;
+
+    virtual void call_outgoing_kmers_hint(node_index kmer,
+                                          const OutgoingEdgeCallback &callback,
+                                          const std::string &spelling_hint) const;
```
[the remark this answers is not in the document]
hmusta: The hint allows the caller to pass the spelling of a k-mer if it's already known (e.g., if you've traversed a certain number of steps already and stored the edge characters).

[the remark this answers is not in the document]
hmusta: good point, it's not needed

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -165,79 +175,54 @@ ::call_outgoing_kmers(node_index node, const OutgoingEdgeCallback &callback) con
...
-            callback(next, c);
-            assert(traverse(node, c) == next);
-            children[alphabet_encoder_[c]] = next;
-            --max_num_edges_left;
-        }
+        assert(has_sentinel_ || c != boss::BOSS::kSentinel);
+        assert(c == boss::BOSS::kSentinel || traverse(node, c) == next);
+        callback(next, c);
+        children[alphabet_encoder_[c]] = next;
+        --max_num_edges_left;
```
karasikov, line 182: In some cases you call dummy k-mers, in some cases you skip them... Can you add comments to specify this?

metagraph/src/graph/alignment/aligner_extender_methods.cpp
```diff
@@ -357,10 +358,31 @@ ::call_outgoing(node_index node,
...
+                    *it = c;
+                    cur_table_i = i_prev;
+                }
+                assert(hint == graph_->get_node_sequence(node));
+            }
+
+            canonical->call_outgoing_kmers(node, hint, [&](node_index next, char c) {
+                if (c != boss::BOSS::kSentinel)
+                    callback(next, c, 0);
+            });
```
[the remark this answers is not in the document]
hmusta: yes, an empty string is the default value for the argument

metagraph/src/graph/representation/canonical_dbg.cpp
hmusta: The idea here is to prevent a dummy sink/source k-mer from being called twice (one from each strand), since it could be that both orientations of a (k-1)-mer exist in the underlying graph.

metagraph/src/graph/alignment/aligner_extender_methods.cpp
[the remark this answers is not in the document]
hmusta: I added comments to `canonical_dbg.hpp` clarify this

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -309,14 +313,18 @@ void CanonicalDBG::call_incoming_kmers(node_index node,
...
+        }
     });
 
-    if (parents[alphabet_encoder_[boss::BOSS::kSentinel]])
-        callback(parents[alphabet_encoder_[boss::BOSS::kSentinel]], boss::BOSS::kSentinel);
+    if (has_sentinel_ && parents[alphabet_encoder_[boss::BOSS::kSentinel]]
+            && max_num_edges_left + has_sentinel_ == alphabet.size()) {
+        callback(parents[alphabet_encoder_[boss::BOSS::kSentinel]],
+                 boss::BOSS::kSentinel);
+    }
```
karasikov, line 327: ah, yeah. Makes sense. Because it might have a dummy sink on the forward strand but the reverse strand doesn't have any suffix, so no dummy source would be called.

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -146,275 +153,236 @@ void CanonicalDBG::map_to_nodes(std::string_view sequence,
...
-        is_palindrome_cache_.Put(prev, true);
+        if (c != boss::BOSS::kSentinel) {
+            callback(prev, c);
+            parents[s] = prev;
+            --max_num_edges_left;
+        }
     });
+
+    if (has_sentinel_ && parents[alphabet_encoder_[boss::BOSS::kSentinel]]
+            && max_num_edges_left + has_sentinel_ == alphabet.size()) {
```
karasikov, line 324:
```suggestion
            && max_num_edges_left + 1 == alphabet.size()) {
```

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -146,275 +153,236 @@ void CanonicalDBG::map_to_nodes(std::string_view sequence,
...
-        adjacent_outgoing_nodes(node, [&](node_index) {
-            if (found)
-                throw std::bad_function_call();
 
-            found = true;
-        });
-    } catch (const std::bad_function_call&) {
-        return true;
+    if (has_sentinel_ && children[alphabet_encoder_[boss::BOSS::kSentinel]]
+            && max_num_edges_left + has_sentinel_ == alphabet.size()) {
```
karasikov, line 235:
```suggestion
            && max_num_edges_left + 1 == alphabet.size()) {
```

PR #457, opened 2023-07-12
fix the flag for subsampling in clustering for multi-brwt

karasikov:


hmusta:
The issue with the CI run in the last commit was fixed by #458

PR #467, opened 2023-07-28
Add missing headers to `binary_matrix.hpp`

rob-p:
Add required `<istream>` and `<ostream>` headers.  

The stream types are referenced within the class and, at least under `g++11`, didn't seem to make their way into scope via another path within this file.  Adding them here allowed the code to compile successfully.


karasikov:
Thanks!

PR #481, opened 2024-01-26
Using SSHash submodule for graph representation

mmarzett:
Added the SSHash data structure as an option for the DBG representation, implemented using SSHash as a submodule.

### DIFF
--- a/metagraph/CMakeLists.txt
+++ b/metagraph/CMakeLists.txt
@@ -407,7 +407,7 @@ set_target_properties(metagraph PROPERTIES OUTPUT_NAME "metagraph_${CMAKE_DBG_AL
 
 set(METALIBS ${METALIBS}
   -lKMC
-  -lhts zlib
+  -lhts zlib deflate
   -lsdsl
   jsoncpp_lib
   progress_bar
--- a/metagraph/CMakeLists.txt
+++ b/metagraph/CMakeLists.txt
@@ -324,7 +324,6 @@ IF(APPLE)
 ENDIF()
 add_subdirectory(external-libraries/spdlog)
 add_subdirectory(external-libraries/sshash SYSTEM)
-target_compile_options(test_alphabet PRIVATE -Wno-strict-aliasing)
 add_subdirectory(external-libraries/DYNAMIC)
 add_subdirectory(external-libraries/zlib)
 target_compile_options(zlib
--- a/metagraph/src/cli/build.cpp
+++ b/metagraph/src/cli/build.cpp
@@ -250,7 +250,7 @@ int build_graph(Config *config) {
                                                              get_verbose()));
         }
 
-    } else if (config->graph_type == Config::GraphType::SSHASH){
+    } else if (config->graph_type == Config::GraphType::SSHASH) {
 
         graph.reset(new DBGSSHash(files.at(0), config->k, config->graph_mode));
         if(files.size() > 1){
--- a/metagraph/src/cli/build.cpp
+++ b/metagraph/src/cli/build.cpp
@@ -251,10 +251,9 @@ int build_graph(Config *config) {
         }
 
     } else if (config->graph_type == Config::GraphType::SSHASH) {
-
         graph.reset(new DBGSSHash(files.at(0), config->k, config->graph_mode));
-        if(files.size() > 1){
-            logger->error("Only one file for SSHash");
+        if (files.size() > 1) {
+            logger->error("DBGSSHash does not support multiple input files.");
             exit(1);
         }
 
--- a/metagraph/src/cli/build.cpp
+++ b/metagraph/src/cli/build.cpp
@@ -250,7 +250,7 @@ int build_graph(Config *config) {
                                                              get_verbose()));
         }
 
-    } else if (config->graph_type == Config::GraphType::SSHASH) {
+    } else if (config->graph_type == Config::GraphType::SSHASH && !config->dynamic) {
         graph.reset(new DBGSSHash(files.at(0), config->k, config->graph_mode, config->num_chars));
         if (files.size() > 1) {
             logger->error("DBGSSHash does not support multiple input files.");
--- a/metagraph/src/cli/build.cpp
+++ b/metagraph/src/cli/build.cpp
@@ -293,9 +293,10 @@ int build_graph(Config *config) {
                 exit(1);
 
             case Config::GraphType::SSHASH:
-            logger->error("SSHash-graph construction"
+                logger->error("SSHash-graph construction"
                               " in dynamic regime is not supported");
                 exit(1);
+
             case Config::GraphType::INVALID:
                 assert(false);
         }
--- a/metagraph/src/cli/build.cpp
+++ b/metagraph/src/cli/build.cpp
@@ -258,7 +258,7 @@ int build_graph(Config *config) {
             exit(1);
         }
 
-    }else {
+    } else {
         //slower method
         switch (config->graph_type) {
 
--- a/metagraph/src/cli/build.cpp
+++ b/metagraph/src/cli/build.cpp
@@ -251,7 +251,7 @@ int build_graph(Config *config) {
         }
 
     } else if (config->graph_type == Config::GraphType::SSHASH) {
-        graph.reset(new DBGSSHash(files.at(0), config->k, config->graph_mode));
+        graph.reset(new DBGSSHash(files.at(0), config->k, config->graph_mode, config->num_chars));
         if (files.size() > 1) {
             logger->error("DBGSSHash does not support multiple input files.");
             exit(1);
--- a/metagraph/src/cli/config/config.cpp
+++ b/metagraph/src/cli/config/config.cpp
@@ -243,6 +243,8 @@ Config::Config(int argc, char *argv[]) {
             alignment_post_chain = true;
         } else if (!strcmp(argv[i], "--align-no-seed-complexity-filter")) {
             alignment_seed_complexity_filter = false;
+        } else if (!strcmp(argv[i], "--num-chars")) {
+            num_chars = atoll(get_value(i++));
         } else if (!strcmp(argv[i], "--max-hull-depth")) {
             max_hull_depth = atoll(get_value(i++));
         } else if (!strcmp(argv[i], "--batch-align")) {
@@ -964,6 +966,7 @@ if (advanced) {
             fprintf(stderr, "\t   --count-kmers \tcount k-mers and build weighted graph [off]\n");
             fprintf(stderr, "\t   --count-width \tnumber of bits used to represent k-mer abundance [8]\n");
             fprintf(stderr, "\t   --index-ranges [INT]\tindex all node ranges in BOSS for suffixes of given length [%zu]\n", kDefaultIndexSuffixLen);
+            fprintf(stderr, "\t   --num-chars [INT]\tif the number of characters is known beforehand, enter it here [0]\n");
             fprintf(stderr, "\t-k --kmer-length [INT] \tlength of the k-mer to use [3]\n");
 #if ! _PROTEIN_GRAPH
             fprintf(stderr, "\t   --mode \t\tk-mer indexing mode: basic / canonical / primary [basic]\n");
--- a/metagraph/src/cli/config/config.cpp
+++ b/metagraph/src/cli/config/config.cpp
@@ -817,8 +817,7 @@ Config::GraphType Config::string_to_graphtype(const std::string &string) {
 
     } else if (string == "sshash") {
         return GraphType::SSHASH;
-    }
-     else {
+    } else {
         std::cerr << "Error: unknown graph representation" << std::endl;
         exit(1);
     }
--- a/metagraph/src/cli/config/config.cpp
+++ b/metagraph/src/cli/config/config.cpp
@@ -815,7 +815,7 @@ Config::GraphType Config::string_to_graphtype(const std::string &string) {
     } else if (string == "bitmap") {
         return GraphType::BITMAP;
 
-    } else if (string == "sshash"){
+    } else if (string == "sshash") {
         return GraphType::SSHASH;
     }
      else {
--- a/metagraph/src/graph/alignment/aligner_labeled.cpp
+++ b/metagraph/src/graph/alignment/aligner_labeled.cpp
@@ -11,11 +11,6 @@ namespace align {
 
 using mtg::common::logger;
 
-class early_term : public std::exception {
-  public:
-    const char* what() const noexcept override { return "Finished iteration"; }
-};
-
 typedef annot::matrix::BinaryMatrix::Row Row;
 typedef annot::matrix::BinaryMatrix::Column Column;
 typedef AnnotationBuffer::Columns Columns;
--- a/metagraph/src/graph/alignment/aligner_labeled.cpp
+++ b/metagraph/src/graph/alignment/aligner_labeled.cpp
@@ -11,6 +11,11 @@ namespace align {
 
 using mtg::common::logger;
 
+class early_term : public std::exception {
+  public:
+    const char* what() { return "Finished iteration"; }
+};
+
 typedef annot::matrix::BinaryMatrix::Row Row;
 typedef annot::matrix::BinaryMatrix::Column Column;
 typedef AnnotationBuffer::Columns Columns;
@@ -415,7 +420,7 @@ void LabeledExtender::call_alignments(score_t end_score,
                     }
 
                     if (label_it == label_end_it)
-                        throw std::bad_function_call();
+                        throw early_term();
 
                     if (c < *label_it)
                         return;
@@ -431,7 +436,7 @@ void LabeledExtender::call_alignments(score_t end_score,
                     }
                 }
             );
-        } catch (const std::bad_function_call&) {}
+        } catch (const early_term&) {}
     }
 
     if (alignment.label_coordinates.empty())
--- a/metagraph/src/graph/alignment/aligner_labeled.cpp
+++ b/metagraph/src/graph/alignment/aligner_labeled.cpp
@@ -13,7 +13,7 @@ using mtg::common::logger;
 
 class early_term : public std::exception {
   public:
-    const char* what() { return "Finished iteration"; }
+    const char* what() const noexcept override { return "Finished iteration"; }
 };
 
 typedef annot::matrix::BinaryMatrix::Row Row;
--- a/metagraph/src/graph/representation/base/sequence_graph.cpp
+++ b/metagraph/src/graph/representation/base/sequence_graph.cpp
@@ -83,14 +83,14 @@ bool DeBruijnGraph::find(std::string_view sequence,
 void DeBruijnGraph
 ::adjacent_outgoing_nodes(node_index node,
                           const std::function<void(node_index)> &callback) const {
-    assert(node > 0 && node <= num_nodes());
+    assert(node > 0 && node <= max_index());
     call_outgoing_kmers(node, [&](auto child, char) { callback(child); });
 }
 
 void DeBruijnGraph
 ::adjacent_incoming_nodes(node_index node,
                           const std::function<void(node_index)> &callback) const {
-    assert(node > 0 && node <= num_nodes());
+    assert(node > 0 && node <= max_index());
     call_incoming_kmers(node, [&](auto parent, char) { callback(parent); });
 }
 
--- a/metagraph/src/graph/representation/base/sequence_graph.cpp
+++ b/metagraph/src/graph/representation/base/sequence_graph.cpp
@@ -81,21 +81,18 @@ bool DeBruijnGraph::find(std::string_view sequence,
 }
 
 bool DeBruijnGraph::operator==(const DeBruijnGraph &other) const {
-    try {
-        other.call_kmers([&](node_index, const std::string &kmer) {
-            if (!find(kmer))
-                throw std::bad_function_call();
-        });
+    bool not_equal = false;
+    other.call_kmers([&](node_index, const std::string &kmer) {
+        if (!find(kmer))
+            not_equal = true;
+    }, [&]() { return not_equal; });
 
-        call_kmers([&](node_index, const std::string &kmer) {
-            if (!other.find(kmer))
-                throw std::bad_function_call();
-        });
-    } catch (const std::bad_function_call&) {
-        return false;
-    }
+    call_kmers([&](node_index, const std::string &kmer) {
+        if (!other.find(kmer))
+            not_equal = true;
+    }, [&]() { return not_equal; });
 
-    return true;
+    return !not_equal;
 }
 
 void DeBruijnGraph::traverse(node_index start,
@@ -413,7 +410,12 @@ void DeBruijnGraph::call_unitigs(const CallPath &callback,
  * Traverse graph and iterate over all nodes
  */
 void DeBruijnGraph
-::call_kmers(const std::function<void(node_index, const std::string&)> &callback) const {
+::call_kmers(const std::function<void(node_index,
+             const std::string&)> &callback,
+             const std::function<bool()> &stop_early) const {
+    if (stop_early())
+        return;
+
     sdsl::bit_vector visited(max_index() + 1, false);
     std::stack<std::pair<node_index, std::string>> nodes;
 
@@ -457,7 +459,7 @@ void DeBruijnGraph
                 sequence.push_back(next_c);
             }
         }
-    });
+    }, stop_early);
 }
 
 void DeBruijnGraph::print(std::ostream &out) const {
--- a/metagraph/src/graph/representation/base/sequence_graph.cpp
+++ b/metagraph/src/graph/representation/base/sequence_graph.cpp
@@ -80,9 +80,22 @@ bool DeBruijnGraph::find(std::string_view sequence,
     return num_kmers_missing <= max_kmers_missing;
 }
 
-bool DeBruijnGraph::operator==(const DeBruijnGraph &) const {
-    throw std::runtime_error("Not implemented");
-    return false;
+bool DeBruijnGraph::operator==(const DeBruijnGraph &other) const {
+    try {
+        other.call_kmers([&](node_index, const std::string &kmer) {
+            if (!find(kmer))
+                throw std::bad_function_call();
+        });
+
+        call_kmers([&](node_index, const std::string &kmer) {
+            if (!other.find(kmer))
+                throw std::bad_function_call();
+        });
+    } catch (const std::bad_function_call&) {
+        return false;
+    }
+
+    return true;
 }
 
 void DeBruijnGraph::traverse(node_index start,
--- a/metagraph/src/graph/representation/base/sequence_graph.cpp
+++ b/metagraph/src/graph/representation/base/sequence_graph.cpp
@@ -80,6 +80,20 @@ bool DeBruijnGraph::find(std::string_view sequence,
     return num_kmers_missing <= max_kmers_missing;
 }
 
+void DeBruijnGraph
+::adjacent_outgoing_nodes(node_index node,
+                          const std::function<void(node_index)> &callback) const {
+    assert(node > 0 && node <= num_nodes());
+    call_outgoing_kmers(node, [&](auto child, char) { callback(child); });
+}
+
+void DeBruijnGraph
+::adjacent_incoming_nodes(node_index node,
+                          const std::function<void(node_index)> &callback) const {
+    assert(node > 0 && node <= num_nodes());
+    call_incoming_kmers(node, [&](auto parent, char) { callback(parent); });
+}
+
 bool DeBruijnGraph::operator==(const DeBruijnGraph &other) const {
     bool not_equal = false;
     other.call_kmers([&](node_index, const std::string &kmer) {
--- a/metagraph/src/graph/representation/canonical_dbg.cpp
+++ b/metagraph/src/graph/representation/canonical_dbg.cpp
@@ -12,11 +12,6 @@ namespace graph {
 using mtg::common::logger;
 using edge_index = boss::BOSS::edge_index;
 
-class early_term : public std::exception {
-  public:
-    const char* what() const noexcept override { return "Finished iteration"; }
-};
-
 inline const DBGSuccinct* get_dbg_succ(const DeBruijnGraph &graph) {
     const DeBruijnGraph *base_graph = &graph;
     while (const auto *wrapper = dynamic_cast<const DBGWrapper<>*>(base_graph)) {
--- a/metagraph/src/graph/representation/canonical_dbg.cpp
+++ b/metagraph/src/graph/representation/canonical_dbg.cpp
@@ -12,6 +12,11 @@ namespace graph {
 using mtg::common::logger;
 using edge_index = boss::BOSS::edge_index;
 
+class early_term : public std::exception {
+  public:
+    const char* what() { return "Finished iteration"; }
+};
+
 inline const DBGSuccinct* get_dbg_succ(const DeBruijnGraph &graph) {
     const DeBruijnGraph *base_graph = &graph;
     while (const auto *wrapper = dynamic_cast<const DBGWrapper<>*>(base_graph)) {
@@ -386,11 +391,11 @@ bool CanonicalDBG::has_multiple_outgoing(node_index node) const {
     try {
         adjacent_outgoing_nodes(node, [&](node_index) {
             if (found)
-                throw std::bad_function_call();
+                throw early_term();
 
             found = true;
         });
-    } catch (const std::bad_function_call&) {
+    } catch (const early_term&) {
         return true;
     }
 
@@ -402,9 +407,9 @@ bool CanonicalDBG::has_single_incoming(node_index node) const {
     try {
         adjacent_incoming_nodes(node, [&](node_index) {
             if (++count > 1)
-                throw std::bad_function_call();
+                throw early_term();
         });
-    } catch (const std::bad_function_call&) {}
+    } catch (const early_term&) {}
 
     return count == 1;
 }
@@ -506,16 +511,20 @@ void CanonicalDBG::call_nodes(const std::function<void(node_index)> &callback,
 }
 
 void CanonicalDBG
-::call_kmers(const std::function<void(node_index, const std::string&)> &callback) const {
+::call_kmers(const std::function<void(node_index, const std::string&)> &callback,
+             const std::function<bool()> &stop_early) const {
     graph_->call_kmers([&](node_index i, const std::string &seq) {
         callback(i, seq);
+        if (stop_early())
+            return;
+
         node_index j = reverse_complement(i);
         if (j != i) {
             std::string rseq(seq);
             ::reverse_complement(rseq.begin(), rseq.end());
             callback(j, rseq);
         }
-    });
+    }, stop_early);
 }
 
 bool CanonicalDBG::operator==(const DeBruijnGraph &other) const {
--- a/metagraph/src/graph/representation/canonical_dbg.cpp
+++ b/metagraph/src/graph/representation/canonical_dbg.cpp
@@ -64,7 +64,7 @@ void CanonicalDBG
     path.reserve(sequence.size() - get_k() + 1);
 
     if (const auto sshash = std::dynamic_pointer_cast<const DBGSSHash>(graph_)) {
-        sshash->map_to_nodes_with_rc(sequence, [&](node_index node, bool orientation) {
+        sshash->map_to_nodes_with_rc<>(sequence, [&](node_index node, bool orientation) {
             callback(node && orientation ? reverse_complement(node) : node);
         }, terminate);
         return;
@@ -180,7 +180,7 @@ void CanonicalDBG::call_outgoing_kmers(node_index node,
     }
 
     if (const auto sshash = std::dynamic_pointer_cast<const DBGSSHash>(graph_)) {
-        sshash->call_outgoing_kmers_with_rc(node, [&](node_index next, char c, bool orientation) {
+        sshash->call_outgoing_kmers_with_rc<>(node, [&](node_index next, char c, bool orientation) {
             callback(orientation ? reverse_complement(next) : next, c);
         });
         return;
@@ -273,7 +273,7 @@ void CanonicalDBG::call_incoming_kmers(node_index node,
     }
 
     if (const auto sshash = std::dynamic_pointer_cast<const DBGSSHash>(graph_)) {
-        sshash->call_incoming_kmers_with_rc(node, [&](node_index prev, char c, bool orientation) {
+        sshash->call_incoming_kmers_with_rc<>(node, [&](node_index prev, char c, bool orientation) {
             callback(orientation ? reverse_complement(prev) : prev, c);
         });
         return;
--- a/metagraph/src/graph/representation/canonical_dbg.cpp
+++ b/metagraph/src/graph/representation/canonical_dbg.cpp
@@ -14,7 +14,7 @@ using edge_index = boss::BOSS::edge_index;
 
 class early_term : public std::exception {
   public:
-    const char* what() { return "Finished iteration"; }
+    const char* what() const noexcept override { return "Finished iteration"; }
 };
 
 inline const DBGSuccinct* get_dbg_succ(const DeBruijnGraph &graph) {
--- a/metagraph/src/graph/representation/hash/dbg_sshash.cpp
+++ b/metagraph/src/graph/representation/hash/dbg_sshash.cpp
@@ -79,13 +79,13 @@ DBGSSHash::DBGSSHash(const std::string &input_filename, size_t k, Mode mode, siz
     build_config.num_threads = get_num_threads();
 
     // silence sshash construction messages when not verbose
-    std::ostringstream devnull;
-    auto cout_buff = std::cout.rdbuf();
+    std::ios orig_state(nullptr);
+    orig_state.copyfmt(std::cout);
     if (!common::get_verbose())
-        std::cout.rdbuf(devnull.rdbuf());
+        std::cout.setstate(std::ios_base::failbit);
     std::visit([&](auto &d) { d.build(input_filename, build_config); }, dict_);
     if (!common::get_verbose())
-        std::cout.rdbuf(cout_buff);
+        std::cout.copyfmt(orig_state);
 
     num_nodes_ = dict_size();
 }
--- a/metagraph/src/graph/representation/hash/dbg_sshash.cpp
+++ b/metagraph/src/graph/representation/hash/dbg_sshash.cpp
@@ -122,7 +122,7 @@ std::string DBGSSHash::file_extension() const {
 
 void DBGSSHash::add_sequence(std::string_view sequence,
                              const std::function<void(node_index)>& on_insertion) {
-    throw std::runtime_error("adding sequences not implemented");
+    throw std::logic_error("adding sequences not supported");
 }
 
 void DBGSSHash::map_to_nodes(std::string_view sequence,
--- a/metagraph/src/graph/representation/hash/dbg_sshash.cpp
+++ b/metagraph/src/graph/representation/hash/dbg_sshash.cpp
@@ -188,7 +188,7 @@ DBGSSHash::node_index DBGSSHash::traverse(node_index node, char next_char) const
 DBGSSHash::node_index DBGSSHash::traverse_back(node_index node, char prev_char) const {
     assert(node > 0 && node <= num_nodes());
     // TODO: if a node is in the middle of a unitig, then we only need to check the previous node index
-    std::string string_kmer = std::string(1, prev_char) + get_node_sequence(node);
+    std::string string_kmer = prev_char + get_node_sequence(node);
     string_kmer.pop_back();
     return kmer_to_node(string_kmer);
 }
--- a/metagraph/src/graph/representation/hash/dbg_sshash.cpp
+++ b/metagraph/src/graph/representation/hash/dbg_sshash.cpp
@@ -101,11 +101,9 @@ DBGSSHash::DBGSSHash(std::string const& input_filename, size_t k, Mode mode)
     sshash::build_configuration build_config;
     build_config.k = k;
     // quick fix for value of m... k/2 but odd
-    build_config.m = std::min(uint64_t((k_ + 1) / 2), sshash::constants::max_m);
+    build_config.m = std::min(uint64_t(k / 2) | 1, sshash::constants::max_m);
     build_config.verbose = common::get_verbose();
     build_config.num_threads = get_num_threads();
-    if (build_config.m % 2 == 0)
-        build_config.m++;
 
     // silence sshash construction messages when not verbose
     if (!common::get_verbose())
--- a/metagraph/src/graph/representation/hash/dbg_sshash.cpp
+++ b/metagraph/src/graph/representation/hash/dbg_sshash.cpp
@@ -107,11 +107,7 @@ void DBGSSHash::map_to_nodes_sequentially(std::string_view sequence,
     for (size_t i = k_ - 1; i < sequence.size() && !terminate(); ++i) {
         uint_kmer.drop_char();
         uint_kmer.kth_char_or(k_ - 1, kmer_t::char_to_uint(sequence[i]));
-#if _PROTEIN_GRAPH
-        callback(sshash_to_graph_index(dict_.lookup_uint(uint_kmer)));
-#else
         callback(sshash_to_graph_index(dict_.lookup_uint(uint_kmer, false)));
-#endif
     }
 }
 
@@ -133,12 +129,8 @@ void DBGSSHash::map_to_nodes_with_rc(std::string_view sequence,
     for (size_t i = k_ - 1; i < sequence.size() && !terminate(); ++i) {
         uint_kmer.drop_char();
         uint_kmer.kth_char_or(k_ - 1, kmer_t::char_to_uint(sequence[i]));
-#if _PROTEIN_GRAPH
-        callback(sshash_to_graph_index(dict_.lookup_uint(uint_kmer)), false);
-#else
         auto res = dict_.lookup_advanced_uint(uint_kmer, true);
         callback(sshash_to_graph_index(res.kmer_id), res.kmer_orientation);
-#endif
     }
 }
 
@@ -147,13 +139,9 @@ DBGSSHash::node_index DBGSSHash::traverse(node_index node, char next_char) const
     kmer_t new_kmer = sshash::util::string_to_uint_kmer<kmer_t>(string_kmer.c_str(), k_);
     new_kmer.drop_char();
     new_kmer.kth_char_or(k_ - 1, kmer_t::char_to_uint(next_char));
-#if _PROTEIN_GRAPH
-    return sshash_to_graph_index(dict_.lookup_uint(new_kmer));
-#else
     auto res = dict_.lookup_advanced_uint(new_kmer, mode_ == CANONICAL);
     node_index next = sshash_to_graph_index(res.kmer_id);
     return res.kmer_orientation ? reverse_complement(next) : next;
-#endif
 }
 
 DBGSSHash::node_index DBGSSHash::traverse_back(node_index node, char prev_char) const {
@@ -161,13 +149,9 @@ DBGSSHash::node_index DBGSSHash::traverse_back(node_index node, char prev_char) 
     kmer_t new_kmer = sshash::util::string_to_uint_kmer<kmer_t>(string_kmer.c_str(), k_);
     new_kmer.append_char(kmer_t::char_to_uint(prev_char));
     new_kmer.take_chars(k_);
-#if _PROTEIN_GRAPH
-    return sshash_to_graph_index(dict_.lookup_uint(new_kmer));
-#else
     auto res = dict_.lookup_advanced_uint(new_kmer, mode_ == CANONICAL);
     node_index prev = sshash_to_graph_index(res.kmer_id);
     return res.kmer_orientation ? reverse_complement(prev) : prev;
-#endif
 }
 
 void DBGSSHash::adjacent_outgoing_nodes(node_index node,
@@ -269,13 +253,9 @@ DBGSSHash::node_index DBGSSHash::kmer_to_node(std::string_view kmer) const {
     if (!num_nodes())
         return npos;
 
-#if _PROTEIN_GRAPH
-    return sshash_to_graph_index(dict_.lookup(kmer.data()));
-#else
     auto res = dict_.lookup_advanced(kmer.data(), mode_ == CANONICAL);
     node_index node = sshash_to_graph_index(res.kmer_id);
     return res.kmer_orientation ? reverse_complement(node) : node;
-#endif
 }
 
 std::pair<DBGSSHash::node_index, bool>
@@ -283,12 +263,8 @@ DBGSSHash::kmer_to_node_with_rc(std::string_view kmer) const {
     if (!num_nodes())
         return std::make_pair(npos, false);
 
-#if _PROTEIN_GRAPH
-    return std::make_pair(sshash_to_graph_index(dict_.lookup(kmer.data())), false);
-#else
     auto res = dict_.lookup_advanced(kmer.data(), true);
     return std::make_pair(sshash_to_graph_index(res.kmer_id), res.kmer_orientation);
-#endif
 }
 
 std::string DBGSSHash::get_node_sequence(node_index node) const {
--- a/metagraph/src/graph/representation/hash/dbg_sshash.cpp
+++ b/metagraph/src/graph/representation/hash/dbg_sshash.cpp
@@ -355,33 +355,90 @@ std::string DBGSSHash::get_node_sequence(node_index node) const {
     return str_kmer;
 }
 
+
+struct generic_loader {
+    generic_loader(std::istream &is) : m_is(is) {
+        if (!m_is.good())
+            throw std::runtime_error("Error in loading stream.");
+    }
+
+    template <typename T>
+    void visit(T& val) {
+        if constexpr (essentials::is_pod<T>::value) {
+            essentials::load_pod(m_is, val);
+        } else {
+            val.visit(*this);
+        }
+    }
+
+    template <typename T, typename Allocator>
+    void visit(std::vector<T, Allocator>& vec) {
+        size_t n;
+        visit(n);
+        vec.resize(n);
+        if constexpr (essentials::is_pod<T>::value) {
+            m_is.read(reinterpret_cast<char*>(vec.data()),
+                      static_cast<std::streamsize>(sizeof(T) * n));
+        } else {
+            for (auto& v : vec) visit(v);
+        }
+    }
+
+private:
+    std::istream &m_is;
+};
+
+struct generic_saver {
+    generic_saver(std::ostream &os) : m_os(os) {
+        if (!m_os.good())
+            throw std::runtime_error("Error in saving stream.");
+    }
+
+    template <typename T>
+    void visit(const T &val) {
+        if constexpr (essentials::is_pod<T>::value) {
+            essentials::save_pod(m_os, val);
+        } else {
+            // TODO: the visit methods in sshash are non-const
+            const_cast<T&>(val).visit(*this);
+        }
+    }
+
+    template <typename T, typename Allocator>
+    void visit(const std::vector<T, Allocator> &vec) {
+        if constexpr (essentials::is_pod<T>::value) {
+            essentials::save_vec(m_os, vec);
+        } else {
+            size_t n = vec.size();
+            visit(n);
+            for (const auto &v : vec) visit(v);
+        }
+    }
+
+private:
+    std::ostream &m_os;
+};
+
+
 void DBGSSHash::serialize(std::ostream& out) const {
-    // TODO
-    throw std::runtime_error("serialize to stream not implemented");
+    generic_saver saver(out);
+
+    saver.visit(num_nodes_);
+    saver.visit(k_);
+    saver.visit(mode_);
+
+    if (num_nodes())
+        std::visit([&](const auto &d) { saver.visit(d); }, dict_);
 }
 
 void DBGSSHash::serialize(const std::string& filename) const {
     std::string suffixed_filename = utils::make_suffix(filename, kExtension);
-    essentials::saver saver(suffixed_filename.c_str());
-
-    // TODO: fix this in the essentials library. for some reason, its saver takes a non-const ref
-    saver.visit(const_cast<size_t&>(num_nodes_));
-    saver.visit(const_cast<size_t&>(k_));
-    saver.visit(const_cast<Mode&>(mode_));
-
-    if (num_nodes())
-        std::visit([&](auto &d) { d.visit(saver); }, const_cast<dict_t&>(dict_));
+    std::ofstream fout(suffixed_filename, std::ios::binary);
+    serialize(fout);
 }
 
-bool DBGSSHash::load(std::istream& in) {
-    throw std::runtime_error("load from stream not implemented");
-    return false;
-}
-
-bool DBGSSHash::load(const std::string& filename) {
-    std::string suffixed_filename = utils::make_suffix(filename, kExtension);
-    essentials::loader loader(suffixed_filename.c_str());
-
+bool DBGSSHash::load(std::istream &in) {
+    generic_loader loader(in);
     size_t num_nodes;
     size_t k;
     Mode mode;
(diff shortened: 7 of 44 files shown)


adamant-pwn:
```

[  FAILED  ] 10 tests, listed below:
[  FAILED  ] DeBruijnGraphTest/4.CallStartNodes, where TypeParam = mtg::graph::DBGSSHash
[  FAILED  ] DeBruijnGraphTest/4.CallUnitigsWithoutTips, where TypeParam = mtg::graph::DBGSSHash
[  FAILED  ] DeBruijnGraphTest/4.CallUnitigsWithoutTips2, where TypeParam = mtg::graph::DBGSSHash
[  FAILED  ] DeBruijnGraphTest/4.CallUnitigsCheckDegree, where TypeParam = mtg::graph::DBGSSHash
[  FAILED  ] DeBruijnGraphTest/4.CallUnitigsCross, where TypeParam = mtg::graph::DBGSSHash
[  FAILED  ] DeBruijnGraphTest/4.get_maximum_indegree, where TypeParam = mtg::graph::DBGSSHash
[  FAILED  ] DeBruijnGraphTest/4.get_degree2, where TypeParam = mtg::graph::DBGSSHash
[  FAILED  ] DeBruijnGraphTest/4.indegree_identity_incoming_indegree, where TypeParam = mtg::graph::DBGSSHash
[  FAILED  ] DeBruijnGraphTest/4.indegree_identity_traverse_back_incoming, where TypeParam = mtg::graph::DBGSSHash
[  FAILED  ] DeBruijnGraphTest/4.CallOutgoingEdges, where TypeParam = mtg::graph::DBGSSHash

10 FAILED TESTS
```

Are these the same tests that were disabled, or is it something new?

adamant-pwn (changes requested):

metagraph/tests/data/sshash_sequences/pthash.tmp.run1700659576911610385.free_slots.bin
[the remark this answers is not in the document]
hmusta: correct, these shouldn't be here, so let's add the `sshash_sequences` directory to the `.gitignore`

metagraph/src/graph/representation/hash/dbg_sshash.hpp
```diff
@@ -0,0 +1,98 @@
...
+#include <tsl/ordered_set.h>
+#include "common/utils/string_utils.hpp"
+#include "graph/representation/base/sequence_graph.hpp"
+#include "../external-libraries/sshash/external/pthash/external/essentials/include/essentials.hpp"
+
+namespace sshash{
+class dictionary;
+}
+namespace mtg {
+namespace graph {
```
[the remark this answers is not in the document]
hmusta: we can do `namespace mtg::graph`

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+void DBGSSHash::serialize(std::ostream &out) const {
+    //TODO
+    throw std::runtime_error("serialize to stream not implemented");
+}
+
+void DBGSSHash::serialize(const std::string &filename) const {
+    std::string suffixed_filename = utils::make_suffix(filename, kExtension);
+    
+    essentials::logger("saving data structure to disk...");
```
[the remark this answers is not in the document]
hmusta:
```suggestion
    common::logger->trace("saving data structure to disk...");
```

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+bool DBGSSHash::load(std::istream &in) {
+    throw std::runtime_error("load from stream not implemented");
+    return false;
+}
+
+bool DBGSSHash::load(const std::string &filename) {
+    std::string suffixed_filename = utils::make_suffix(filename, kExtension);
+    uint64_t num_bytes_read = essentials::load(*dict_, suffixed_filename.c_str());
+    bool verbose = true; // temp
```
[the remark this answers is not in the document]
hmusta: I'm guessing that the intention was to print out some debugging information that's only accessible with verbose runs from the CLI (via the `-v` flag)

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+void DBGSSHash::call_kmers(
+        const std::function<void(node_index, const std::string &)> &callback) const {
+    for (size_t node_idx = 1; node_idx <= num_nodes(); ++node_idx) {
+        callback(node_idx, get_node_sequence(node_idx));
+    }
+}
+
+DBGSSHash::node_index DBGSSHash::kmer_to_node(std::string_view kmer) const {
+    uint64_t ssh_idx = dict_->lookup(kmer.begin(), false);
```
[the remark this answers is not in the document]
hmusta: The ideal thing (to avoid the copy) would be to pass the end iterator as well

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+bool DBGSSHash::has_no_incoming(node_index node) const {
+    return DBGSSHash::indegree(node) == 0;
+}
+
+bool DBGSSHash::has_single_incoming(node_index node) const {
+    return DBGSSHash::indegree(node) == 1;
+}
+
+void DBGSSHash::call_kmers(
```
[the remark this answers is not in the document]
hmusta: The default implementation in `sequence_graph.cpp` should be good enough. This method is almost never used.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,234 @@
...
+                    + bool(nb.backward_G.kmer_id + 1)
+                    + bool(nb.backward_T.kmer_id + 1);
+    return in_deg;
+}
+
+bool DBGSSHash::has_no_incoming(node_index node) const {
+    return DBGSSHash::indegree(node) == 0;
+}
+
+bool DBGSSHash::has_single_incoming(node_index node) const {
```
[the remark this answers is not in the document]
hmusta: The default implementation involves computing the indegree and checking if it's equal to 1. I feel like this can be done more efficiently for DBGSSHash

metagraph/src/cli/build.cpp
```diff
@@ -249,7 +250,15 @@ int build_graph(Config *config) {
                                                              get_verbose()));
         }
 
-    } else {
+    } else if (config->graph_type == Config::GraphType::SSHASH){
```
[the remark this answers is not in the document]
hmusta: AFAIK SSHash can't be made dynamic anyway, so I think we can keep this as-is, but then do a nested check for `!config->dynamic` and terminate if that check fails

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+DBGSSHash::node_index DBGSSHash::kmer_to_node(std::string_view kmer) const {
+    uint64_t ssh_idx = dict_->lookup(kmer.begin(), false);
+    return ssh_idx + 1;
+}
+
+std::string DBGSSHash::get_node_sequence(node_index node) const {
+    std::string str_kmer = "";
+    str_kmer.append(k_, ' ');
+    uint64_t ssh_idx = node - 1; // switch back to sshash idx!!!
```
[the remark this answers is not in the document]
hmusta: we can add static constexpr methods called `DBGSSHash::graph_to_hash_index` and `DBGSSHash::hash_to_graph_index` to handle this

hmusta:
> ```
> 
> [  FAILED  ] 10 tests, listed below:
> [  FAILED  ] DeBruijnGraphTest/4.CallStartNodes, where TypeParam = mtg::graph::DBGSSHash
> [  FAILED  ] DeBruijnGraphTest/4.CallUnitigsWithoutTips, where TypeParam = mtg::graph::DBGSSHash
> [  FAILED  ] DeBruijnGraphTest/4.CallUnitigsWithoutTips2, where TypeParam = mtg::graph::DBGSSHash
> [  FAILED  ] DeBruijnGraphTest/4.CallUnitigsCheckDegree, where TypeParam = mtg::graph::DBGSSHash
> [  FAILED  ] DeBruijnGraphTest/4.CallUnitigsCross, where TypeParam = mtg::graph::DBGSSHash
> [  FAILED  ] DeBruijnGraphTest/4.get_maximum_indegree, where TypeParam = mtg::graph::DBGSSHash
> [  FAILED  ] DeBruijnGraphTest/4.get_degree2, where TypeParam = mtg::graph::DBGSSHash
> [  FAILED  ] DeBruijnGraphTest/4.indegree_identity_incoming_indegree, where TypeParam = mtg::graph::DBGSSHash
> [  FAILED  ] DeBruijnGraphTest/4.indegree_identity_traverse_back_incoming, where TypeParam = mtg::graph::DBGSSHash
> [  FAILED  ] DeBruijnGraphTest/4.CallOutgoingEdges, where TypeParam = mtg::graph::DBGSSHash
> 
> 10 FAILED TESTS
> ```
> 
> Are these the same tests that were disabled, or is it something new?

`DeBruijnGraphTest/4.CallStartNodes` seems to SegFault because of a multi-threading issue. This goes away if I change its hard-coding from 8 threads to 1

metagraph/tests/graph/all/test_dbg_helpers.cpp
```diff
@@ -126,6 +131,47 @@ build_graph<DBGBitmap>(uint64_t k,
...
+    
+    // use DBGHashString to get contigs for SSHash 
+    auto string_graph = build_graph<DBGHashString>(k, sequences,
+                           mode);
+
+    std::vector<std::string> contigs;
+    string_graph->call_sequences([&](const std::string &contig, const auto &) {
+        contigs.push_back(contig);
+    }, 1, false);
+    std::string dump_path = "../tests/data/sshash_sequences/contigs.fa";
```
hmusta, line 168: please add this directory to the `.gitignore`

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+}
+
+void DBGSSHash::map_to_nodes(std::string_view sequence,
+                             const std::function<void(node_index)> &callback,
+                             const std::function<bool()> &terminate) const {
+    map_to_nodes_sequentially(sequence, callback, terminate);
+}
+
+void DBGSSHash ::map_to_nodes_sequentially(std::string_view sequence,
```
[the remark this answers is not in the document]
hmusta: Depending on the graph type, it may be much faster to get the successor of a known k-mer than to look for each k-mer independently.

For SSHash, this is the case. Rather than encoding the k-1 redundant characters between subsequent k-mers, we can encode the first one and keep updating the hash as we move along the sequence.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,234 @@
+#include "dbg_sshash.hpp"
+#include <dictionary.hpp>
```
hmusta, line 2: please update to the latest commit of sshash. I fixed a graph traversal bug that was affecting how we traverse our graphs.

hmusta (changes requested):

metagraph/src/graph/representation/hash/dbg_sshash.cpp
hmusta: This has been fixed

metagraph/src/graph/representation/hash/dbg_sshash.hpp
```diff
@@ -0,0 +1,98 @@
+#ifndef __DBG_SSHASH_HPP__
+#define __DBG_SSHASH_HPP__
+
+#include <iostream>
+#include <tsl/ordered_set.h>
+#include "common/utils/string_utils.hpp"
+#include "graph/representation/base/sequence_graph.hpp"
+#include "../external-libraries/sshash/external/pthash/external/essentials/include/essentials.hpp"
```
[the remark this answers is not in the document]
hmusta: This has been fixed

metagraph/src/graph/representation/hash/dbg_sshash.cpp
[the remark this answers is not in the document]
hmusta: I've disabled this

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -58,7 +58,7 @@ void DBGSSHash ::map_to_nodes_sequentially(std::string_view sequence,
     if (terminate() || sequence.size() < k_)
         return;
 
-    kmer_t uint_kmer = sshash::util::string_to_uint_kmer<kmer_t>(sequence.data(), k_ - 1);
+    kmer_t uint_kmer = sshash::util::string_to_uint_kmer<kmer_t>(std::string(sequence).c_str(), k_ - 1);
```
hmusta, line 61: since we know that the sequence size is at least `k_` at this point, and since `k_` is known by `sshash`, can we switch to using `sequence.c_str()` to avoid the copy?

hmusta (changes requested):

metagraph/src/graph/representation/hash/dbg_sshash.cpp
adamant-pwn: Hmm. I wrote it like this because `string_view` is not guaranteed to be null-terminated. But you're right, we don't actually rely on it in sshash, so I suppose it's fine to keep using `.data()`, as long as we're aware of potential consequences if sshash eventually starts relying on it being null-terminated (unlikely). Will update shortly.

metagraph/CMakeLists.txt
```diff
@@ -319,6 +323,8 @@ IF(APPLE)
     set(THREADS_PREFER_PTHREAD_FLAG ON)
 ENDIF()
 add_subdirectory(external-libraries/spdlog)
+add_subdirectory(external-libraries/sshash SYSTEM)
+target_compile_options(test_alphabet PRIVATE -Wno-strict-aliasing)
```
adamant-pwn, line 327: Is `-Wno-strict-aliasing` still needed here? I think it should've been fixed by https://github.com/ratschlab/sshash/pull/3/commits/889fe5c1cbef86f877d2413714ea5e899c26f328.

metagraph/src/cli/build.cpp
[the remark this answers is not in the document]
adamant-pwn: Not sure I understand. Why keep it different from [this](https://github.com/ratschlab/metagraph/pull/481/files#diff-c13de607dbf5412cdb54c428f602b85159590cb80c7925142790cdb75209e308R185) and [this](https://github.com/ratschlab/metagraph/pull/481/files#diff-c13de607dbf5412cdb54c428f602b85159590cb80c7925142790cdb75209e308R100) lines?

More importantly, if we don't put `&& !config->dynamic` here, we will not get into [this case](https://github.com/ratschlab/metagraph/pull/481/files#diff-c13de607dbf5412cdb54c428f602b85159590cb80c7925142790cdb75209e308R296-R299) below, which would make it unreachable code, and we most likely don't want that.

metagraph/src/cli/build.cpp
```diff
@@ -249,7 +250,15 @@ int build_graph(Config *config) {
...
+
+        graph.reset(new DBGSSHash(files.at(0), config->k, config->graph_mode));
+        if(files.size() > 1){
+            logger->error("Only one file for SSHash");
+            exit(1);
+        }
+
+    }else {
```
adamant-pwn, line 261: Missing whitespace.

metagraph/src/cli/build.cpp
adamant-pwn, line 253: Missing whitespace between `){`.

metagraph/src/cli/build.cpp
```diff
@@ -249,7 +250,15 @@ int build_graph(Config *config) {
                                                              get_verbose()));
         }
 
-    } else {
+    } else if (config->graph_type == Config::GraphType::SSHASH){
+
+        graph.reset(new DBGSSHash(files.at(0), config->k, config->graph_mode));
+        if(files.size() > 1){
```
adamant-pwn, line 256: Missing whitespaces between `if(` and `){`.

metagraph/src/cli/build.cpp
```diff
@@ -249,7 +250,15 @@ int build_graph(Config *config) {
                                                              get_verbose()));
         }
 
-    } else {
+    } else if (config->graph_type == Config::GraphType::SSHASH){
+
+        graph.reset(new DBGSSHash(files.at(0), config->k, config->graph_mode));
```
adamant-pwn, line 255: nit: `files[0]`? Also `graph = std::make_unique<DBGSSHash>(...)` seems to be the [preferred approach](https://stackoverflow.com/a/41247901/4628730), but maybe not that important, given that we use `graph.reset(new ...)` everywhere throughout the file.

metagraph/src/cli/build.cpp
```diff
@@ -249,7 +250,15 @@ int build_graph(Config *config) {
...
+        if(files.size() > 1){
+            logger->error("Only one file for SSHash");
+            exit(1);
+        }
+
```
adamant-pwn, line 260: Redundant newline?

metagraph/src/cli/build.cpp
```diff
@@ -249,7 +250,15 @@ int build_graph(Config *config) {
                                                              get_verbose()));
         }
 
-    } else {
+    } else if (config->graph_type == Config::GraphType::SSHASH){
+
+        graph.reset(new DBGSSHash(files.at(0), config->k, config->graph_mode));
+        if(files.size() > 1){
+            logger->error("Only one file for SSHash");
```
adamant-pwn, line 257: nit: "SSHash-graph does not support multiple input files"?

metagraph/src/cli/build.cpp
```diff
@@ -284,6 +293,10 @@ int build_graph(Config *config) {
                               " in dynamic regime is not supported");
                 exit(1);
 
+            case Config::GraphType::SSHASH:
+            logger->error("SSHash-graph construction"
+                              " in dynamic regime is not supported");
```
adamant-pwn, line 298: Misaligned indentation.

metagraph/src/cli/config/config.cpp
```diff
@@ -815,7 +815,10 @@ Config::GraphType Config::string_to_graphtype(const std::string &string) {
     } else if (string == "bitmap") {
         return GraphType::BITMAP;
 
-    } else {
+    } else if (string == "sshash"){
```
adamant-pwn, line 818: Missing whitespace in `){`.

metagraph/src/cli/config/config.cpp
```diff
@@ -815,7 +815,10 @@ Config::GraphType Config::string_to_graphtype(const std::string &string) {
...
+        return GraphType::SSHASH;
+    }
+     else {
```
adamant-pwn, line 821: Extra newline.

metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -62,6 +63,13 @@ ::map_to_nodes_sequentially(std::string_view sequence,
     std::vector<node_index> path;
     path.reserve(sequence.size() - get_k() + 1);
 
+    if (const auto sshash = std::dynamic_pointer_cast<const DBGSSHash>(graph_)) {
```
adamant-pwn, line 66: Why do we need to do it here, and why only for DBGSSHash?

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,494 @@
+#include "dbg_sshash.hpp"
+
+#include <sdsl/uint256_t.hpp>
```
adamant-pwn, line 3: Do we have some tests that may be used as speed benchmarks for using it?

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,494 @@
...
+
+
+namespace mtg {
+namespace graph {
+
+const std::string DBGSSHash::alphabet_ = kmer::KmerExtractor2Bit().alphabet;
+
+using Kmer64 = uint64_t;
+using Kmer128 = __uint128_t;
+using Kmer256 = sdsl::uint256_t;
```
adamant-pwn, line 18: Do we have some tests that may be used as speed benchmarks for using it?

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,494 @@
...
+
+template <typename KmerInt>
+class Dictionary : public IDictionary {
```
adamant-pwn, line 21: Why do we need this convoluted strcture of Dictionary and IDictionary, rather than using `sshash::dictionary` directly, or at least inhereting from it directly? If it is about `const` qualifiers, would it make sense to add them to `sshash::dictionary` in our `sshash` fork instead?

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,494 @@
...
+
+constexpr DeBruijnGraph::node_index sshash_to_graph_index(uint64_t idx) {
+    return idx + 1;
+}
+constexpr uint64_t graph_index_to_sshash(DeBruijnGraph::node_index idx) {
+    return idx - 1;
+}
+
+DBGSSHash::DBGSSHash(size_t k, Mode mode) : k_(k), num_nodes_(0), mode_(mode) {
+    size_t odd_k = k_ % 2 == 0 ? k_ + 1 : k_;
```
adamant-pwn, line 85: nit: `k_ | 1`?

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,494 @@
...
+
+    if (odd_k * IDictionary::bits_per_char <= 64) {
+        dict_ = std::make_unique<Dictionary<Kmer64>>();
+    } else if (odd_k * IDictionary::bits_per_char <= 128) {
+        dict_ = std::make_unique<Dictionary<Kmer128>>();
+    } else if (odd_k * IDictionary::bits_per_char <= 256) {
+        dict_ = std::make_unique<Dictionary<Kmer256>>();
+    } else {
+        common::logger->error("k too big: {} > {}", odd_k, 256 / IDictionary::bits_per_char);
+        throw std::runtime_error("k fail");
```
adamant-pwn, line 95: nit: `std::length_error` would probably be more appropriate.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,494 @@
...
+    }
+}
+
+DBGSSHash::DBGSSHash(std::string const& input_filename, size_t k, Mode mode)
+      : DBGSSHash(k, mode) {
+    sshash::build_configuration build_config;
+    build_config.k = k;
+    // quick fix for value of m... k/2 but odd
+    build_config.m = std::min(uint64_t((k_ + 1) / 2), sshash::constants::max_m);
```
adamant-pwn, line 104: TODO in sshash: `sshash::constants::max_m` should be moved to `kmer_t::max_m`, since we now work with large alphabets. Metagraph should also accommodate to somehow be aware of an actual `kmer_t`?

adamant-pwn: Also if `m` must be odd, it's better to use `(k_ / 2) | 1`, otherwise for `k_ = 7` we'll have `m = 4`.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,494 @@
...
+    build_config.verbose = common::get_verbose();
+    build_config.num_threads = get_num_threads();
+    if (build_config.m % 2 == 0)
```
adamant-pwn, line 107: Just initiate `m` with `(k_ / 2) | 1`?

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,494 @@
...
+        build_config.m++;
+
+    // silence sshash construction messages when not verbose
+    if (!common::get_verbose())
+        std::cout.setstate(std::ios_base::failbit);
+
+    dict_->build(input_filename, build_config);
+    if (!common::get_verbose())
+        std::cout.clear();
```
adamant-pwn, line 116: Eh, somewhat hacky. Generally I'd be wary of doing this because what if e.g. some third party uses metagraph and wants to silence something too, and we would unsilence it prematurely? I feel like
```cpp
std::ostringstream local;
auto cout_buff = std::cout.rdbuf();
std::cout.rdbuf(local.rdbuf());

/* call library function here !!! */

std::cout.rdbuf(cout_buff);
```
from [here](https://cplusplus.com/forum/general/283883/#msg1229098) is a better way of silencing, because it would at least restore `std::cout` to exactly the state it was before the call. Welp, hope we won't get into multithreading issues while doing so...

metagraph/src/graph/representation/hash/dbg_sshash.cpp
adamant-pwn, line 21: Should we maybe have a base untemplated class in sshash fork instead, and inherit `sshash::dictionary` from it?

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,494 @@
...
+    num_nodes_ = dict_->size();
+}
+
+std::string DBGSSHash::file_extension() const {
+    return kExtension;
+}
+
+void DBGSSHash::add_sequence(std::string_view sequence,
+                             const std::function<void(node_index)>& on_insertion) {
+    throw std::runtime_error("adding sequences not implemented");
```
adamant-pwn, line 127: nit: it seems that `std::logic_error` is considered more appropriate for these things. Also it should probably be "not supported" rather than "not implemented" because by design sshash is a static structure?

P.S. Though I suppose there are certain tricks that would make it expandable at the cost of extra $\log_2 n$ factor, where $n$ is the number of sequences...

metagraph/src/graph/representation/hash/dbg_sshash.cpp
adamant-pwn: Or, alternatively, use the approach similar to [dbg_hash_fast.cpp](https://github.com/ratschlab/metagraph/blob/master/metagraph/src/graph/representation/hash/dbg_hash_fast.cpp), that is structure `DBGSSHash` this way, rather than `sshash::dictionary`? Then it'd be more consistent with how other DBG implementations are done.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,494 @@
...
+                             const std::function<bool()>& terminate) const {
+    if (mode_ != CANONICAL) {
+        map_to_nodes_sequentially(sequence, callback, terminate);
+    } else {
+        map_to_nodes_with_rc(
+                sequence, [&](node_index node, bool) { callback(node); }, terminate);
+    }
+}
+
+DBGSSHash::node_index DBGSSHash::reverse_complement(node_index node) const {
```
adamant-pwn, line 141: How does sshash handles this? Does it have the same index for a k-mer and its reverse complement?

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,494 @@
...
+    throw std::runtime_error("adding sequences not implemented");
+}
+
+void DBGSSHash::map_to_nodes(std::string_view sequence,
+                             const std::function<void(node_index)>& callback,
+                             const std::function<bool()>& terminate) const {
+    if (mode_ != CANONICAL) {
+        map_to_nodes_sequentially(sequence, callback, terminate);
+    } else {
+        map_to_nodes_with_rc(
```
adamant-pwn, line 136: Is there a reason to distinguish between this and [this](https://github.com/ratschlab/metagraph/pull/481/files#diff-e90dbf65e72f4eaf563486b42faf095ec39582fe0b24acacea4b1bc3a7b6d79fR177-R179)?

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,494 @@
...
+    };
+
+    if (auto dict64 = dynamic_cast<Dictionary<Kmer64>*>(dict_.get())) {
+        stream_map(*dict64);
+    } else if (auto dict128 = dynamic_cast<Dictionary<Kmer128>*>(dict_.get())) {
+        stream_map(*dict128);
+    } else if (auto dict256 = dynamic_cast<Dictionary<Kmer256>*>(dict_.get())) {
+        stream_map(*dict256);
+    } else {
+        common::logger->error("");
```
adamant-pwn, line 201: Probably better to provide a meaningful error message here. On the other hand, maybe it's better to assert that this never happens?

metagraph/src/graph/representation/hash/dbg_sshash.cpp
adamant-pwn: Or maybe use
```cpp
std::variant<
    sshash::dictionary<Kmer64>,
    sshash::dictionary<Kmer128>,
    sshash::dictionary<Kmer256>>
```

with `std::visit`, to reduce overhead in polymorphysm of dictionary? It may simplify [this](https://github.com/ratschlab/metagraph/pull/481/files#diff-e90dbf65e72f4eaf563486b42faf095ec39582fe0b24acacea4b1bc3a7b6d79fR194-R199) and similar checks, and also remove the need of repeatedly declaring functions that just mirror base behavior, if I understand correctly.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,494 @@
...
+        throw std::runtime_error("k fail");
+    }
+}
+
+void DBGSSHash::map_to_nodes_with_rc(std::string_view sequence,
```
adamant-pwn, line 206: It looks very annoying that we need to implement essentially the same thing twice. Maybe we can factor it our in a function that has a `template<bool rc = false>` to control whether we want to use this variant or the previous one?

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,494 @@
...
+}
+
+DBGSSHash::node_index DBGSSHash::traverse(node_index node, char next_char) const {
+    // TODO: if a node is in the middle of a unitig, then we only need to check the next node index
+    return kmer_to_node(get_node_sequence(node).substr(1) + next_char);
+}
+
+DBGSSHash::node_index DBGSSHash::traverse_back(node_index node, char prev_char) const {
+    // TODO: if a node is in the middle of a unitig, then we only need to check the previous node index
+    std::string string_kmer = std::string(1, prev_char) + get_node_sequence(node);
```
adamant-pwn, line 250: nit: wouldn't `prev_char + ...` work too?

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,494 @@
...
+    string_kmer.pop_back();
+    return kmer_to_node(string_kmer);
+}
+
+void DBGSSHash::adjacent_outgoing_nodes(node_index node,
+                                        const std::function<void(node_index)>& callback) const {
+    assert(node > 0 && node <= num_nodes());
```
adamant-pwn, line 257: Is there a reason to have this check here specifically, but not in other places where `node_index node` is an argument?

metagraph/CMakeLists.txt
hmusta: correct, this isn't needed anymore

metagraph/src/cli/build.cpp
hmusta:
```suggestion
    } else {
```

metagraph/src/cli/config/config.cpp
hmusta:
```suggestion
    } else if (string == "sshash") {
```

metagraph/src/graph/representation/canonical_dbg.cpp
hmusta: DBGSSHash and DBGSuccinct have special functionality that allows us to implement these methods more efficiently, whereas the other types work fine with the naive implementation

metagraph/src/graph/representation/hash/dbg_sshash.cpp
hmusta: I'd prefer to have an untemplated parent class in sshash to handle this, then we can initialize the dictionary with the appropriate template in the constructer and load. For this to work consistently, however, we'd need to implement a sequential access method to avoid exposing `kmer_t`.

I don't really like the way it's done in DBGHashFast since there are a lot of methods that need to be implemented, so it produces a lot of extra code.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,494 @@
...
+                                    const IncomingEdgeCallback& callback) const {
+    assert(node > 0 && node <= num_nodes());
+    std::string kmer = get_node_sequence(node);
+    auto call_impl = [&](const auto &dict) {
+        using kmer_t = typename std::remove_reference<decltype(dict)>::type::kmer_t;
+        sshash::neighbourhood<kmer_t> nb = dict.kmer_backward_neighbours(kmer.c_str(), mode_ == CANONICAL);
+        for (size_t i = 0; i < kmer_t::alphabet_size; i++) {
+            if (nb.backward[i].kmer_id != sshash::constants::invalid_uint64) {
+                node_index prev = sshash_to_graph_index(nb.backward[i].kmer_id);
+                callback(nb.backward[i].kmer_orientation ? reverse_complement(prev) : prev, kmer_t::alphabet[i]);
```
adamant-pwn, line 304: See above.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,494 @@
...
+        call_impl(*dict128);
+    } else if (auto dict256 = dynamic_cast<Dictionary<Kmer256>*>(dict_.get())) {
+        call_impl(*dict256);
+    } else {
+        common::logger->error("");
+        throw std::runtime_error("k fail");
+    }
+}
+
+void DBGSSHash::call_outgoing_kmers_with_rc(
```
adamant-pwn, line 321: I'd also prefer to govern this (and the next function) via `template<bool rc = false>` too.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
hmusta: no, this check should actually be on all such functions

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,494 @@
...
+    adjacent_incoming_nodes(node, [&](node_index) { ++res; });
+    return res;
+}
+
+void DBGSSHash::call_nodes(
+        const std::function<void(node_index)>& callback,
+        const std::function<bool()> &terminate) const {
+    for (size_t node_idx = 1; !terminate() && node_idx <= dict_->size(); ++node_idx) {
+        callback(node_idx);
+        if (mode_ == CANONICAL && !terminate()) {
```
adamant-pwn, line 394: Does `call_nodes` provide any guarantees on the order of calls? It may be better to run rc nodes in a second loop, so that we don't constantly jump between the two node types in memory.

metagraph/CMakeLists.txt
hmusta:
```suggestion
```

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,494 @@
...
+    if (!num_nodes())
+        return npos;
+
+    auto res = dict_->lookup_advanced(kmer.data(), mode_ == CANONICAL);
+    node_index node = sshash_to_graph_index(res.kmer_id);
+    return res.kmer_orientation ? reverse_complement(node) : node;
+}
+
+std::pair<DBGSSHash::node_index, bool>
+DBGSSHash::kmer_to_node_with_rc(std::string_view kmer) const {
```
adamant-pwn, line 412: Let's govern this by `template<bool rc = false>` as well.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,494 @@
...
+
+    if (node > node_canonical)
+        ::reverse_complement(str_kmer.begin(), str_kmer.end());
+
+    return str_kmer;
+}
+
+void DBGSSHash::serialize(std::ostream& out) const {
+    // TODO
+    throw std::runtime_error("serialize to stream not implemented");
```
adamant-pwn, line 434: nit: `std::logic_error`?

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,494 @@
...
+    saver.visit(const_cast<size_t&>(num_nodes_));
+    saver.visit(const_cast<size_t&>(k_));
+    saver.visit(const_cast<Mode&>(mode_));
+
+    if (num_nodes())
+        dict_->visit(saver);
+}
+
+bool DBGSSHash::load(std::istream& in) {
+    throw std::runtime_error("load from stream not implemented");
```
adamant-pwn, line 451: nit: `std::logic_error`?

metagraph/src/graph/representation/hash/dbg_sshash.cpp
hmusta: in CANONICAL mode, only one orientation is stored in memory. The index of the reverse complement is simply the same index offset by the size of the underlying dictionary, so both orientations of a k-mer, in fact, point to the same "location" in the underlying memory.

If we were to do these separately instead of interleaved, we'd essentially have to do two passes through the dictionary instead of one.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,494 @@
...
+    *this = DBGSSHash(k, mode);
+    num_nodes_ = num_nodes;
+
+    if (num_nodes_)
+        dict_->visit(loader);
+
+    return true;
+}
+
+bool DBGSSHash::operator==(const DeBruijnGraph& other) const {
```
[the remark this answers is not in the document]
hmusta: We only use this for unit testing; in real-world usage, we never actually have to do this.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,494 @@
...
+    try {
+        other.call_kmers([&](node_index, const std::string &kmer) {
+            if (!find(kmer))
+                throw std::bad_function_call();
```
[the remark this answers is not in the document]
hmusta: yes, this is a hack for early termination. I guess a better approach would be to add a `terminate` callback to `call_kmers`

metagraph/src/graph/representation/hash/dbg_sshash.cpp
hmusta: sshash actually does have a verbose mode, but they unfortunately don't have a quiet mode...the ideal thing would be for us to add it

metagraph/tests/graph/all/test_dbg_helpers.cpp
```diff
@@ -39,6 +40,9 @@ template<> size_t max_test_k<DBGHashFast>() {
 template<> size_t max_test_k<DBGHashString>() {
     return 100;
 }
+template<> size_t max_test_k<DBGSSHash>() {
+    return 255 / kmer::KmerExtractor2Bit::bits_per_char;
```
adamant-pwn, line 44: `DBGSSHash` currently uses the value from `IDictionary` rather than from `KmerExtractor2Bit`.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
hmusta: I think it'll be a lot of work to get MetaGraph to work with `kmer_t`, it might be better in the end to have methods in the dictionary class that handle the high-level functionality so we don't have to expose `kmer_t`.

Also, moving `max_m` to `kmer_t` would require making more changes to how sshash computes minimizers. AFAIK, they assume that each m-mer can fit in a uint64_t

metagraph/tests/graph/all/test_dbg_helpers.cpp
```diff
@@ -126,6 +132,62 @@ build_graph<DBGBitmap>(uint64_t k,
     return graph;
 }
 
+void writeFastaFile(const std::vector<std::string>& sequences, const std::string& outputFilename) {
+    std::ofstream fastaFile(outputFilename);
+
+    if (!fastaFile.is_open()) {
+        std::cerr << "Error: Unable to open the output file." << std::endl;
```
adamant-pwn, line 139: Should we use `common::logger` instead?

metagraph/tests/graph/all/test_dbg_helpers.cpp
```diff
@@ -126,6 +132,62 @@ build_graph<DBGBitmap>(uint64_t k,
...
+        return;
+    }
+
+    for (size_t i = 0; i < sequences.size(); ++i) {
+        fastaFile << ">"<< "\n" << sequences[i] << "\n";
```
adamant-pwn, line 144: Missing whitespace in `">"<<`.

metagraph/tests/graph/all/test_dbg_helpers.cpp
```diff
@@ -126,6 +132,62 @@ build_graph<DBGBitmap>(uint64_t k,
...
+template <>
+std::shared_ptr<DeBruijnGraph>
+build_graph<DBGSSHash>(uint64_t k,
+                       std::vector<std::string> sequences,
+                       DeBruijnGraph::Mode mode) {
+    if (sequences.empty())
+        return std::make_shared<DBGSSHash>(k, mode);
+
+    // use DBGHashString to get contigs for SSHash
+    auto string_graph = build_graph<DBGHashString>(k, sequences, mode);
```
adamant-pwn, line 158: Are there faster methods for this?

metagraph/tests/graph/all/test_dbg_helpers.cpp
hmusta:
```suggestion
        common::logger->error("Unable to open output file {}", outputFilename);
```

metagraph/tests/graph/all/test_dbg_helpers.cpp
hmusta:
```suggestion
        fastaFile << ">" << "\n" << sequences[i] << "\n";
```

metagraph/tests/graph/all/test_dbg_helpers.cpp
```diff
@@ -126,6 +132,62 @@ build_graph<DBGBitmap>(uint64_t k,
...
+
+    std::string dump_path = "../tests/data/sshash_sequences/contigs.fa";
+    writeFastaFile(contigs, dump_path);
+
+    std::shared_ptr<DBGSSHash> graph;
+    try {
+        graph = std::make_shared<DBGSSHash>(dump_path, k, mode);
+    } catch (const std::runtime_error &e) {
+        if (strcmp(e.what(), "each partition must contain more than one key: use less partitions") == 0) {
+            return build_graph<DBGHashOrdered>(k, sequences, mode);
```
adamant-pwn, line 178: Why?

metagraph/tests/graph/all/test_dbg_helpers.cpp
```diff
@@ -126,6 +132,62 @@ build_graph<DBGBitmap>(uint64_t k,
...
+        } else {
+            throw e;
+        }
+    }
+
+    if (mode == DeBruijnGraph::PRIMARY)
+        return std::make_shared<CanonicalDBG>(
```
adamant-pwn, line 185: Why?

metagraph/tests/graph/all/test_dbg_helpers.cpp
hmusta: The idea here is that pthash assumes that each partition has at least two elements, which is violated for small graphs. So, we first try to run with DBGSSHash and switch over to a different graph if it fails.

See here for the issue I opened with pthash about this"
https://github.com/jermp/pthash/issues/63

The ideal thing would be to modify pthash to work with empty or single-element partitions so we don't have to do this hack.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
adamant-pwn: It still should be moved to `kmer_t`, even under assumption that it fits `uint64_t`, because `max_m` should depend on `bits_per_char`, which is a property of `kmer_t`.

metagraph/tests/graph/all/test_dbg_helpers.cpp
hmusta: A PRIMARY-mode graph only has one orientation for each k-mer, so we use CanonicalDBG as a wrapper class to gain access to both orientations of each k-mer while only storing one.

adamant-pwn (changes requested):
Thanks for the updates! I think I marked all my current concerns in the open comments.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
hmusta: In sshash, they use the same index for both orientations. They have a bool flag in the lookup_result struct that stores whether your query was in the same or the opposite orientation from the k-mer that's stored

metagraph/src/graph/representation/hash/dbg_sshash.cpp
adamant-pwn: What do you think about using `std::variant<sshash::dictionary<...>>` instead? Is there a reason not to do it? The problem with introducing untemplated parent class to sshash itself, from my perspective, would be that sshash itself currently doesn't seem to have a need for it, so it'd be an unnecessary complication of their codebase.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
adamant-pwn: Let's try rdbuf swap for now? And we can open an issue in sshash upstream requesting a silent mode.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
adamant-pwn: Isn't doing one pass and doing something twice within it essentially the same as doing two passes and doing something once within them? But at least it'd guarantee that indices are called in an increasing order, so there should be no jumps in memory between regular and reversed nodes, if they're accessed by callback?

metagraph/src/graph/representation/hash/dbg_sshash.cpp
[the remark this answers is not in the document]
adamant-pwn: Okay... Would it make sense to move this implementation into a base class?

metagraph/src/graph/representation/hash/dbg_sshash.cpp
adamant-pwn: Since we also consider even `k` here, it might be that there are "unused" nodes for self-complement k-mers. Are we fine with that? Generally, in this case, maybe it's better to return `node + dict_->size()` anyway, so that we don't waste time on comparison of the node with its revcompl? Would it break anything if we have a "virtual" revcompl node when it is exactly equal to the initial node?

metagraph/tests/graph/all/test_dbg_helpers.cpp
adamant-pwn: But it doesn't test DBGSSHash if it builds some other kind of graph, so it would make the test somewhat bogus on DBGSSHash... Okay, I suppose we'll have to look into pthash eventually and try to fix it ourselves, if small cases are of no interest to Giulio. Maybe not necessarily as a part of this PR, but then we should keep track of it in an issue or smt.

metagraph/src/cli/build.cpp
adamant-pwn: Yeah, this is just a suggestion for the better wording of the error message.

adamant-pwn:
@hmusta hi, what's the current plan with this PR? As in, who should work on it at this point? I can try to implement the stuff that I suggested in the review, but I think I still need some input on some of the comments, esp. https://github.com/ratschlab/metagraph/pull/481#discussion_r1629606622.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
hmusta: I think there are some parts of the code that rely on the fact that nodes with different indices necessarily have different spellings. Since we don't use even k that much anyway, I'd rather make the code for that case correct rather than extra efficient.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
hmusta: which line is this referring to? the link doesn't work for me

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,234 @@
...
+    case 'T':
+        ssh_idx = nb.backward_T.kmer_id;
+        break;
+    default:
+        break;
+    }
+    return ssh_idx + 1;
+}
+
+void DBGSSHash ::adjacent_outgoing_nodes(node_index node,
```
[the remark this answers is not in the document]
hmusta: The reason why we kept this implementation out of the parent class is because we wanted to give a hint to any potential users that `adjacent_outgoing_nodes` may be implemented more efficiently than `call_outgoing_kmers`, so we force them to implement them separately.

Do you think this is too much hand-holding?

metagraph/src/graph/representation/hash/dbg_sshash.cpp
hmusta: ok, I implemented this

metagraph/src/graph/representation/hash/dbg_sshash.cpp
adamant-pwn: I think it was about one of cases in `map_to_nodes_sequentially, but the difference is somewhat clear now.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,356 @@
...
+    throw std::logic_error("adding sequences not supported");
+}
+
+void DBGSSHash::map_to_nodes(std::string_view sequence,
+                             const std::function<void(node_index)>& callback,
+                             const std::function<bool()>& terminate) const {
+    if (mode_ != CANONICAL) {
+        map_to_nodes_sequentially(sequence, callback, terminate);
+    } else {
+        map_to_nodes_with_rc(
```
adamant-pwn, line 92:
```suggestion
        map_to_nodes_with_rc<true>(
```
To make it more explicit and for consistency with all other uses.

hmusta: already fixed :)

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,395 @@
...
+
+    // use a value of m recommended here:
+    // https://twitter.com/giulio_pibiri/status/1803417277213114501
+    // otherwise, use the next odd value greater than or equal to floor(k / 2)
+    build_config.m = std::min(
+        num_chars > 0
+            ? uint64_t(ceil(log(static_cast<double>(num_chars)) / log(static_cast<double>(alphabet_.size()))) + 1)
+            : uint64_t(k / 2) | 1,
+        sshash::constants::max_m
+    );
```
adamant-pwn, line 70: Should we do `build_config.m |= 1` after initial assignment? otherwise the first case may accidentally give even `m`, and I'm not sure if sshash is good with them.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,395 @@
...
+        auto res = d.lookup_advanced(kmer.data(), with_rc);
+        return std::make_pair(sshash_to_graph_index(res.kmer_id), res.kmer_orientation);
+    }, dict_);
+}
+template <>
+std::pair<DBGSSHash::node_index, bool>
+DBGSSHash::kmer_to_node_with_rc<true>(std::string_view) const;
+template <>
+std::pair<DBGSSHash::node_index, bool>
+DBGSSHash::kmer_to_node_with_rc<false>(std::string_view) const;
```
adamant-pwn, line 327: Huh, why? Isn't it enough to have `template<bool with_rc>` version above?

metagraph/src/graph/representation/canonical_dbg.cpp
adamant-pwn: I don't really understand... Why do we continue mapping here, even where there is a mismatch? It seems inconsistent with how other graph types are handled. And I don't really see how the implementation is more efficient this way? I mean, `map_to_nodes_sequentially` does exactly this when `mode == CANONICAL` anyway?

metagraph/src/graph/representation/hash/dbg_sshash.cpp
adamant-pwn: Oh, is it to fix linking errors? You probably want to move it to the header then, I assume?

metagraph/tests/graph/all/test_dbg_helpers.cpp
adamant-pwn: Eh? I assumed the name `CanonicalDBG` meant it's supposed to be operated while working in `mode == CANONICAL`. Should there be some kind of assert in DBGSSHash then that PRIMARY mode is not supported directly? It looks somewhat confusing...

hmusta: You can still build a PRIMARY-mode graph and do k-mer set operations on it, so I wouldn't exactly say that it's not supported. It's just not useful for directly querying.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
hmusta: Hmm...so it turned out that rearranging the functions in the cpp fixed things. The rule is that one should provide the implementations of the template methods before the implementations of any other methods that instantiate them.

hmusta: > Oh, is it to fix linking errors? You probably want to move it to the header then, I assume?

Given that the implementation of `kmer_to_node_with_rc` uses some implementation details of the class (e.g., the node index shifting), I'd rather keep it in the cpp if possible.

metagraph/src/graph/representation/canonical_dbg.cpp
hmusta: The idea here is that if a k-mer's not found (a mismatch), we don't know if the k-mer is truly missing, or if it's only present in the reverse complement orientation. So, we have to check both. With the SSHash graph, the underlying dictionary class already handles this check, so I'm delegating this logic to that class (assuming that it's more efficient than us implementing the logic with its API)

metagraph/src/graph/representation/hash/dbg_sshash.cpp
hmusta: I'm not sure if `std::logic_error` is appropriate here. The only reason why this is not supported is because the `essentials::saver` doesn't have a constructor that takes an `std::ostream`, though it could.

hmusta: We can maybe open a pull request there.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
hmusta: > Hmm...so it turned out that rearranging the functions in the cpp fixed things. The rule is that one should provide the implementations of the template methods before the implementations of any other methods that instantiate them.

turns out it did need explicit instantiation, I was just doing it wrong...

metagraph/src/graph/representation/hash/dbg_sshash.cpp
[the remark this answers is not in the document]
hmusta: Fixed. I also replaced a few of the `bad_function_call` calls with a different custom exception to make it more robust.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
hmusta: hmm...yea, I just realized that if some external process relies on node indices to access an array, then this will indeed lead to jumps in memory. So let's do this as two loops then, since it's a cheap operation anyway.

metagraph/src/cli/build.cpp
hmusta: We're using `.at` so that it would throw an exception if there are no files passed in. In the end, this is only a single call.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
hmusta: I wrote a custom saver and loader so we can implement these.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
hmusta: see above

metagraph/src/graph/representation/hash/dbg_sshash.cpp
hmusta: I'm afraid of multithreading issues, to be honest, that's why I didn't use the approach you're suggesting. But I agree that the one I did is not ideal either...

adamant-pwn: Is there any reason to believe that the current approach is more thread-safe than rdbuf swap? As the link says, 
> **Any** "hack" that temporarily disables/redirects std::cout around library calls is not thread-safe.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
adamant-pwn: Could be a nice thing to add in the future... I suppose not necessarily as a part of this PR.

metagraph/src/graph/representation/canonical_dbg.cpp
adamant-pwn: Okay, fair enough I suppose.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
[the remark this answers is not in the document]
adamant-pwn: Yes, I'd say so. I think if the users really need it to be more efficient, they'll figure out that it should be overriden, and this argument kind of applies to any other virtual function with a default implementation anyway?

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,455 @@
...
+#endif
+
+template <typename T>
+struct template_parameter;
+
+template <template <typename ...> class C, typename T>
+struct template_parameter<C<T>> { using type = T; };
+
+template <typename T>
+using get_kmer_t = typename template_parameter<typename std::remove_const<typename std::remove_reference<T>::type>::type>::type;
```
adamant-pwn, line 27:
```suggestion
using get_kmer_t = typename template_parameter<std::decay_t<T>>::type;
```

metagraph/src/graph/representation/base/sequence_graph.cpp
```diff
@@ -80,9 +80,33 @@ bool DeBruijnGraph::find(std::string_view sequence,
...
+::adjacent_outgoing_nodes(node_index node,
+                          const std::function<void(node_index)> &callback) const {
+    assert(node > 0 && node <= max_index());
+    call_outgoing_kmers(node, [&](auto child, char) { callback(child); });
+}
+
+void DeBruijnGraph
+::adjacent_incoming_nodes(node_index node,
+                          const std::function<void(node_index)> &callback) const {
+    assert(node > 0 && node <= max_index());
```
adamant-pwn, line 93: Using `max_index()` should fix unit test failure on DBGHashFast, but is it intended that they are different?

metagraph/src/graph/alignment/aligner_labeled.cpp
```diff
@@ -11,6 +11,11 @@ namespace align {
 
 using mtg::common::logger;
 
+class early_term : public std::exception {
+  public:
+    const char* what() const noexcept override { return "Finished iteration"; }
+};
+
```
adamant-pwn, line 18: I think it would be much better to define it just once somewhere, rather than doing it over and over again in different files.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -0,0 +1,455 @@
...
+    std::visit([&](const auto &d) { d.access(ssh_idx, str_kmer.data()); }, dict_);
+
+    if (node > node_canonical)
+        ::reverse_complement(str_kmer.begin(), str_kmer.end());
+
+    return str_kmer;
+}
+
+
+struct generic_loader {
```
adamant-pwn, line 349: Should we try merging this into loader/saver upstream, rather than doing it here? Or at least do it in our fork of essentials and add it as a submodule directly, rather than using it as a submodule of a submodule of a submodule.

metagraph/src/graph/representation/base/sequence_graph.cpp
hmusta: yes, if there's a graph without a contiguous index space. we do require, however, than `max_index() >= num_nodes()`

metagraph/src/graph/representation/hash/dbg_sshash.cpp
hmusta: if we fork essentials and add this, won't it cause problems with `pthash` since there'll be two different versions of the same methods? Also, `dictionary::visit` is designed to work with anything that has interface (and we only use it once here), so unless Giulio has a need for a generic loader, we can just keep it here

adamant-pwn: I've made a bunch of forks for essentials and pthash and propagated necessary changes through them. It also includes making saving const, so in any case I think this is the way we should use them. I will also try to merge these changes upstream.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
hmusta: True...it was more of a hunch that changing the status of cout would be more thread safe than changing the rdbuf, but I have no actual proof of that.

adamant-pwn: I think I'd still try to use rdbuf for now then. I feel like messing with global state without properly reverting it back is a somewhat more unwanted and unexpected behavior for third-party users.  We can then reconsider if any actual issues occur?

hmusta: Sounds good

metagraph/src/graph/alignment/aligner_labeled.cpp
adamant-pwn: I moved it to `sequence_graph.hpp`. @hmusta Let me know if it's better to put in some other place.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
adamant-pwn: Changed to rdbuf swap. Let's see how it behaves now.

adamant-pwn (approved):
There is still one open comment related to https://github.com/jermp/pthash/issues/63. I think otherwise the current state is good? We should probably open a separate issue to track https://github.com/jermp/pthash/issues/63 if we merge this PR before it is fixed upstream.

metagraph/tests/graph/all/test_dbg_helpers.cpp
adamant-pwn: Should be fixed now that jermp/pthash#63 is settled. Closing for now, will reopen if tests still fail.

hmusta (approved):

PR #482, opened 2024-05-02
Fix CI failures for MacOS

adamant-pwn:
MacOS workflows currently produce the following error. The PR is a quick dirty fix to it.

```sh
Run pip3 install parameterized
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try brew install
    xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a Python library that isn't in Homebrew,
    use a virtual environment:
    
    python3 -m venv path/to/venv
    source path/to/venv/bin/activate
    python3 -m pip install xyz
    
    If you wish to install a Python application that isn't in Homebrew,
    it may be easiest to use 'pipx install xyz', which will manage a
    virtual environment for you. You can install pipx with
    
    brew install pipx
    
    You may restore the old behavior of pip by passing
    the '--break-system-packages' flag to pip, or by adding
    'break-system-packages = true' to your pip.conf file. The latter
    will permanently disable this error.
    
    If you disable this error, we STRONGLY recommend that you additionally
    pass the '--user' flag to pip, or set 'user = true' in your pip.conf
    file. Failure to do this can result in a broken Homebrew installation.
    
    Read more about this behavior here: <https://peps.python.org/pep-0668/>
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -167,7 +167,7 @@ jobs:
       run: brew install bzip2 libomp boost jemalloc autoconf automake libtool
 
     - name: install dependencies for integration tests
-      run: pip3 install parameterized
+      run: pip3 install parameterized --break-system-packages --user
 
     - name: install sdsl-lite
       run: |
--- a/metagraph/CMakeLists.txt
+++ b/metagraph/CMakeLists.txt
@@ -269,7 +269,6 @@ include_directories(
   external-libraries/spdlog/include
   external-libraries/zlib
   external-libraries/sdust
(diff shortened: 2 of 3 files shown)


adamant-pwn:
Apparently, https://github.com/ratschlab/metagraph/blob/580bae690d98e4c912d73f64a0de76379ddc7633/metagraph/src/common/utils/simd_utils.hpp#L88-L97 stops compiling after https://github.com/simd-everywhere/simde-no-tests/commit/02c7a67ed825018f9efdf2a7e4f39d8196f65337.

PR #484, opened 2024-05-22
Use BOSS indexing in DBGSuccinct + make RowDiff independent

adamant-pwn:
WIP pull request for CI.

### DIFF
--- a/metagraph/integration_tests/test_transform_anno.py
+++ b/metagraph/integration_tests/test_transform_anno.py
@@ -56,8 +56,9 @@ class TestColumnOperations(TestingBase):
         self.assertEqual(res.returncode, 0)
         out = res.stdout.decode().split('\n')[2:]
         self.assertEqual('labels:  100', out[0])
-        self.assertEqual('objects: 46960', out[1])
-        self.assertEqual('density: 0.0185072', out[2])
+        self.assertIn((out[1], out[2]), [
+            ('objects: 47633', 'density: 0.0182458'), # DBGSuccinct with dummy nodes
+            ('objects: 46960', 'density: 0.0185072')])
         self.assertEqual(f'representation: {self.anno_repr}', out[3])
 
     def tearDown(self):
@@ -82,8 +83,11 @@ class TestColumnOperations(TestingBase):
         self.assertEqual(res.returncode, 0)
         out = res.stdout.decode().split('\n')[2:]
         self.assertEqual('labels:  1', out[0])
-        self.assertEqual('objects: 46960', out[1])
-        self.assertEqual(f'density: {expected_density}', out[2])
+        split = out[2].split()
+        self.assertEqual(out[2], split[0] + ' ' + split[1])
+        density = float(split[1])
+        self.assertIn(out[1], ['objects: 47633', 'objects: 46960'])
+        self.assertLess(abs(density - expected_density * 46960 / 47633), 10**-6)
         self.assertEqual(f'representation: {self.anno_repr}', out[3])
 
     def test_aggregate_columns(self):
@@ -104,9 +108,11 @@ class TestColumnOperations(TestingBase):
         self.assertEqual(res.returncode, 0)
         out = res.stdout.decode().split('\n')[2:]
         self.assertEqual('labels:  1', out[0])
-        self.assertEqual('objects: 46960', out[1])
-        self.assertEqual(f'density: {expected_density}', out[2])
-        self.assertEqual(f'representation: {self.anno_repr}', out[3])
+        split = out[2].split()
+        self.assertEqual(out[2], split[0] + ' ' + split[1])
+        density = float(split[1])
+        self.assertIn(out[1], ['objects: 47633', 'objects: 46960'])
+        self.assertLess(abs(density - expected_density * 46960 / 47633), 10**-6)
 
     def test_aggregate_columns_filtered(self):
         self._check_aggregation_min_max_value(0, 0, 0)
--- a/metagraph/integration_tests/test_transform_anno.py
+++ b/metagraph/integration_tests/test_transform_anno.py
@@ -30,12 +30,14 @@ class TestColumnOperations(TestingBase):
                          cls.tempdir.name + '/graph',
                          20, cls.graph_repr, 'basic', '--mask-dummy')
 
-        res = cls._get_stats(f'{cls.tempdir.name}/graph{graph_file_extension[cls.graph_repr]}')
-        assert(res.returncode == 0)
-        out = res.stdout.decode().split('\n')[2:]
-        assert('k: 20' == out[0])
-        assert('nodes (k): 46960' == out[1])
-        assert('mode: basic' == out[2])
+        stats_graph = cls._get_stats(f'{cls.tempdir.name}/graph{graph_file_extension[cls.graph_repr]}')
+        assert(stats_graph['returncode'] == 0)
+        assert(stats_graph['k'] == '20')
+        assert(stats_graph['nodes (k)'] == '46960')
+        assert(stats_graph['mode'] == 'basic')
+
+        cls.num_nodes = stats_graph['nodes (k)']
+        cls.max_index = stats_graph['max index (k)']
 
         cls._annotate_graph(
             TEST_DATA_DIR + '/transcripts_100.fa',
@@ -52,14 +54,15 @@ class TestColumnOperations(TestingBase):
         self.annotation = f'annotation{anno_file_extension[self.anno_repr]}';
 
         # check annotation
-        res = self._get_stats(f'-a {self.annotation}')
-        self.assertEqual(res.returncode, 0)
-        out = res.stdout.decode().split('\n')[2:]
-        self.assertEqual('labels:  100', out[0])
-        self.assertIn((out[1], out[2]), [
-            ('objects: 47633', 'density: 0.0182458'), # DBGSuccinct with dummy nodes
-            ('objects: 46960', 'density: 0.0185072')])
-        self.assertEqual(f'representation: {self.anno_repr}', out[3])
+        stats_annotation = self._get_stats(f'-a {self.annotation}')
+        self.assertEqual(stats_annotation['returncode'], 0)
+        self.assertEqual(stats_annotation['labels'], '100')
+        self.assertEqual(stats_annotation['objects'], self.max_index)
+        self.assertAlmostEqual(
+            float(stats_annotation['density']),
+            0.0185072 * int(self.num_nodes) / int(self.max_index),
+            places=6)
+        self.assertEqual(stats_annotation['representation'], self.anno_repr)
 
     def tearDown(self):
         os.chdir(self.old_cwd)
@@ -79,16 +82,15 @@ class TestColumnOperations(TestingBase):
         res = subprocess.run(command.split(), stdout=PIPE)
         self.assertEqual(res.returncode, 0)
 
-        res = self._get_stats(f'-a aggregated{anno_file_extension[self.anno_repr]}')
-        self.assertEqual(res.returncode, 0)
-        out = res.stdout.decode().split('\n')[2:]
-        self.assertEqual('labels:  1', out[0])
-        split = out[2].split()
-        self.assertEqual(out[2], split[0] + ' ' + split[1])
-        density = float(split[1])
-        self.assertIn(out[1], ['objects: 47633', 'objects: 46960'])
-        self.assertLess(abs(density - expected_density * 46960 / 47633), 10**-6)
-        self.assertEqual(f'representation: {self.anno_repr}', out[3])
+        stats_annotation = self._get_stats(f'-a aggregated{anno_file_extension[self.anno_repr]}')
+        self.assertEqual(stats_annotation['returncode'], 0)
+        self.assertEqual(stats_annotation['labels'], '1')
+        self.assertEqual(stats_annotation['objects'], self.max_index)
+        self.assertAlmostEqual(
+            float(stats_annotation['density']),
+            float(expected_density) * int(self.num_nodes) / int(self.max_index),
+            places=5)
+        self.assertEqual(stats_annotation['representation'], self.anno_repr)
 
     def test_aggregate_columns(self):
         self._check_aggregation_min(0, 1)
@@ -104,15 +106,15 @@ class TestColumnOperations(TestingBase):
         res = subprocess.run(command.split(), stdout=PIPE)
         self.assertEqual(res.returncode, 0)
 
-        res = self._get_stats(f'-a aggregated{anno_file_extension[self.anno_repr]}')
-        self.assertEqual(res.returncode, 0)
-        out = res.stdout.decode().split('\n')[2:]
-        self.assertEqual('labels:  1', out[0])
-        split = out[2].split()
-        self.assertEqual(out[2], split[0] + ' ' + split[1])
-        density = float(split[1])
-        self.assertIn(out[1], ['objects: 47633', 'objects: 46960'])
-        self.assertLess(abs(density - expected_density * 46960 / 47633), 10**-6)
+        stats_annotation = self._get_stats(f'-a aggregated{anno_file_extension[self.anno_repr]}')
+        self.assertEqual(stats_annotation['returncode'], 0)
+        self.assertEqual(stats_annotation['labels'], '1')
+        self.assertEqual(stats_annotation['objects'], self.max_index)
+        self.assertAlmostEqual(
+            float(stats_annotation['density']),
+            float(expected_density) * int(self.num_nodes) / int(self.max_index),
+            places=5)
+        self.assertEqual(stats_annotation['representation'], self.anno_repr)
 
     def test_aggregate_columns_filtered(self):
         self._check_aggregation_min_max_value(0, 0, 0)
--- a/metagraph/src/annotation/binary_matrix/row_diff/row_diff.cpp
+++ b/metagraph/src/annotation/binary_matrix/row_diff/row_diff.cpp
@@ -64,22 +64,22 @@ IRowDiff::get_rd_ids(const std::vector<BinaryMatrix::Row> &row_ids) const {
                 graph::AnnotatedSequenceGraph::anno_to_graph_index(row);
 
         while (true) {
-            row = graph::AnnotatedSequenceGraph::graph_to_anno_index(
-                graph_->validate_edge(boss_edge)
-            );
+            if (graph_->is_valid(boss_edge)) {
+                row = graph::AnnotatedSequenceGraph::graph_to_anno_index(boss_edge);
 
-            auto [it, is_new] = node_to_rd.try_emplace(row, node_to_rd.size());
-            rd_paths_trunc[i].push_back(it.value());
+                auto [it, is_new] = node_to_rd.try_emplace(row, node_to_rd.size());
+                rd_paths_trunc[i].push_back(it.value());
 
-            // If a node had been reached before, we interrupt the diff path.
-            // The annotation for that node will have been reconstructed earlier
-            // than for other nodes in this path as well. Thus, we will start
-            // reconstruction from that node and don't need its successors.
-            if (!is_new)
-                break;
+                // If a node had been reached before, we interrupt the diff path.
+                // The annotation for that node will have been reconstructed earlier
+                // than for other nodes in this path as well. Thus, we will start
+                // reconstruction from that node and don't need its successors.
+                if (!is_new)
+                    break;
 
-            if (anchor_[row])
-                break;
+                if (anchor_[row])
+                    break;
+            }
 
             boss_edge = boss.row_diff_successor(boss_edge, rd_succ);
         }
--- a/metagraph/src/annotation/binary_matrix/row_diff/row_diff.cpp
+++ b/metagraph/src/annotation/binary_matrix/row_diff/row_diff.cpp
@@ -64,7 +64,7 @@ IRowDiff::get_rd_ids(const std::vector<BinaryMatrix::Row> &row_ids) const {
                 graph::AnnotatedSequenceGraph::anno_to_graph_index(row);
 
         while (true) {
-            if (graph_->is_valid(boss_edge)) {
+            if (graph_->in_graph(boss_edge)) {
                 row = graph::AnnotatedSequenceGraph::graph_to_anno_index(boss_edge);
 
                 auto [it, is_new] = node_to_rd.try_emplace(row, node_to_rd.size());
--- a/metagraph/src/annotation/binary_matrix/row_diff/row_diff.cpp
+++ b/metagraph/src/annotation/binary_matrix/row_diff/row_diff.cpp
@@ -60,12 +60,11 @@ IRowDiff::get_rd_ids(const std::vector<BinaryMatrix::Row> &row_ids) const {
     for (size_t i = 0; i < row_ids.size(); ++i) {
         Row row = row_ids[i];
 
-        graph::boss::BOSS::edge_index boss_edge = graph_->kmer_to_boss_index(
-                graph::AnnotatedSequenceGraph::anno_to_graph_index(row));
+        graph::boss::BOSS::edge_index boss_edge = 
+                graph::AnnotatedSequenceGraph::anno_to_graph_index(row);
 
         while (true) {
-            row = graph::AnnotatedSequenceGraph::graph_to_anno_index(
-                    graph_->boss_to_kmer_index(boss_edge));
+            row = graph::AnnotatedSequenceGraph::graph_to_anno_index(boss_edge);
 
             auto [it, is_new] = node_to_rd.try_emplace(row, node_to_rd.size());
             rd_paths_trunc[i].push_back(it.value());
--- a/metagraph/src/annotation/binary_matrix/row_diff/row_diff.cpp
+++ b/metagraph/src/annotation/binary_matrix/row_diff/row_diff.cpp
@@ -64,7 +64,9 @@ IRowDiff::get_rd_ids(const std::vector<BinaryMatrix::Row> &row_ids) const {
                 graph::AnnotatedSequenceGraph::anno_to_graph_index(row);
 
         while (true) {
-            row = graph::AnnotatedSequenceGraph::graph_to_anno_index(boss_edge);
+            row = graph::AnnotatedSequenceGraph::graph_to_anno_index(
+                graph_->validate_edge(boss_edge)
+            );
 
             auto [it, is_new] = node_to_rd.try_emplace(row, node_to_rd.size());
             rd_paths_trunc[i].push_back(it.value());
--- a/metagraph/src/annotation/row_diff_builder.cpp
+++ b/metagraph/src/annotation/row_diff_builder.cpp
@@ -11,6 +11,7 @@
 #include "common/elias_fano/elias_fano_merger.hpp"
 #include "common/utils/file_utils.hpp"
 #include "common/vectors/bit_vector_sdsl.hpp"
+#include "common/vectors/bit_vector_dyn.hpp"
 #include "graph/annotated_dbg.hpp"
 
 const uint64_t BLOCK_SIZE = 1 << 25;
@@ -26,6 +27,7 @@ namespace annot {
 using namespace mtg::annot::matrix;
 using mtg::common::logger;
 using mtg::graph::boss::BOSS;
+using node_index = graph::DeBruijnGraph::node_index;
 namespace fs = std::filesystem;
 
 using anchor_bv_type = RowDiff<ColumnMajor>::anchor_bv_type;
@@ -264,6 +266,23 @@ void sum_and_call_counts(const fs::path &dir,
     }
 }
 
+std::shared_ptr<const bit_vector> get_last(const graph::DeBruijnGraph &graph) {
+    if (auto* dbg_succ = dynamic_cast<graph::DBGSuccinct const*>(&graph)) {
+        return std::shared_ptr<const bit_vector>(
+            std::shared_ptr<const bit_vector>{}, &dbg_succ->get_boss().get_last());
+    } else {
+        bit_vector_dyn last_bv(graph.max_index() + 1);
+        graph.call_nodes([&](node_index v) {
+            std::pair<char, node_index> last;
+            graph.call_outgoing_kmers(v, [&](node_index u, char c) {
+                last = std::max(last, std::pair{c, u});
+            });
+            last_bv.set(last.second, true);
+        });
+        return std::make_shared<bit_vector_dyn>(std::move(last_bv));
+    }
+}
+
 rd_succ_bv_type route_at_forks(const graph::DeBruijnGraph &graph,
                                const std::string &rd_succ_filename,
                                const std::string &count_vectors_dir,
@@ -282,7 +301,7 @@ rd_succ_bv_type route_at_forks(const graph::DeBruijnGraph &graph,
         logger->trace("RowDiff successors will be set to the adjacent nodes with"
                       " the largest number of labels");
 
-        const bit_vector &last = *graph.get_last();
+        const bit_vector &last = *get_last(graph);
         graph::DeBruijnGraph::node_index graph_idx = to_node(0);
 
         std::vector<uint32_t> outgoing_counts;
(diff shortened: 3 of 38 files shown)


adamant-pwn:
Hi @hmusta, can you took a look here? I think it's still far from complete, but I'd still apreciate some input on the general direction to make sure that what I'm doing is aligned with what we want.

metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -293,12 +302,12 @@ rd_succ_bv_type route_at_forks(const graph::DBGSuccinct &graph,
...
+                if (last[from_graph_index(graph, graph_idx)]) {
                     // pick the node with the largest count
                     size_t max_pos = std::max_element(outgoing_counts.rbegin(),
                                                       outgoing_counts.rend())
                                      - outgoing_counts.rbegin();
-                    rd_succ_bv[graph.kmer_to_boss_index(graph_idx - max_pos)] = true;
+                    rd_succ_bv[from_graph_index(graph, graph_idx - max_pos)] = true;
                     outgoing_counts.resize(0);
                 }
                 graph_idx++;
```
adamant-pwn, line 313: GitHub doesn't allow me to put comments on unedited sections :slightly_frowning_face: 

But essentially, changing `route_at_forks` is somewhat straightforward, but `build_pred_succ` and to a smaller extent `assign_anchors` appear to depend more heavily on the stuff from `DBGSuccinct`.

For example, instead of
```cpp
    graph::DBGSuccinct graph(2);
    logger->trace("Loading graph...");
    if (!graph.load(graph_fname)) {
        logger->error("Cannot load graph from {}", graph_fname);
        std::exit(1);
    }
```

What actual class should I try to load? Should it be a template parameter to `build_pred_succ` and `assign_anchors`? Also can you tell me a bit more on a high level what `build_pred_succ` and its BOSS-specific part in particular are supposed to do?

hmusta: I think it would make more sense to modify these functions so they take `const DeBruijnGraph&` as an argument instead of the path to the graph. There's a loading function in `cli/load/load_graph.hpp`, but we'd have to do more refactoring to make that accessible to this function.

`build_pred_succ` is supposed to create predecessor and successor arrays, so for each node index, they store the index of the first predecessor and last successor (respectively). "First" and "last" are determined lexicographically

hmusta: In the current version of the function, it accepts an optional bit vector `rd_succ` which marks nodes that are successors. If this vector exists, then the index of the successor of a node for which `rd_succ` stores `true` is used instead of the last successor

hmusta: For predecessors, the `bwd` + `call_incoming_to_target` methods can be replaced with `DeBruijnGraph::adjacent_incoming_nodes`

metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -378,34 +391,49 @@ void build_pred_succ(const std::string &graph_fname,
...
+                                    uint64_t node_index = succinct->boss_to_kmer_index(pred);
+                                    pred_buf.push_back(to_row(node_index));
+                                    pred_boundary_buf.push_back(0);
+                                }
                             }
-                        }
-                    );
+                        );
+                    }
+                }
```
adamant-pwn, line 423: I need some external opinion on whether this is actually equivalent to the new code below, and whether we want to preserve this legacy code that works with BOSS directly. For row_diff_traverse and row_diff_successor I simply make overloads in DBGSuccinct that call them in BOSS instead of using base class implementations. I assume we might want to factor this piece out into a separate function as well?

adamant-pwn:
First of all, I'll wait for workflows to run through to make sure it still works as intended on DBGSuccinct. After that, I wonder if there is some simple enough way to test this with some other graph types. Some of my older notes from a few months ago say the following:

- Test_row_diff.cpp checks work with annotations
- Look into build_anno_graph
- Run LabeledAlignerTest on other graph types

Would be nice to check whether it is still relevant.

hmusta:
We can test this by modifying the following block in `build_anno_graph` to work with `DeBruijnGraph`:
https://github.com/ratschlab/metagraph/blob/3da66a968c9ce757a3855ff67de8ea0abaec32a0/metagraph/tests/annotation/test_annotated_dbg_helpers.cpp#L92

Then, we can add instantiations for other graph types (let's start with DBGHashFast for now) with RowDiff annotations in these places:
https://github.com/ratschlab/metagraph/blob/3da66a968c9ce757a3855ff67de8ea0abaec32a0/metagraph/tests/annotation/test_annotated_dbg_helpers.cpp#L245
https://github.com/ratschlab/metagraph/blob/3da66a968c9ce757a3855ff67de8ea0abaec32a0/metagraph/tests/annotation/test_annotated_dbg.cpp#L1002
https://github.com/ratschlab/metagraph/blob/3da66a968c9ce757a3855ff67de8ea0abaec32a0/metagraph/tests/annotation/test_annotated_dbg.cpp#L1015
https://github.com/ratschlab/metagraph/blob/3da66a968c9ce757a3855ff67de8ea0abaec32a0/metagraph/tests/annotation/test_aligner_labeled.cpp#L53

metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -282,7 +300,7 @@ rd_succ_bv_type route_at_forks(const graph::DBGSuccinct &graph,
         logger->trace("RowDiff successors will be set to the adjacent nodes with"
                       " the largest number of labels");
 
-        const bit_vector &last = graph.get_boss().get_last();
+        const bit_vector &last = *graph.get_last();
```
hmusta, line 303:
```suggestion
        const bit_vector &last = graph->get_last();
```
Is there a reason not to do it like this? I think it looks cleaner

metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -451,7 +472,7 @@ void assign_anchors(const std::string &graph_fname,
             optimize_anchors = true;
     }
 
-    sdsl::bit_vector anchors_bv(boss.get_last().size(), false);
+    sdsl::bit_vector anchors_bv(graph.get_last()->size(), false);
```
hmusta, line 475:
```suggestion
    sdsl::bit_vector anchors_bv(graph.max_index() + 1, false);
```
there's no need to generate a last vector here if we're only using its size

metagraph/src/graph/representation/succinct/dbg_succinct.cpp
```diff
@@ -516,6 +516,27 @@ ::call_source_nodes(const std::function<void(node_index)> &callback) const {
     });
 }
 
+std::shared_ptr<const bit_vector> DBGSuccinct
+::get_last() const {
+    return std::shared_ptr<const bit_vector>(&get_boss().get_last(), [](const bit_vector*) {
+        // Do not destruct BOSS's last with shared_ptr
+    });
```
hmusta, line 523:
```suggestion
    return std::shared_ptr<const bit_vector>(std::shared_ptr<const bit_vector>{}, &get_boss().get_last());
```
No need to write a custom destructor, you can just use the aliasing contructor.

metagraph/src/graph/representation/base/sequence_graph.cpp
```diff
@@ -420,6 +421,71 @@ void DeBruijnGraph::call_unitigs(const CallPath &callback,
...
+    sdsl::bit_vector visited(max_index() + 1);
+    auto finalised = visited;
+    std::vector<size_t> distance(max_index() + 1);
+    assert(terminal->size() == visited.size());
+    assert(rd_succ.size() == visited.size());
+    auto set_terminal = [&](int v) {
+        distance[v] = 0;
+        (*terminal)[v] = true;
+    };
+    call_nodes([&](node_index v) {
```
hmusta, line 449: Once we know this works, can we add `num_threads` as a parameter to `call_nodes` so we can do this with multiple threads?

The `set_bit`, `fetch_bit`, etc. methods used in `BOSS::row_diff_traverse` can be used here for thread-safe atomic operations on `sdsl::bit_vector`.

hmusta (changes requested):

metagraph/src/annotation/row_diff_builder.cpp
hmusta: It's important to keep this legacy code since we don't preserve a `dummy` vector for BOSS tables to save space. We can also factor this out into a helper function if it cleans things up.

metagraph/src/annotation/row_diff_builder.cpp
adamant-pwn: Yes, it's the equivalent of `*(graph.get_last())` actually, not of `graph->get_last()`.

metagraph/src/graph/representation/succinct/dbg_succinct.cpp
adamant-pwn: Didn't know it's a thing, thanks for pointing out!

metagraph/src/annotation/row_diff_builder.cpp
adamant-pwn: @hmusta I merged this, but it seems to trigger the assert in workflows:

https://github.com/ratschlab/metagraph/blob/4f7b681c18d1c05e582d29375eb374e58613c608/metagraph/src/graph/representation/succinct/boss.cpp#L2791

So, the value we put here must _exactly_ match what `graph.get_last()->size()` returns, which is `boss().get_last()` in case of `DBGSuccinct`. I assume in BOSS it's actually not `max_index() + 1`?

hmusta: hmm...maybe I'm missing something in the call tree, but for `DBGSuccinct`, shouldn't it be the case that `max_index() + 1 == num_nodes() + 1 == boss.num_edges() + 1 == boss.get_W().size() - 1 + 1`?

adamant-pwn: Just checked, in that specific test case
```
max_index() + 1: 7,
num_nodes() + 1: 7,
boss.num_edges() + 1: 12,
boss.get_W().size(): 12,
last()->size(): 12
```
Because of dummy nodes maybe?

adamant-pwn: FYI, The test is https://github.com/ratschlab/metagraph/blob/f2f320fa3fd3c9472260abc2abe0857618d8181e/metagraph/tests/annotation/test_aligner_labeled.cpp#L61-L70

hmusta: Ahh, I think I get what's going on now. Some of the unit tests still generate a dummy mask for the BOSS table, so `num_nodes()` skips the dummy edges. In that case, we should use `boss.get_last().size()` for `DBGSuccinct` and `max_index() + 1` for everything else.

adamant-pwn: Also playing around with `kmer_to_boss_index` shows that the mapping is as follows:
```
1 -> (boss) 3
2 -> (boss) 4
3 -> (boss) 6
4 -> (boss) 8
5 -> (boss) 9
6 -> (boss) 10
```

hmusta: Makes sense, the corresponding BOSS edges look like this (with the node IDs after `->` if the graph has a mask)
```
 0 $$$ $
 1 $$$ A
 2 $$$ G
 3 CAA T -> 1
 4 GCA A -> 2
 5 $GC A
 6 TGC T -> 3
 7 $$G C
 8 ATG C -> 4
 9 AAT G -> 5
10 GCT T -> 6
11 CTT $
```

metagraph/src/annotation/row_diff_builder.cpp
adamant-pwn: Actually, upon reading this code again I don't really understand why we want to preserve it.

1. After doing `boss_idx = succinct->kmer_to_boss_index(i)`, isn't the check that `boss_idx` is dummy redundant? I thought dummy nodes are those that are outside of the graph nodes, so it shouldn't be possible to get mapped into one.
2. What follows within the `if` seems to be equivalent to `row_diff_successor` call, so why not just call it instead of re-implementing it?
3. The part of finding predecessors also seems to be equivalent to `adjacent_incoming_nodes`, because `adjacent_incoming_nodes` presumably shouldn't call any dummy nodes, but only actual nodes from the graph? I'm also not sure if this specific implementation is more efficient than `adjacent_incoming_nodes`?

hmusta: `kmer_to_boss_index` is the identity map if a dummy edge mask is not present, so the dummy edge checks are not done. Also, `adjacent_incoming_nodes` and `adjacent_outgoing_nodes` don't do dummy edge checks if a mask is not present.

hmusta: But I agree with (2), we can replace that block with `row_diff_successor`

adamant-pwn: UPD: Regarding 1, it seems that there are nodes such that `boss_idx` is not dummy, but `boss.fwd(boss_idx)` is dummy. This is very confusing to me, why does this ever happen?

metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -264,7 +264,31 @@ void sum_and_call_counts(const fs::path &dir,
...
+        return idx;
+    }
+}
+size_t get_last_size(const graph::DeBruijnGraph &graph) {
+    if (dynamic_cast<graph::DBGSuccinct const*>(&graph)) {
+        return graph.get_last()->size();
+    } else {
+        return graph.max_index() + 1;
+    }
+}
```
adamant-pwn, line 289: I feel like it's idiomatically better to make these virtual functions, but I'm not sure if it's appropriate to place them in `graph::DeBruijnGraph`?

metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -378,34 +397,28 @@ void build_pred_succ(const std::string &graph_fname,
...
+                auto j = graph.row_diff_successor(i, rd_succ);
+                succ_buf.push_back(to_row(j));
+                succ_boundary_buf.push_back(0);
+            }
+            if(rd_succ[from_graph_index(graph, i)]) {
+                graph.adjacent_incoming_nodes(i, [&](auto pred) {
+                    pred_buf.push_back(to_row(pred));
+                    pred_boundary_buf.push_back(0);
+                });
+            }
```
adamant-pwn, line 420: I still think it should be fine to do it this way even we work with DBGSuccinct. What I mean is, as I understand it, dummy nodes are supposed to be something BOSS-internal, and not exposed in DeBruijnGraph interface, right? Then, if a dummy node is returned from `adjacent_incoming_nodes`, shouldn't it be considered a bug of DBGSuccinct, rather than smt that needs to be bypassed in this specific place?

On a side note, it looks like `pred` here is never a dummy node here, at least in `LabeledAlignerTest` group.

metagraph/src/annotation/row_diff_builder.cpp
adamant-pwn: I'll close this for now, let's move to that refer to the current state of the code.

metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -378,34 +397,28 @@ void build_pred_succ(const std::string &graph_fname,
...
-                        [&](BOSS::edge_index pred) {
-                            // dummy predecessors are ignored
-                            if (!dummy[pred]) {
-                                uint64_t node_index = graph.boss_to_kmer_index(pred);
-                                pred_buf.push_back(to_row(node_index));
-                                pred_boundary_buf.push_back(0);
-                            }
-                        }
-                    );
+                if ((*dummy)[next]) {
```
adamant-pwn, line 406: I'm trying to understand what is the intention of this check in the context of `row_diff_successor` function. Generally, do we want `row_diff_successor` to be well-defined for all nodes of the graph and return a graph node? If yes, then this convention would be violated for terminal nodes. How do we want to handle it? Should `row_diff_successor` return an `std::optional` on nodes that don't have a successor? Which nodes should it be in the terms of `graph::DeBruijnGraph`, if any?

metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -378,34 +397,28 @@ void build_pred_succ(const std::string &graph_fname,
         std::vector<bool> pred_boundary_buf;
 
         for (uint64_t i = start; i < std::min(start + BS, graph.num_nodes() + 1); ++i) {
-            BOSS::edge_index boss_idx = graph.kmer_to_boss_index(i);
-            if (!dummy[boss_idx]) {
+            bool find_succ = true;
+            if (succinct) { // Legacy code for DBGSuccinct
+                const BOSS &boss = succinct->get_boss();
+                BOSS::edge_index boss_idx = succinct->kmer_to_boss_index(i);
```
adamant-pwn, line 403: I removed the check that `boss_idx` is not dummy, because, from my perspective, nodes from `1` to `num_nodes()` should never be mapped into a dummy node and if they do, it probably should be treated as a misbehavior of `DBGSuccinct`? Note that it never seemed to actually be a dummy node in `LabeledAlignerTest` test group.

metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -378,34 +397,35 @@ void build_pred_succ(const std::string &graph_fname,
...
+                    BOSS::edge_index next = boss.fwd(boss_idx);
+                    assert(next);
+                if ((*dummy)[next]) {
+                    find_succ = false;
+                        }
+            }
+            auto with_rd_succ = [&](bit_vector const& rd_succ) {
```
adamant-pwn, line 410: This seems to fix `RowDiff` test group. But generally I'm not sure if it's really OK for `route_at_forks` to just return an empty bit vector... Should I change it to return optional to indicate that it sometimes fails?

metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -378,34 +397,39 @@ void build_pred_succ(const std::string &graph_fname,
...
-                    while (rd_succ.size() && !rd_succ[next]) {
-                        next--;
-                        assert(!boss.get_last(next));
-                    }
-                    succ_buf.push_back(to_row(graph.boss_to_kmer_index(next)));
+            bool skip_succ = false, skip_all = false;
+            if (succinct) { // Legacy code for DBGSuccinct
+                BOSS::edge_index boss_idx = from_graph_index(graph, i);
+                if((*dummy)[boss_idx]) {
+                    skip_all = true;
```
adamant-pwn, line 404: I added this check, and also a check below that a predecessor is not dummy to fix integration tests. But still, I feel like it shouldn't ever happen, unless I misunderstand the concept of BOSS dummy edges.

metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -378,34 +397,39 @@ void build_pred_succ(const std::string &graph_fname,
...
+                } else {
+                    skip_succ = (*dummy)[succinct->get_boss().fwd(boss_idx)];
```
adamant-pwn, line 406: I'm trying to understand what is the intention of this check in the context of `row_diff_successor` function. Generally, do we want `row_diff_successor` to be well-defined for all nodes of the graph and return a graph node? If yes, then this convention would be violated for terminal nodes. How do we want to handle it? Should `row_diff_successor` return an `std::optional` on nodes that don't have a successor? Which nodes should it be in the terms of `graph::DeBruijnGraph`, if any?

metagraph/src/annotation/row_diff_builder.cpp
hmusta: `get_last()` is only used in the row diff builder and DBGSuccinct's implementation, so I don't think that warrants making it part of `DeBruijnGraph`.

One way around all of this would be to change the functionality of the dummy mask so that it keeps the indices the same, but it's simply used to avoid traversing to dummy edges and to keep a count of the number of real nodes.

hmusta: It'll mean that node indices in DBGSuccinct are not contiguous, but I think that's the lesser of two evils

adamant-pwn:
I hopefully fixed all unit tests. It seems that some integration tests still fail, so I'll look into it further over the next days. In the meantime I think it'd be nice to try to get it reviewed and merged, as the change becomes large and difficult to manage within one PR.

metagraph/src/annotation/row_diff_builder.cpp
adamant-pwn: Or maybe make `route_at_forks` return `shared_ptr` too, which will be `rd_succ` when possible and `get_last()` otherwise. Should simplify some branching in the code.

metagraph/integration_tests/test_transform_anno.py
```diff
@@ -104,9 +108,11 @@ def _check_aggregation_min_max_value(self, min_count, max_value, expected_densit
...
         out = res.stdout.decode().split('\n')[2:]
         self.assertEqual('labels:  1', out[0])
-        self.assertEqual('objects: 46960', out[1])
-        self.assertEqual(f'density: {expected_density}', out[2])
-        self.assertEqual(f'representation: {self.anno_repr}', out[3])
+        split = out[2].split()
+        self.assertEqual(out[2], split[0] + ' ' + split[1])
+        density = float(split[1])
+        self.assertIn(out[1], ['objects: 47633', 'objects: 46960'])
+        self.assertLess(abs(density - expected_density * 46960 / 47633), 10**-6)
```
adamant-pwn, line 115: As far as I can tell, all failing integration tests were seemingly failing due to `annotations.num_objects()` now returning the number of BOSS edges, including dummy ones, and `print_annotation_stats` doesn't seem to contain any information about the initial graph to use `graph.num_nodes()` for the computations instead. Is it okay to update integration tests like this, or should I propagate `num_nodes()` into annotations somehow when it's different from `num_objects()`?

hmusta: You can include both a graph and an annotation into a `metagraph stat` run and it'll print both. So I think the best thing to do is to update the `stat` print-out for a graph to also print `max_index()`, then you can check if that matches the annotation's `num_objects()`

adamant-pwn: Okay, I refactored _get_stats to return a dict, and also added checks based on max_index and its scale related to num_nodes into integration tests. Hope they'll pass now :smiling_face_with_tear:

metagraph/tests/annotation/test_converters.cpp
```diff
@@ -433,14 +444,16 @@ void test_row_diff_separate_columns(uint32_t k,
...
     std::map<std::string, std::vector<uint64_t>> col_annotations;
-    for (uint32_t anno_idx = 0; anno_idx < graph->num_nodes(); ++anno_idx) {
+    uint32_t anno_idx = 0;
+    graph->call_nodes([&](auto node_idx) {
         for (const auto &label : annotations[anno_idx]) {
-            col_annotations[label].push_back(anno_idx);
+            col_annotations[label].push_back(node_idx - 1);
         }
-    }
+        ++anno_idx;
```
hmusta, line 452: do we still need this? why don't we just set `auto anno_idx = AnnotatedDBG::graph_to_anno_index(node_idx)` in the callback and work with that?

metagraph/tests/annotation/test_converters.cpp
```diff
@@ -382,13 +386,15 @@ void test_row_diff(uint32_t k,
...
 
-    ColumnCompressed initial_annotation(graph->num_nodes());
+    ColumnCompressed initial_annotation(graph->max_index());
     std::unordered_set<std::string> all_labels;
-    for (uint32_t anno_idx = 0; anno_idx < graph->num_nodes(); ++anno_idx) {
+    uint32_t anno_idx = 0;
+    graph->call_nodes([&](uint32_t node_idx) {
         const std::vector<std::string> &labels = annotations[anno_idx];
-        initial_annotation.add_labels({anno_idx}, labels);
+        initial_annotation.add_labels({node_idx - 1}, labels);
```
hmusta, line 394: see below

metagraph/tests/annotation/test_converters.cpp
```diff
@@ -210,8 +210,8 @@ TEST(RowDiff, succ) {
 
         sdsl::int_vector_buffer succ(succ_file, std::ios::in);
         ASSERT_EQ(expected_succ.size(), succ.size());
-        for (uint32_t i = 0; i < succ.size(); ++i) {
-            EXPECT_EQ(expected_succ[i], succ[i]) << max_depth << " " << i;
+        for (uint32_t i = 0; i < expected_succ.size(); ++i) {
+            EXPECT_EQ(expected_succ[i] + 1, graph->rank_node(succ[i] + 1)) << max_depth << " " << i;
```
hmusta, line 214: how about we replace the `+1`s with `AnnotatedDBG::anno_to_graph_index` so it's more clear what's going on here?

metagraph/tests/annotation/test_converters.cpp
```diff
@@ -223,7 +223,7 @@ TEST(RowDiff, succ) {
         sdsl::int_vector_buffer pred(pred_file, std::ios::in);
         EXPECT_EQ(expected_pred.size(), pred.size());
         for (uint32_t i = 0; i < pred.size(); ++i) {
-            EXPECT_EQ(expected_pred[i], pred[i]) << max_depth << " " << i;
+            EXPECT_EQ(expected_pred[i] + 1, graph->rank_node(pred[i] + 1)) << max_depth << " " << i;
```
hmusta, line 226: see above

metagraph/tests/annotation/row_diff/test_row_diff.cpp
```diff
@@ -187,47 +187,59 @@ TEST(RowDiff, GetAnnotationMasked) {
...
+        { 0, 0, 0, 0, 1, 0, 1, 1 }
+    };
+    sdsl::bit_vector cols_concrete[2];
+    cols_concrete[0].resize(graph.max_index() + 1);
+    cols_concrete[1].resize(graph.max_index() + 1);
+    graph.call_nodes([&](auto i) {
+        auto rank = graph.rank_node(i) - 1;
+        bterminal[i - 1] = bterminal_masked[rank];
+        cols_concrete[0][i - 1] = cols_masked[0][rank];
+        cols_concrete[1][i - 1] = cols_masked[1][rank];
```
hmusta, line 203: should we replace these `-1`s with `AnnotatedDBG::graph_to_anno_index`?

metagraph/tests/annotation/row_diff/test_row_diff.cpp
```diff
@@ -187,47 +187,59 @@ TEST(RowDiff, GetAnnotationMasked) {
...
 
     ColumnMajor mat(std::move(cols));
 
     RowDiff<ColumnMajor> annot(&graph, std::move(mat));
     annot.load_anchor(fterm_temp.name());
 
-    EXPECT_EQ("CTAG", graph.get_node_sequence(1));
-    ASSERT_THAT(annot.get_rows({0})[0], ElementsAre(0, 1));
+    EXPECT_EQ("CTAG", graph.get_node_sequence(graph.select_node(1)));
+    ASSERT_THAT(annot.get_rows({graph.select_node(1) - 1})[0], ElementsAre(0, 1));
```
hmusta, line 221: see above re `graph_to_anno_index`

metagraph/tests/annotation/row_diff/test_row_diff.cpp
```diff
@@ -299,57 +311,68 @@ TEST(RowDiff, GetAnnotationBifurcationMasked) {
...
-
     ColumnMajor mat(std::move(cols));
 
     RowDiff<ColumnMajor> annot(&graph, std::move(mat));
     annot.load_anchor(fterm_temp.name());
 
-    EXPECT_EQ("CTAG", graph.get_node_sequence(1));
-    ASSERT_THAT(annot.get_rows({0})[0], ElementsAre(0, 1));
+    EXPECT_EQ("CTAG", graph.get_node_sequence(graph.select_node(1)));
+    ASSERT_THAT(annot.get_rows({graph.select_node(1) - 1})[0], ElementsAre(0, 1));
```
hmusta, line 348: see above re. `graph_to_anno_index`

metagraph/tests/annotation/row_diff/test_row_diff.cpp
```diff
@@ -299,57 +311,68 @@ TEST(RowDiff, GetAnnotationBifurcationMasked) {
...
+        {0, 1, 1, 0, 0, 0, 0, 0, 1, 0 }
+    };
+    sdsl::bit_vector cols_concrete[2];
+    cols_concrete[0].resize(graph.max_index() + 1);
+    cols_concrete[1].resize(graph.max_index() + 1);
+    graph.call_nodes([&](auto i) {
+        auto rank = graph.rank_node(i) - 1;
+        bterminal[i - 1] = bterminal_masked[rank];
+        cols_concrete[0][i - 1] = cols_masked[0][rank];
+        cols_concrete[1][i - 1] = cols_masked[1][rank];
```
hmusta, line 327: see above re. `graph_to_anno_index`

metagraph/src/annotation/binary_matrix/row_diff/row_diff.cpp
```diff
@@ -60,25 +60,26 @@ IRowDiff::get_rd_ids(const std::vector<BinaryMatrix::Row> &row_ids) const {
...
-            // If a node had been reached before, we interrupt the diff path.
-            // The annotation for that node will have been reconstructed earlier
-            // than for other nodes in this path as well. Thus, we will start
-            // reconstruction from that node and don't need its successors.
-            if (!is_new)
-                break;
-
-            if (anchor_[row])
-                break;
+            if (graph_->is_valid(boss_edge)) {
```
hmusta, line 67: `BOSS::row_diff_successor` is undefined for dummy edges, so I think we should make this an assert instead

metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -326,7 +346,72 @@ rd_succ_bv_type route_at_forks(const graph::DBGSuccinct &graph,
...
-void build_pred_succ(const std::string &graph_fname,
+node_index row_diff_successor(const graph::DeBruijnGraph &graph,
+                              node_index node,
+                              const bit_vector &rd_succ) {
+    if (auto* dbg_succ = dynamic_cast<graph::DBGSuccinct const*>(&graph)) {
+        return dbg_succ->get_boss().row_diff_successor(node, rd_succ);
+    } else {
+        node_index succ = graph::DeBruijnGraph::npos;
+        graph.adjacent_outgoing_nodes(node, [&](node_index adjacent_node) {
+            if(rd_succ[adjacent_node]) {
```
hmusta, line 357:
```suggestion
            if (rd_succ[adjacent_node]) {
```

metagraph/tests/annotation/row_diff/test_row_diff.cpp
```diff
@@ -260,34 +272,34 @@ TEST(RowDiff, GetAnnotationBifurcation) {
     RowDiff<ColumnMajor> annot(&graph, std::move(mat));
     annot.load_anchor(fterm_temp.name());
 
-    EXPECT_EQ("CTAG", graph.get_node_sequence(4));
+    EXPECT_EQ("CTAG", graph.get_node_sequence(graph.select_node(4)));
     ASSERT_THAT(annot.get_rows({3})[0], ElementsAre(0, 1));
```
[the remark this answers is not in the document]
adamant-pwn: I don't think so? `select_node` simply maps `[1, num_node]` into `[1, max_index]`, so if e.g. graph has no dummy nodes, it'd just return the argument.

metagraph/tests/annotation/row_diff/test_row_diff.cpp
adamant-pwn: Done

metagraph/tests/annotation/row_diff/test_row_diff.cpp
adamant-pwn: Done

metagraph/tests/annotation/row_diff/test_row_diff.cpp
adamant-pwn: Done.

metagraph/tests/annotation/row_diff/test_row_diff.cpp
adamant-pwn: Done.

metagraph/src/annotation/row_diff_builder.cpp
adamant-pwn: Done.

metagraph/src/graph/representation/base/sequence_graph.cpp
adamant-pwn: Ack, might be best to get back to it after some initial version is landed.

metagraph/src/graph/representation/succinct/dbg_succinct.cpp
```diff
@@ -164,34 +164,43 @@ void DBGSuccinct::call_incoming_kmers(node_index node,
...
     );
 }
 
+void DBGSuccinct::call_nodes(const std::function<void(node_index)> &callback,
+                             const std::function<bool()> &terminate) const {
+    for (node_index i = 1; i <= max_index() && !terminate(); ++i) {
+        if (is_valid(i)) {
+            callback(i);
+        }
+    }
```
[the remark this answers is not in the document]
adamant-pwn: I generally dislike the idea of using exceptions for control flow. I implemented this approach for now, but I feel like there should be other ways to work with it. For example, encode termination condition with the return value (instead of `void`). But it seems like it will require another huge overhaul of the codebase...

On the same note, it seems that some functions use `terminate()` for this, and others `stop_early()` :pensive:

metagraph/tests/annotation/row_diff/test_row_diff.cpp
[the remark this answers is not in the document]
adamant-pwn: Oh, nevermind, I misunderstood the question. `select_node` only needed in masked mode.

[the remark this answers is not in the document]
adamant-pwn: Well, eyes, but it's only really needed for `masked` version, while without masks it's consecutive indexing, so `select_node` is excessive. Do you still want me to add it?

metagraph/src/graph/representation/base/sequence_graph.cpp
hmusta: This is now handled in #505

PR #488, opened 2024-06-19
Fix a bunch of compilation errors

adamant-pwn:
Pull request for CI and review.


metagraph/CMakeLists.txt
```diff
@@ -319,6 +319,8 @@ IF(APPLE)
     set(THREADS_PREFER_PTHREAD_FLAG ON)
 ENDIF()
 add_subdirectory(external-libraries/spdlog)
+target_compile_options(spdlog PRIVATE -Wno-dangling-reference)
+add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-Wno-dangling-reference>)
```
adamant-pwn, line 323: I'm not sure what's the best course of action here is.

Basically, on newer g++ version, spdlog throws the following warning:
```console
/home/adamant/metagraph/metagraph/external-libraries/spdlog/include/spdlog/fmt/bundled/core.h:2886:44: in ‘constexpr’ expansion of ‘fmt::v8::make_format_args<>(((const std::__cxx11::basic_string<char>&)args#0), ((const long unsigned int&)args#1), ((const std::__cxx11::basic_string<char>&)args#2))’
/home/adamant/metagraph/metagraph/external-libraries/spdlog/include/spdlog/fmt/bundled/core.h:1566:15: error: possibly dangling reference to a temporary [-Werror=dangling-reference]
1566 | const auto& arg = arg_mapper<Context>().map(val);
| ^~~
/home/adamant/metagraph/metagraph/external-libraries/spdlog/include/spdlog/fmt/bundled/core.h:1566:46: note: the temporary was destroyed at the end of the full expression ‘fmt::v8::detail::arg_mapper<fmt::v8::basic_format_context<fmt::v8::appender, char> >().fmt::v8::detail::arg_mapper<fmt::v8::basic_format_context<fmt::v8::appender, char> >::map<std::__cxx11::basic_string<char> >((* & val))’
1566 | const auto& arg = arg_mapper<Context>().map(val);
| ~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~
cc1plus: all warnings being treated as errors
```

Unfortunately it's due to code in a header file, meaning that it also affects the compilation of metagraph source files. I've also tried update spdlog version, but it produces some other kinds of errors starting at `v1.14.0`.

metagraph/src/graph/representation/bitmap/dbg_bitmap.hpp
```diff
@@ -90,6 +90,7 @@ class DBGBitmap : public DeBruijnGraph {
     bool operator==(const DBGBitmap &other) const { return equals(other, false); }
     bool operator!=(const DBGBitmap &other) const { return !(*this == other); }
 
+    using DeBruijnGraph::operator!=;
```
adamant-pwn, line 93: Otherwise, we get
```console

/home/adamant/metagraph/metagraph/src/graph/representation/base/sequence_graph.hpp:224:18: error: ‘virtual bool mtg::graph::DeBruijnGraph::operator!=(const mtg::graph::DeBruijnGraph&) const’ was hidden [-Werror=overloaded-virtual=]
  224 |     virtual bool operator!=(const DeBruijnGraph &other) const { return !operator==(other); }
      |                  ^~~~~~~~
/home/adamant/metagraph/metagraph/src/graph/representation/bitmap/dbg_bitmap.hpp:91:10: note:   by ‘bool mtg::graph::DBGBitmap::operator!=(const mtg::graph::DBGBitmap&) const’
   91 |     bool operator!=(const DBGBitmap &other) const { return !(*this == other); }
      |          ^~~~~~~~
cc1plus: all warnings being treated as errors
```

metagraph/CMakeLists.txt
```diff
@@ -319,6 +319,10 @@ IF(APPLE)
     set(THREADS_PREFER_PTHREAD_FLAG ON)
 ENDIF()
 add_subdirectory(external-libraries/spdlog)
+if (CMAKE_CXX_COMPILER_ID MATCHES "GNU")
+  target_compile_options(spdlog PRIVATE -Wno-dangling-reference)
+  add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-Wno-dangling-reference>)
+endif()
```
adamant-pwn, line 325: I'm not sure what's the best course of action here.

Basically, on newer g++ version, spdlog throws the following warning:
```console
/home/adamant/metagraph/metagraph/external-libraries/spdlog/include/spdlog/fmt/bundled/core.h:2886:44: in ‘constexpr’ expansion of ‘fmt::v8::make_format_args<>(((const std::__cxx11::basic_string<char>&)args#0), ((const long unsigned int&)args#1), ((const std::__cxx11::basic_string<char>&)args#2))’
/home/adamant/metagraph/metagraph/external-libraries/spdlog/include/spdlog/fmt/bundled/core.h:1566:15: error: possibly dangling reference to a temporary [-Werror=dangling-reference]
1566 | const auto& arg = arg_mapper<Context>().map(val);
| ^~~
/home/adamant/metagraph/metagraph/external-libraries/spdlog/include/spdlog/fmt/bundled/core.h:1566:46: note: the temporary was destroyed at the end of the full expression ‘fmt::v8::detail::arg_mapper<fmt::v8::basic_format_context<fmt::v8::appender, char> >().fmt::v8::detail::arg_mapper<fmt::v8::basic_format_context<fmt::v8::appender, char> >::map<std::__cxx11::basic_string<char> >((* & val))’
1566 | const auto& arg = arg_mapper<Context>().map(val);
| ~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~
cc1plus: all warnings being treated as errors
```

Unfortunately it's due to code in a header file, meaning that it also affects the compilation of metagraph source files. I've also tried update spdlog version, but it produces some other kinds of errors starting at `v1.14.0`.

adamant-pwn:
Note: We also need to merge https://github.com/karasikov/caches/pull/6 as a part of this change, and update the submodule. But generally, do we still need to diverge from the upstream `cache`?

metagraph/src/graph/representation/bitmap/dbg_bitmap.hpp
hmusta: We rarely use this function, so I think the best thing would be to implement dummy versions of operator== and operator!= (similar to #481) as the defaults, then override them if necessary. I don't remember why this is implemented this way with `using`.

metagraph/CMakeLists.txt
hmusta: are the errors produced in our code, or theirs? also, is there an open issue for it?

metagraph/src/graph/representation/bitmap/dbg_bitmap.hpp
adamant-pwn: There are already default operators in `DebruijnGraph`. I needed to add `using` here because without it, `virtual bool operator!=(const DeBruijnGraph &other) const` is hidden by `bool operator!=(const DBGBitmap&)`. I think implementing it via `using` should work essentially the same as you suggest, it's just here to make it explicit that default comparison should be used for other graphs instead of `operator!=(const DBGBitmap&)`.

hmusta: ok, then it's not a problem

metagraph/CMakeLists.txt
adamant-pwn: Ah, nevermind, I actually managed to fix this by updating to `v1.14.1`. I just needed to also add `#include <spdlog/fmt/std.h>` in `common/loger.hpp`.

PR #497, opened 2024-07-25
Sshash module superkmer anno

adamant-pwn:
Pull request to track merging of further changes from @mmarzett.


adamant-pwn:
Also see https://github.com/ratschlab/sshash/pull/10 for related changes in sshash repo.

hmusta:
As mentioned during a previous meeting, I think annotating based on monotigs (monochromatic contigs) makes more sense. I've started an implementation at #500. Are there any fixes from here that we should port over there?

adamant-pwn:
I'm not very familiar with this branch, and my fixes mostly were to properly merge changes from #481 into it and make sure it still compiles and passes current tests. Is #500 implemented fully from scratch without relying on the content of this branch? If so, we can drop this branch, I suppose.

PR #500, opened 2024-10-02
Option to annotate a representative node from each monochromatic contig

hmusta:
Currently this is only implemented for SSHash graphs.

This change stores a bit vector that indicates the breakpoints for monochromatic unitigs in the SSHash graph. This allows us to maintain a lossless annotation while only annotating these breakpoints.

### DIFF
--- a/metagraph/src/graph/representation/hash/dbg_sshash.cpp
+++ b/metagraph/src/graph/representation/hash/dbg_sshash.cpp
@@ -103,11 +103,15 @@ DBGSSHash::DBGSSHash(const std::string &input_filename,
         common::logger->trace("Marking colour boundaries");
         sdsl::bit_vector id(num_nodes_);
         {
-            size_t pos = 0;
             ProgressBar progress_bar(num_nodes(),
                                      "Parsing sequences",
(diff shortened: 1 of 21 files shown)


metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -89,6 +98,81 @@ DBGSSHash::DBGSSHash(const std::string &input_filename, size_t k, Mode mode, siz
...
+
+    if (is_monochromatic) {
+        common::logger->trace("Marking colour boundaries");
+        sdsl::bit_vector id(num_nodes_);
+        {
+            ProgressBar progress_bar(num_nodes(),
+                                     "Parsing sequences",
+                                     std::cerr, !common::get_verbose());
+            seq_io::read_fasta_file_critical(input_filename, [&](auto *read_stream) {
+                auto indices = mtg::graph::map_to_nodes_sequentially(*this, std::string(read_stream->seq.s, read_stream->seq.l));
```
hmusta, line 110: TODO: make this multithreaded

PR #502, opened 2024-10-08
Fix for queries with invalid characters

hmusta:
Changes:
- Pre-encode query sequences to discard invalid characters
- Add `AnnotatedDBG` and `LabeledAligner` tests with DBGSSHash to test annotated graph queries

### DIFF
--- a/metagraph/src/graph/representation/canonical_dbg.cpp
+++ b/metagraph/src/graph/representation/canonical_dbg.cpp
@@ -180,7 +180,7 @@ void CanonicalDBG::call_outgoing_kmers(node_index node,
     }
 
     if (const auto sshash = std::dynamic_pointer_cast<const DBGSSHash>(graph_)) {
-        sshash->call_outgoing_kmers_with_rc<>(node, [&](node_index next, char c, bool orientation) {
+        sshash->call_outgoing_kmers_with_rc<true>(node, [&](node_index next, char c, bool orientation) {
             callback(orientation ? reverse_complement(next) : next, c);
         });
         return;
@@ -273,7 +273,7 @@ void CanonicalDBG::call_incoming_kmers(node_index node,
     }
 
     if (const auto sshash = std::dynamic_pointer_cast<const DBGSSHash>(graph_)) {
-        sshash->call_incoming_kmers_with_rc<>(node, [&](node_index prev, char c, bool orientation) {
+        sshash->call_incoming_kmers_with_rc<true>(node, [&](node_index prev, char c, bool orientation) {
             callback(orientation ? reverse_complement(prev) : prev, c);
         });
         return;
--- a/metagraph/src/graph/representation/canonical_dbg.cpp
+++ b/metagraph/src/graph/representation/canonical_dbg.cpp
@@ -64,7 +64,7 @@ void CanonicalDBG
     path.reserve(sequence.size() - get_k() + 1);
 
     if (const auto sshash = std::dynamic_pointer_cast<const DBGSSHash>(graph_)) {
-        sshash->map_to_nodes_with_rc<>(sequence, [&](node_index node, bool orientation) {
+        sshash->map_to_nodes_with_rc<true>(sequence, [&](node_index node, bool orientation) {
             callback(node && orientation ? reverse_complement(node) : node);
         }, terminate);
         return;
--- a/metagraph/src/graph/representation/hash/dbg_sshash.cpp
+++ b/metagraph/src/graph/representation/hash/dbg_sshash.cpp
@@ -100,33 +100,33 @@ void DBGSSHash::add_sequence(std::string_view sequence,
     throw std::logic_error("adding sequences not supported");
 }
 
-void DBGSSHash
-::map_to_nodes_with_rc_advanced(std::string_view sequence,
-                                const std::function<void(sshash::lookup_result)>& callback,
-                                bool with_rc,
-                                const std::function<bool()>& terminate) const {
+template <class Dict, bool with_rc>
+void map_to_nodes_with_rc_advanced(size_t k,
+                                   const Dict &dict,
+                                   std::string_view sequence,
+                                   const std::function<void(sshash::lookup_result)>& callback,
+                                   const std::function<bool()>& terminate) {
     size_t n = sequence.size();
-    if (terminate() || n < k_)
+    if (terminate() || n < k)
         return;
 
-    std::visit([&](const auto &dict) {
-        using kmer_t = get_kmer_t<decltype(dict)>;
+    using kmer_t = get_kmer_t<Dict>;
 
-        std::vector<bool> invalid_char(n);
-        for (size_t i = 0; i < n; ++i) {
-            invalid_char[i] = !kmer_t::is_valid(sequence[i]);
-        }
-        auto invalid_kmer = utils::drag_and_mark_segments(invalid_char, true, k_);
+    std::vector<char> seq_encoded(n);
+    for (size_t i = 0; i < n; ++i) {
+        seq_encoded[i] = !kmer_t::is_valid(sequence[i]);
+    }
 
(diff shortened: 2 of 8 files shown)


metagraph/src/graph/representation/canonical_dbg.cpp
```diff
@@ -64,7 +64,7 @@ ::map_to_nodes_sequentially(std::string_view sequence,
     path.reserve(sequence.size() - get_k() + 1);
 
     if (const auto sshash = std::dynamic_pointer_cast<const DBGSSHash>(graph_)) {
-        sshash->map_to_nodes_with_rc<>(sequence, [&](node_index node, bool orientation) {
+        sshash->map_to_nodes_with_rc<true>(sequence, [&](node_index node, bool orientation) {
```
hmusta, line 67: This was the default value for the template argument, I just made it explicit here in case the default changes in the future.

metagraph/tests/graph/all/test_dbg_helpers.cpp
```diff
@@ -155,7 +160,7 @@ build_graph<DBGSSHash>(uint64_t k,
         return std::make_shared<DBGSSHash>(k, mode);
 
     // use DBGHashString to get contigs for SSHash
```
adamant-pwn, line 162:
```suggestion
    // use DBGHashFast to get contigs for SSHash
```

metagraph/tests/graph/all/test_dbg_helpers.cpp
```diff
@@ -1,6 +1,10 @@
 #include "test_dbg_helpers.hpp"
 
+#include "../../annotation/test_annotated_dbg_helpers.hpp"
+#include "annotation/representation/column_compressed/annotate_column_compressed.hpp"
+
 #include "gtest/gtest.h"
+#include "graph/annotated_dbg.hpp"
```
adamant-pwn, line 7: Why? The PR doesn't seem to add anything that needs them in this file?

adamant-pwn (changes requested):

metagraph/tests/graph/all/test_dbg_helpers.cpp
hmusta: Thanks, this made it here accidentally, it should be part of #500 instead

metagraph/src/graph/representation/hash/dbg_sshash.hpp
```diff
@@ -117,6 +117,12 @@ class DBGSSHash : public DeBruijnGraph {
     size_t num_nodes_;
     Mode mode_;
 
+    void map_to_nodes_with_rc_advanced(
```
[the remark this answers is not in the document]
hmusta: We can remove it from the interface. I'd rather keep this as a separate function in the implementation because it might be useful later for things like mapping sequences to contigs.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -99,6 +100,36 @@ void DBGSSHash::add_sequence(std::string_view sequence,
...
+
+        auto invalid = utils::drag_and_mark_segments(seq_encoded, 1, k_);
+
+        kmer_t uint_kmer = sshash::util::string_to_uint_kmer<kmer_t>(sequence.data(), k_ - 1);
+        uint_kmer.pad_char();
+        for (size_t i = k_ - 1; i < sequence.size() && !terminate(); ++i) {
+            uint_kmer.drop_char();
+            uint_kmer.kth_char_or(k_ - 1, kmer_t::char_to_uint(sequence[i]));
+            callback(!invalid[i] ? dict.lookup_advanced_uint(uint_kmer, with_rc)
+                                 : sshash::lookup_result());
```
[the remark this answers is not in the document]
hmusta: `utils::drag_and_mark_segments` doesn't work with `std::vector<bool>` since we don't have direct access to its underlying storage, so I revised this slightly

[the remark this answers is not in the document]
adamant-pwn: Do we actually need to have access to its internal storage for `drag_and_mark_segments`? Maybe a better thing to do would be to change https://github.com/ratschlab/metagraph/blob/f2f320fa3fd3c9472260abc2abe0857618d8181e/metagraph/src/common/algorithms.hpp#L56-L57

into

```cpp
 = std::find(array.begin(), array.end(), label) 
	     - array.begin();
```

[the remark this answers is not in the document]
hmusta: good point, I actually don't think this needs to use `data`, so we can just use iterators

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -99,6 +100,35 @@ void DBGSSHash::add_sequence(std::string_view sequence,
...
 
+template <class Dict, bool with_rc>
+void map_to_nodes_with_rc_advanced(size_t k,
+                                   const Dict &dict,
+                                   std::string_view sequence,
+                                   const std::function<void(sshash::lookup_result)>& callback,
+                                   const std::function<bool()>& terminate) {
+    size_t n = sequence.size();
+    if (terminate() || n < k)
+        return;
```
adamant-pwn, line 111: If we have this check here, it's probably fine to remove it from `map_to_nodes_with_rc`. And we  should probably also move `if (!num_nodes())` check from `map_to_nodes_with_rc` here as well.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -99,6 +100,35 @@ void DBGSSHash::add_sequence(std::string_view sequence,
     throw std::logic_error("adding sequences not supported");
 }
 
+template <class Dict, bool with_rc>
+void map_to_nodes_with_rc_advanced(size_t k,
```
adamant-pwn, line 104: Possibly rename to `map_to_nodes_with_rc_internal` or `map_to_nodes_with_rc_impl`.

adamant-pwn: Ok, let's do `impl`. I think it better conveys our intention for why this function exists.

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -99,32 +100,53 @@ void DBGSSHash::add_sequence(std::string_view sequence,
     throw std::logic_error("adding sequences not supported");
 }
 
-template <bool with_rc>
-void DBGSSHash::map_to_nodes_with_rc(std::string_view sequence,
-                                     const std::function<void(node_index, bool)>& callback,
-                                     const std::function<bool()>& terminate) const {
-    if (terminate() || sequence.size() < k_)
+template <class Dict, bool with_rc>
```
adamant-pwn, line 103: Maybe make `Dict` the second parameter and let it be auto-deduced?

metagraph/src/graph/representation/hash/dbg_sshash.cpp
```diff
@@ -99,32 +100,53 @@ void DBGSSHash::add_sequence(std::string_view sequence,
...
     }
 
+    using kmer_t = get_kmer_t<Dict>;
+
+    std::vector<bool> seq_encoded(n);
+    for (size_t i = 0; i < n; ++i) {
+        seq_encoded[i] = !kmer_t::is_valid(sequence[i]);
+    }
+
+    auto invalid_kmer = utils::drag_and_mark_segments(seq_encoded, 1, k);
```
adamant-pwn, line 127:
```suggestion
    std::vector<bool> invalid_char(n);
    for (size_t i = 0; i < n; ++i) {
        invalid_char[i] = !kmer_t::is_valid(sequence[i]);
    }

    auto invalid_kmer = utils::drag_and_mark_segments(invalid_char, true, k);
```

adamant-pwn (approved):
Looks good to me!

PR #505, opened 2024-10-15
Use RowDiff with DBGHashFast in test_aligner_labeled

adamant-pwn:
Also uses `in_graph(node)` checks more extensively throughout the code base and serializes  `get_last(graph)` instead of `rd_succ` for non-BOSS graphs.

### DIFF
--- a/metagraph/integration_tests/base.py
+++ b/metagraph/integration_tests/base.py
@@ -37,10 +37,19 @@ class TestingBase(unittest.TestCase):
     def _get_stats(graph_path):
         stats_command = METAGRAPH + ' stats ' + graph_path + ' --mmap'
         res = subprocess.run(stats_command.split(), stdout=PIPE, stderr=PIPE)
-        assert(res.returncode == 0)
+        if res.returncode != 0:
+            raise AssertionError(f"Command '{stats_command}' failed with return code {res.returncode} and error: {res.stderr.decode()}")
         stats_command = METAGRAPH + ' stats ' + graph_path + MMAP_FLAG
         res = subprocess.run(stats_command.split(), stdout=PIPE, stderr=PIPE)
-        return res
+        parsed = dict()
+        parsed['returncode'] = res.returncode
+        res = res.stdout.decode().split('\n')[2:]
+        for line in res:
+            if ': ' in line:
+                x, y = map(str.strip, line.split(':', 1))
+                assert(x not in parsed or parsed[x] == y)
+                parsed[x] = y
+        return parsed
 
     @staticmethod
     def _build_graph(input, output, k, repr, mode='basic', extra_params=''):
--- a/metagraph/integration_tests/base.py
+++ b/metagraph/integration_tests/base.py
@@ -6,7 +6,12 @@ from tempfile import TemporaryDirectory
 
 script_path = os.path.dirname(os.path.realpath(__file__))
 
-METAGRAPH = f'{os.getcwd()}/metagraph'
+USE_GDB = False
+METAGRAPH_EXE = f'{os.getcwd()}/metagraph'
+GDB_PREFIX = 'gdb -ex run -ex quit --args ' if USE_GDB else ''
+METAGRAPH = GDB_PREFIX + METAGRAPH_EXE
+DNA_MODE = os.readlink(METAGRAPH_EXE).endswith("_DNA")
+PROTEIN_MODE = os.readlink(METAGRAPH_EXE).endswith("_Protein")
 
 TEST_DATA_DIR = os.path.join(script_path, '..', 'tests', 'data')
 
--- a/metagraph/integration_tests/base.py
+++ b/metagraph/integration_tests/base.py
@@ -6,12 +6,14 @@ from tempfile import TemporaryDirectory
 
 script_path = os.path.dirname(os.path.realpath(__file__))
 
-USE_GDB = False
 METAGRAPH_EXE = f'{os.getcwd()}/metagraph'
-GDB_PREFIX = 'gdb -ex run -ex quit --args ' if USE_GDB else ''
-METAGRAPH = GDB_PREFIX + METAGRAPH_EXE
 DNA_MODE = os.readlink(METAGRAPH_EXE).endswith("_DNA")
 PROTEIN_MODE = os.readlink(METAGRAPH_EXE).endswith("_Protein")
+METAGRAPH = METAGRAPH_EXE
+
+def update_prefix(PREFIX):
+    global METAGRAPH
+    METAGRAPH = PREFIX + METAGRAPH_EXE
 
 TEST_DATA_DIR = os.path.join(script_path, '..', 'tests', 'data')
 
--- a/metagraph/integration_tests/main.py
+++ b/metagraph/integration_tests/main.py
@@ -5,6 +5,7 @@ import fnmatch
 import sys
 import argparse
 from helpers import TimeLoggingTestResult
+from base import update_prefix
 
 
 """Run all integration tests"""
@@ -31,8 +32,12 @@ if __name__ == '__main__':
     try:
         parser = argparse.ArgumentParser(description='Metagraph integration tests.')
         parser.add_argument('--test_filter', dest='filter', type=str, default="*",
-                            help='filter test cases (default: run all)')
+                    help='filter test cases (default: run all)')
+        parser.add_argument('--gdb', dest='use_gdb', action='store_true',
+                    help='run metagraph with gdb')
         args = parser.parse_args()
+        if args.use_gdb:
+            update_prefix('gdb -ex run -ex bt -ex quit --args ')
 
         result = unittest.TextTestRunner(
             resultclass=TimeLoggingTestResult
--- a/metagraph/integration_tests/main.py
+++ b/metagraph/integration_tests/main.py
@@ -32,9 +32,9 @@ if __name__ == '__main__':
     try:
         parser = argparse.ArgumentParser(description='Metagraph integration tests.')
         parser.add_argument('--test_filter', dest='filter', type=str, default="*",
-                    help='filter test cases (default: run all)')
+                            help='filter test cases (default: run all)')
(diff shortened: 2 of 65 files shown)


metagraph/integration_tests/base.py
```diff
@@ -6,7 +6,12 @@
 
 script_path = os.path.dirname(os.path.realpath(__file__))
 
-METAGRAPH = f'{os.getcwd()}/metagraph'
+USE_GDB = False
```
adamant-pwn, line 9: What would be a consistent way to read it from params?

hmusta: You can add a flag here:
https://github.com/ratschlab/metagraph/blob/7535d81511478109a7857c0e93f133948fe3ed00/metagraph/integration_tests/main.py#L33

metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -428,14 +418,14 @@ void build_pred_succ(const graph::DeBruijnGraph &graph,
                   outfbase);
 
     std::optional<sdsl::bit_vector> dummy;
-    auto* succinct = dynamic_cast<graph::DBGSuccinct const*>(&graph);
-    if (succinct) {
+    if (auto* succinct = dynamic_cast<graph::DBGSuccinct const*>(&graph)) {
         dummy = succinct->get_boss().mark_all_dummy_edges(num_threads);
```
adamant-pwn, line 422: Wouldn't it be better to do `succinct->mask_dummy_kmers(num_threads, false)` here or somewhere on the way to this call? Then we can get rid of annoying dummy checks below.

metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -331,38 +333,26 @@ rd_succ_bv_type route_at_forks(const graph::DeBruijnGraph &graph,
...
+ 
     } else {
         logger->warn("No count vectors could be found in {}. The last outgoing"
                      " edges will be selected for assigning RowDiff successors",
                      count_vectors_dir);
+        rd_succ = last;
+        if(dynamic_cast<graph::DBGSuccinct const*>(&graph)) {
+            rd_succ_bv_type().serialize(f);
+        } else {
+            rd_succ_bv_type(*rd_succ).serialize(f);
```
adamant-pwn, line 347: Not sure what's the best way to avoid copy-construction here, and if it's generally possible for typical results of get_last.

metagraph/src/annotation/row_diff_builder.cpp
adamant-pwn: Well, here goes nothing. I've tried to add it in some way. I guess, should be further discussed in the relevant section of the review.

metagraph/src/annotation/binary_matrix/row_diff/row_diff.cpp
```diff
@@ -54,34 +75,36 @@ IRowDiff::get_rd_ids(const std::vector<BinaryMatrix::Row> &row_ids) const {
...
+
+            if (anchor_[row])
+                break;
+
+            if (fork_succ_.size()) {
+                node = row_diff_successor(*graph_, node, fork_succ_);
+            } else { // Only happens in DBGSuccinct
+                auto *dbg_succ = dynamic_cast<graph::DBGSuccinct const*>(graph_);
+                assert(dbg_succ);
+                node = row_diff_successor(*graph_, node, dbg_succ->get_boss().get_last());
```
adamant-pwn, line 105: Would it be problematic to, instead of doing this, assign `get_last()` to `fork_succ` during its load, if it was detected that it is empty? I assume in this case we would also need to change `fork_succ_` itself into a pointer to `bit_vector`?

metagraph/src/annotation/row_diff_builder.cpp
```diff
@@ -331,38 +335,26 @@ rd_succ_bv_type route_at_forks(const graph::DeBruijnGraph &graph,
...
-
+        rd_succ = std::make_shared<rd_succ_bv_type>(std::move(rd_succ_bv));
+        rd_succ->serialize(f);
+ 
     } else {
         logger->warn("No count vectors could be found in {}. The last outgoing"
                      " edges will be selected for assigning RowDiff successors",
                      count_vectors_dir);
+        rd_succ = last;
+        if(dynamic_cast<graph::DBGSuccinct const*>(&graph)) {
```
adamant-pwn, line 346: I really don't like doing this. If we try to serialize `last`, what exactly is the issue with conversion? How much of a slowdown would it be? Alternatively, would it be an option to make some general bit_vector load and add class tags to serialized versions?

metagraph/src/graph/representation/base/sequence_graph.cpp
```diff
@@ -447,7 +448,8 @@ ::call_kmers(const std::function<void(node_index,
 
             while (!visited[node]) {
                 visited[node] = true;
-                callback(node, sequence);
+                if (in_graph(node))
+                    callback(node, sequence);
```
adamant-pwn, line 452: Probably should be an assert, rather than `if`.

metagraph/src/graph/representation/hash/dbg_hash_fast.cpp
```diff
@@ -57,33 +57,33 @@ class DBGHashFastImpl : public DBGHashFast::DBGHashFastInterface {
     }
 
     void add_sequence(std::string_view sequence,
-                      const std::function<void(node_index)> &on_insertion);
+                      const std::function<void(node_index)> &on_insertion) override final;
```
adamant-pwn, line 60: Why did it compile without them? I couldn't find the answer online...

metagraph/src/graph/representation/succinct/dbg_succinct.cpp
```diff
@@ -770,24 +770,29 @@ void DBGSuccinct::serialize(const std::string &filename) const {
...
-    if (!out.good())
-        throw std::ios_base::failure("Can't write to file " + out_filename);
+        valid_edges->serialize(out);
+    };
 
-    valid_edges_->serialize(out);
+    if (valid_edges_) {
+        serialize_valid_edges(valid_edges_);
+    } else {
+        serialize_valid_edges(generate_dummy_kmers(1, false));
```
adamant-pwn, line 794: With the new indexing, I _really_ don't want to allow these graphs to walk around with invalid `in_graph` function. Would lead (and leads if this is removed) to all kinds of hard to debug behaviors. But I'm also not sure what's the best place to propagate this information, and how much overhead it may add?

metagraph/integration_tests/test_align.py
```diff
@@ -366,7 +363,7 @@ def test_simple_align_edit_distance_all_graphs(self, representation):
 
         params = self._get_stats(self.tempdir.name + '/genome.MT' + graph_file_extension[representation])
         self.assertEqual('11', params['k'])
-        self.assertEqual('16461', params['nodes (k)'])
+        self.assertEqual('16438', params['nodes (k)'])
```
adamant-pwn, line 366: This number decreases if `valid_edges` is generated and serialized for DBGSuccinct. Probably because dummy nodes are excluded this way. Same with the change above.

metagraph/tests/annotation/test_converters.cpp
```diff
@@ -299,9 +299,9 @@ TEST(RowDiff, ConvertFromColumnCompressedSameLabels) {
             EXPECT_EQ(labels.size() * expected_relations[max_depth - 1],
                       annotator.num_relations());
 
-            for (uint32 i = 0; i < annotator.num_objects(); ++i) {
-                ASSERT_THAT(annotator.get_labels(i), ContainerEq(labels));
-            }
+            graph->call_nodes([&](uint32_t node_idx) {
+                ASSERT_THAT(annotator.get_labels(node_idx - 1), ContainerEq(labels));
```
hmusta, line 303: replace `-1` with `AnnotatedDBG::graph_to_anno_index`

metagraph/tests/annotation/test_converters.cpp
```diff
@@ -353,9 +353,9 @@ TEST(RowDiff, ConvertFromColumnCompressedSameLabelsMultipleColumns) {
                 ASSERT_EQ(graph->max_index(), annotator.num_objects());
                 EXPECT_EQ(expected_relations[max_depth - 1], annotator.num_relations());
 
-                for (uint32 idx = 0; idx < annotator.num_objects(); ++idx) {
-                    ASSERT_THAT(annotator.get_labels(idx), ElementsAre(labels[i]));
-                }
+                graph->call_nodes([&](uint32_t node_idx) {
+                    ASSERT_THAT(annotator.get_labels(node_idx - 1), ElementsAre(labels[i]));
```
hmusta, line 357: replace `-1` with `AnnotatedDBG::graph_to_anno_index`

metagraph/integration_tests/main.py
```diff
@@ -31,8 +32,12 @@ def create_test_suite(filter_pattern="*"):
     try:
         parser = argparse.ArgumentParser(description='Metagraph integration tests.')
         parser.add_argument('--test_filter', dest='filter', type=str, default="*",
-                            help='filter test cases (default: run all)')
+                    help='filter test cases (default: run all)')
```
hmusta, line 35:
```suggestion
                            help='filter test cases (default: run all)')
```

hmusta (changes requested):
I haven't looked through everything yet, but here's my first batch of comments

metagraph/src/annotation/row_diff_builder.cpp
hmusta: I used a dynamic_cast to bit_vector_stat (with an assertion check) and std::move here to avoid the copy

metagraph/src/graph/representation/succinct/dbg_succinct.cpp
hmusta: I think we should save this for a later pull request, since I think it makes sense to first explore how much of a storage overhead this would introduce.

PR #518, opened 2025-05-12
use sparse bit vectors for rainbow matrices

hmusta:


karasikov:
Closing this, as RAM overhead has been dramatically reduced in #606 and it's no longer an issue.

PR #522, opened 2025-05-16
Batch aggregation (two-level batching)

karasikov:


hmusta (approved):
This is fine for now, but I'm gonna change it in #521 since it'll cause race conditions there

metagraph/src/annotation/binary_matrix/multi_brwt/clustering.cpp
```diff
@@ -16,7 +16,7 @@ namespace matrix {
 
 using mtg::common::logger;
 
-typedef std::vector<std::vector<uint64_t>> Partition;
+typedef std::vector<std::vector<uint32_t>> Partition;
```
hmusta, line 19: Should we just make this `Column` as well?

metagraph/src/annotation/binary_matrix/multi_brwt/clustering.cpp
```diff
@@ -258,7 +258,7 @@ Partition greedy_matching(const std::vector<T> &columns, size_t num_threads) {
         ++progress_bar;
     }
 
-    for (size_t i = 0; i < columns.size(); ++i) {
+    for (uint32_t i = 0; i < columns.size(); ++i) {
```
hmusta, line 261: same here

hmusta (changes requested):

metagraph/src/annotation/binary_matrix/multi_brwt/clustering.cpp
karasikov: this always needs to be 32, it's a specific to this .cpp file

karasikov: ah, no, I confused it with something else. Yeah, this can be either, RAM is not an issue here. But still specific to this file

PR #535, opened 2025-09-30
Updated Dockerfile

karasikov:
* ~~copy locally compiled htslib .so files to /usr/lib (since after [#445](https://github.com/ratschlab/metagraph/pull/445) metagraph is linked against a locally compiled htslib)~~ not needed after [#534](https://github.com/ratschlab/metagraph/pull/534). Now htslib is always linked statically.
* check if the binary runs fine
* removed unnecessary libs

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,39 +8,21 @@ FROM ubuntu:22.04 AS metagraph_dev_env
 RUN export DEBIAN_FRONTEND="noninteractive" && apt-get update && apt-get install -y \
     autoconf \
     automake \
-    binutils-dev \
     ccache \
-    curl \
     g++-11 \
     git \
     cmake \
-    libcurl4-gnutls-dev \
     libboost-all-dev \
-    libbrotli-dev \
     libbz2-dev \
     libdeflate-dev \
-    libdouble-conversion-dev \
-    libevent-dev \
-    libgflags-dev \
-    libgoogle-glog-dev \
-    libhts-dev \
-    libiberty-dev \
     libjemalloc-dev \
-    liblz4-dev \
-    liblzma-dev \
     libzstd-dev \
-    libsnappy-dev \
     libssl-dev \
-    libtool \
-    libunwind-dev \
     make \
     pkg-config \
     python3 \
     python3-pip \
     python3-venv \
-    vim \
-    wget \
-    zlib1g-dev \
     && rm -rf /var/lib/apt/lists/*
 
 WORKDIR /opt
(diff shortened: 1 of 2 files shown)


metagraph/CMakeLists.txt
```diff
@@ -376,13 +387,43 @@ add_library(sdust STATIC
...
+    foreach(a ${CMAKE_OSX_ARCHITECTURES})
+      list(APPEND f -arch ${a})
+    endforeach()
+    string(REPLACE ";" " " f "${f}")
+    set(${outvar} "${f}" PARENT_SCOPE)
+  else()
+    set(${outvar} "" PARENT_SCOPE)
+  endif()
+endfunction()
+make_osx_arch_flags(OSX_ARCH_FLAGS)
```
adamant-pwn, line 404: is it possible to add a CI that reproduces whatever was failing previously, that now requires us to do this?

adamant-pwn:
Why are there cancelled checks?

metagraph/CMakeLists.txt
```diff
@@ -376,13 +387,43 @@ add_library(sdust STATIC
...
-  CONFIGURE_COMMAND autoreconf -i ${CMAKE_CURRENT_SOURCE_DIR}/external-libraries/htslib && ${CMAKE_CURRENT_SOURCE_DIR}/external-libraries/htslib/configure --prefix=${CMAKE_BINARY_DIR}/external-libraries/htslib --disable-libcurl --disable-bz2 --disable-lzma
+  PREFIX "${CMAKE_BINARY_DIR}/HTSLIB-build"
+  CONFIGURE_COMMAND
+    ${CMAKE_COMMAND} -E env
+      "CPPFLAGS=${OSX_ARCH_FLAGS}\ -I${CMAKE_SOURCE_DIR}/external-libraries/zlib"
+      "CFLAGS=${OSX_ARCH_FLAGS}\ -I${CMAKE_SOURCE_DIR}/external-libraries/zlib"
+      "LDFLAGS=${OSX_ARCH_FLAGS}\ -L${CMAKE_BINARY_DIR}/external-libraries/zlib"
+    /bin/sh -c "autoreconf -i ${CMAKE_CURRENT_SOURCE_DIR}/external-libraries/htslib && exec ${CMAKE_CURRENT_SOURCE_DIR}/external-libraries/htslib/configure --disable-libcurl --disable-bz2 --disable-lzma"
+  BUILD_COMMAND $(MAKE) -j${N} lib-static
+  #INSTALL_COMMAND $(MAKE) install prefix=${HTSLIB_INSTALL}  # this installs both static and dynamic
```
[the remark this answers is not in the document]
adamant-pwn: But it's not the only library we link against dynamically? Are the others installed system-wide? Or is it the only one?

karasikov:
> Why are there cancelled checks?

Maybe a github's bug

metagraph/CMakeLists.txt
adamant-pwn, line 67: Did you notice this flag? I think we should treat htslib in accordance with the flag, rather than force it to link statically and remove .so files altogether. Maybe use `${HTSLIB_INSTALL}/lib/libhts.a` or `${HTSLIB_INSTALL}/lib/libhts.so` depending on the `BUILD_STATIC` flag in `METALIBS`?

karasikov: BUILD_STATIC says, build fully static binary. It's a different thing.
Here I want to avoid the dependency on libhts.so even for dynamic metagraph binaries

metagraph/CMakeLists.txt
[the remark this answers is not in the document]
adamant-pwn: Alright, let's keep this part like this.

metagraph/CMakeLists.txt
karasikov: implemented in #534

PR #538, opened 2025-10-01
Improvements in threading, parallel integration tests

karasikov:
- Improved nested parallelism in annotation construction (up to ~2x faster)

- Switched from using std::thread (ThreadPool) to omp tasks in query.cpp and annotate.cpp, where the tasks create subroutines using omp parallel (flaky with some implementations of omp, e.g., static build in Linux CI runners)

- Improvements in threading

- Parallel integration tests (use all cores by default)

- Don't rerun integration tests on failure -- ensure there are no flaky tests in the CI. Previously, a bunch of tests were flaky. Now all those are fixed, and only `metagraph annotate` sometimes fails in the CI for certain builds (could be an omp bug), but it never fails on real examples in practice, hence we allow reruns for that annotation command in the integration tests.

### DIFF
--- a/metagraph/integration_tests/main.py
+++ b/metagraph/integration_tests/main.py
@@ -29,9 +29,8 @@ def load_tests_from_module(module_name, filter_pattern="*"):
                 module_tests.append(test)
     return module_tests
 
-def create_test_suite(filter_pattern="*"):
+def create_test_suite(test_files, filter_pattern="*"):
     """Create a test suite with filtered tests"""
-    test_files = glob.glob(os.path.dirname(os.path.realpath(__file__)) + '/test_*.py')
     module_names = [os.path.basename(f)[:-3] for f in test_files]
     all_tests = []
     for module_name in module_names:
@@ -88,16 +87,11 @@ def run_test_parallel(test_info, filter_pattern="*"):
         'output': output_text
     }
 
-def run_tests_parallel(max_workers, filter_pattern="*"):
+def run_tests_parallel(max_workers, test_files, filter_pattern="*", chunk_size = 20):
     """Run test files in parallel with chunking"""
-    # Find all test files
-    test_files = glob.glob(os.path.dirname(os.path.realpath(__file__)) + '/test_*.py')
-
     # Create chunks from test files, respecting class boundaries
     # Note: test_api uses server ports, so don't chunk it to avoid port collisions
-    chunk_size = 20
     all_chunks = []
-
     for test_file in test_files:
         module_name = os.path.basename(test_file)[:-3]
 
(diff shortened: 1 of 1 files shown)


.github/workflows/macos.yml
```diff
@@ -209,15 +205,11 @@ jobs:
 
       - &run-integration-tests
         name: Run integration tests
-        uses: nick-fields/retry@v2
-        with:
-          timeout_minutes: 1000
-          max_attempts: ${{ env.INTEGRATION_TEST_ATTEMPTS }}
```
adamant-pwn: And this

metagraph/integration_tests/main.py
```diff
@@ -4,46 +4,266 @@
...
+        for suite in unittest.defaultTestLoader.loadTestsFromName(module_name):
+            for test in suite:
+                if matches_filter(test, filter_pattern):
+                    all_tests.append(test)
+
+        # Add tests (slice if chunk, all if file)
+        tests_to_run = all_tests[start_idx:end_idx] if is_chunk else all_tests
+        test_suite = unittest.TestSuite(tests_to_run)
+    except Exception as e:
+        return {'test': test_name, 'success': False, 'error': str(e), 'duration': 0}
```
adamant-pwn: Can we factor it out, to share with `create_test_suite`?

metagraph/integration_tests/main.py
```diff
@@ -4,46 +4,266 @@
...
+def run_tests_parallel(filter_pattern="*", max_workers=4):
+    """Run test files in parallel with chunking"""
+    # Find all test files
+    test_files = glob.glob(os.path.dirname(os.path.realpath(__file__)) + '/test_*.py')
+
+    if not test_files:
+        print("No test files found!")
+        return False
+
+    # Create chunks from test files, respecting class boundaries
```
adamant-pwn: How balanced are the chunks? AFAIK, 70% of all tests or so are query tests?

metagraph/integration_tests/main.py
```diff
@@ -4,46 +4,266 @@
...
+
+    return {
+        'test': test_name,
+        'success': success,
+        'tests_run': result.testsRun,
+        'duration': duration,
+        'output': output_text
+    }
+
+def run_tests_parallel(filter_pattern="*", max_workers=4):
```
adamant-pwn: Why default value for `max_workers`? We always specify it, right?

adamant-pwn (changes requested):

metagraph/integration_tests/main.py
```diff
@@ -4,46 +4,266 @@
...
+                        'start_idx': start_idx + i,
+                        'end_idx': min(start_idx + i + chunk_size, start_idx + count),
+                        'chunk_id': len(all_chunks)
+                    })
+
+    print(f"Running {len(all_chunks)} test chunks in parallel (num_workers={max_workers})")
+    print(f"Filter pattern: {filter_pattern}")
+    print("-" * 60)
+
+    # Set up signal handling for graceful shutdown
```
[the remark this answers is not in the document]
karasikov: yes, it runs in both

adamant-pwn (approved):
Looks good to me, modulo some minor comments / suggestions.

PR #540, opened 2025-10-03
support hosting multiple graphs in one process

karasikov:
* support hosting multiple graphs
* filter which ones to query with flag `graphs`

### DIFF
--- a/metagraph/src/cli/server.cpp
+++ b/metagraph/src/cli/server.cpp
@@ -1,4 +1,5 @@
 #include <json/json.h>
+#include <tsl/hopscotch_map.h>
 #include <server_http.hpp>
 
 #include "common/logger.hpp"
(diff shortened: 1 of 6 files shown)


metagraph/src/cli/server.cpp
```diff
@@ -277,37 +271,122 @@ bool check_data_ready(std::shared_future<T> &data, shared_ptr<HttpServer::Respon
...
+                    }
+                } else {
+                    for (const auto &[name, _] : indexes) {
+                        graphs_to_query.push_back(name);
+                    }
+                }
+
+                Json::Value merged;
+                for (const auto &name : graphs_to_query) {
+                    for (const auto &[graph_fname, anno_fname] : indexes[name]) {
```
adamant-pwn, line 356: Do we plan to parallelize this?

PR #549, opened 2025-10-08
Support a coordinate-to-accession map to a graph, mapping an annotation label+coordinate to an accession

hmusta:
For now, this is implemented as a graph extension, but the ideal thing would be to have a wrapper for MultiLabelAnnotation.

### DIFF
--- a/metagraph/src/cli/annotate.cpp
+++ b/metagraph/src/cli/annotate.cpp
@@ -396,16 +396,16 @@ int annotate_graph(Config *config) {
 
     const auto &files = config->fnames;
 
-    assert(config->infbase_annotators.size() <= 1);
-
-    const auto graph = load_critical_dbg(config->infbase);
-
     if (config->accessions) {
         common::logger->trace("Constructing sequence accession map");
         mtg::graph::RowTuplesToId accessions(files);
         accessions.serialize(config->outfbase);
         return 0;
     }
+
+    assert(config->infbase_annotators.size() <= 1);
+
+    const auto graph = load_critical_dbg(config->infbase);
 
     if (!config->separately) {
         annotate_data(graph, *config, files, config->outfbase);
--- a/metagraph/src/cli/annotate.cpp
+++ b/metagraph/src/cli/annotate.cpp
@@ -7,6 +7,7 @@
 #include "common/batch_accumulator.hpp"
 #include "common/threads/threading.hpp"
 #include "annotation/representation/row_compressed/annotate_row_compressed.hpp"
+#include "graph/graph_extensions/row_tuples_to_id.hpp"
 #include "seq_io/formats.hpp"
 #include "seq_io/sequence_io.hpp"
 #include "seq_io/kmc_parser.hpp"
@@ -398,6 +399,13 @@ int annotate_graph(Config *config) {
     assert(config->infbase_annotators.size() <= 1);
 
     const auto graph = load_critical_dbg(config->infbase);
+
+    if (config->accessions) {
+        common::logger->trace("Constructing sequence accession map");
+        mtg::graph::RowTuplesToId accessions(files);
+        accessions.serialize(config->outfbase);
+        return 0;
+    }
 
     if (!config->separately) {
         annotate_data(graph, *config, files, config->outfbase);
--- a/metagraph/src/cli/load/load_annotated_graph.cpp
+++ b/metagraph/src/cli/load/load_annotated_graph.cpp
@@ -25,6 +25,14 @@ using mtg::common::logger;
 std::unique_ptr<AnnotatedDBG> initialize_annotated_dbg(std::shared_ptr<DeBruijnGraph> graph,
                                                        const Config &config,
                                                        size_t max_chunks_open) {
+    if (config.accessions) {
+        common::logger->trace("Loading sequence accession map");
+        auto accessions = std::make_shared<RowTuplesToId>();
+        auto filename_base = utils::remove_suffix(config.infbase, graph->file_extension());
+        accessions->load(filename_base);
+        graph->add_extension(accessions);
+    }
+
     uint64_t max_index = graph->max_index();
 
     auto base_graph = graph;
@@ -80,16 +88,7 @@ std::unique_ptr<AnnotatedDBG> initialize_annotated_dbg(std::shared_ptr<DeBruijnG
 }
 
 std::unique_ptr<AnnotatedDBG> initialize_annotated_dbg(const Config &config) {
-    auto graph = load_critical_dbg(config.infbase);
-    if (config.accessions) {
-        common::logger->trace("Loading sequence accession map");
-        auto accessions = std::make_shared<RowTuplesToId>();
-        auto filename_base = utils::remove_suffix(config.infbase, graph->file_extension());
-        accessions->load(filename_base);
-        graph->add_extension(accessions);
-    }
-
-    return initialize_annotated_dbg(graph, config);
+    return initialize_annotated_dbg(load_critical_dbg(config.infbase), config);
 }
 
 
--- a/metagraph/src/cli/load/load_annotated_graph.cpp
+++ b/metagraph/src/cli/load/load_annotated_graph.cpp
@@ -12,6 +12,7 @@
 #include "load_graph.hpp"
 #include "load_annotation.hpp"
 #include "graph/graph_extensions/row_tuples_to_id.hpp"
+#include "common/utils/string_utils.hpp"
 
 
 namespace mtg {
@@ -80,8 +81,13 @@ std::unique_ptr<AnnotatedDBG> initialize_annotated_dbg(std::shared_ptr<DeBruijnG
 
 std::unique_ptr<AnnotatedDBG> initialize_annotated_dbg(const Config &config) {
     auto graph = load_critical_dbg(config.infbase);
-    if (config.accessions)
-        graph->add_extension(std::make_shared<RowTuplesToId>(config.fnames));
+    if (config.accessions) {
+        common::logger->trace("Loading sequence accession map");
+        auto accessions = std::make_shared<RowTuplesToId>();
+        auto filename_base = utils::remove_suffix(config.infbase, graph->file_extension());
+        accessions->load(filename_base);
+        graph->add_extension(accessions);
+    }
 
     return initialize_annotated_dbg(graph, config);
 }
--- a/metagraph/src/cli/load/load_annotated_graph.cpp
+++ b/metagraph/src/cli/load/load_annotated_graph.cpp
@@ -11,6 +11,7 @@
 #include "cli/config/config.hpp"
 #include "load_graph.hpp"
 #include "load_annotation.hpp"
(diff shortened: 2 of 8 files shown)


karasikov:
It would be good to have some Usage commands, how to run this.
(If not in the docs, at least in the PR description above.)

metagraph/src/graph/graph_extensions/row_tuples_to_id.hpp
```diff
@@ -0,0 +1,41 @@
...
+    void serialize(const std::string &filename_base) const;
+
+    bool is_compatible(const SequenceGraph &graph, bool verbose = true) const;
+
+    std::vector<std::tuple<std::string, size_t, std::vector<SmallVector<uint64_t>>>>
+    rows_tuples_to_label_tuples(const std::vector<RowTuples> &rows_tuple) const;
+
+  private:
+    std::vector<std::vector<std::string>> seq_id_labels_;
+    std::vector<bit_vector_sd> seq_delims_;
```
karasikov: How can we be sure the ordering is consistent with the order of the annotation columns?

metagraph/src/graph/graph_extensions/row_tuples_to_id.cpp
```diff
@@ -0,0 +1,123 @@
...
+    for (size_t i = 0; i < rows_tuples.size(); ++i) {
+        const auto &row_tuples = rows_tuples[i];
+        for (const auto &[col, tuple] : row_tuples) {
+            assert(col < seq_id_labels_.size());
+            assert(col < seq_delims_.size());
+            const auto &labels = seq_id_labels_[col];
+            const auto &delims = seq_delims_[col];
+            for (uint64_t coord : tuple) {
+                uint64_t seq_id = delims.rank1(coord);
+                uint64_t conv_coord = seq_id > 0 ? coord - delims.select1(seq_id) : coord;
```
karasikov:
```suggestion
                uint64_t seq_id = coord ? delims.rank1(coord - 1) : 0;
                uint64_t conv_coord = seq_id > 0 ? coord - delims.select1(seq_id) - 1 : coord;
```

karasikov: This can be optimized. Often coordinates are consequitive, so we can reuse previously computed values

metagraph/src/annotation/coord_to_accession.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+    CoordToAccession() {}
+    // Constructor from sequence names and number of k-mers in them, per annotation column.
+    CoordToAccession(std::vector<std::vector<std::string>> &&headers,
+                     std::vector<std::vector<uint64_t>> &&num_kmers);
+
+    bool load(const std::string &filename_base);
+    void serialize(const std::string &filename_base) const;
+
+    std::vector<std::tuple<std::string, size_t, std::vector<Tuple>>>
```
adamant-pwn: Can we use a struct with named fields, rather than `std::tuple` here?

metagraph/src/annotation/coord_to_accession.hpp
```diff
@@ -0,0 +1,53 @@
...
+namespace mtg {
+namespace annot {
+
+// A mapping from k-mer coordinates in columns to sequence accessions (headers).
+// Maps k-mer coordinates within annotation columns to sequence accessions (sequence headers).
+// When querying with flag `--accessions`, this mapping is used to transform coordinates and labels
+// in query results to show sequence-based (e.g., "<seq1>:0-0-3\t<seq3>:0-1-5") instead of file-
+// based coordinates (e.g., "<file.fa>:0-0-3:0-23-27"). This allows querying coordinate annotations
+// but displaying results as if sequences were annotated by their headers instead of source files.
+class CoordToAccession {
```
adamant-pwn: Why do we use the term "accessions" here, but I feel like most other places in the code refer to them as "headers"?

metagraph/src/annotation/coord_to_accession.cpp
```diff
@@ -0,0 +1,102 @@
...
+    for (size_t i = 0; i < rows_tuples.size(); ++i) {
+        const auto &row_tuples = rows_tuples[i];
+        for (const auto &[col, tuple] : row_tuples) {
+            const auto &delims = seq_delims_[col];
+            for (uint64_t coord : tuple) {
+                uint64_t seq_id = coord ? delims.rank1(coord - 1) : 0;
+                uint64_t conv_coord = seq_id > 0 ? coord - delims.select1(seq_id) - 1 : coord;
+                auto &tuple = conv_coords[std::make_pair(col, seq_id)];
+                tuple.resize(rows_tuples.size());
+                tuple[i].emplace_back(conv_coord);
```
adamant-pwn:
```suggestion
                auto [it, _] = conv_coords.try_emplace(std::make_pair(col, seq_id), rows_tuples.size());
                it.value()[i].emplace_back(conv_coord);
```
Seems semantically better.

metagraph/src/annotation/coord_to_accession.cpp
```diff
@@ -0,0 +1,102 @@
...
+std::vector<std::tuple<std::string, size_t, std::vector<Tuple>>>
+CoordToAccession::rows_tuples_to_label_tuples(const std::vector<RowTuples> &rows_tuples) const {
+    // RowTuples = Vector<std::pair<Column, Tuple>>
+    tsl::hopscotch_map<std::pair<Column, size_t>, std::vector<Tuple>> conv_coords;
+    for (size_t i = 0; i < rows_tuples.size(); ++i) {
+        const auto &row_tuples = rows_tuples[i];
+        for (const auto &[col, tuple] : row_tuples) {
+            const auto &delims = seq_delims_[col];
+            for (uint64_t coord : tuple) {
+                uint64_t seq_id = coord ? delims.rank1(coord - 1) : 0;
```
adamant-pwn:
```suggestion
                size_t seq_id = coord ? delims.rank1(coord - 1) : 0;
```
To match the type in `conv_coords`.

metagraph/src/cli/load/load_annotated_graph.cpp
```diff
@@ -63,11 +67,26 @@ std::unique_ptr<AnnotatedDBG> initialize_annotated_dbg(std::shared_ptr<DeBruijnG
...
+        if (config.accessions && config.identity != Config::ANNOTATE) {
+            if (config.infbase_annotators.size() > 1) {
+                logger->error("Cannot merge multiple CoordToAccession mappings."
+                              " Query a single annotation at a time.");
+                exit(1);
+            }
+            coord_to_accession = std::make_unique<annot::CoordToAccession>();
+            auto coord_to_acc_fname = utils::remove_suffix(config.infbase_annotators.at(0),
+                                                           annotation_temp->file_extension())
+                                                    + annot::CoordToAccession::kExtension;
```
adamant-pwn: But there is already `utils::make_suffix(filename_base, kExtension)` in `CoordToAccession::load`?

metagraph/src/cli/annotate.cpp
```diff
@@ -237,12 +240,48 @@ void annotate_data(std::shared_ptr<graph::DeBruijnGraph> graph,
...
+                false, // config.filename_anno
+                true, // config.annotate_sequence_headers
+                false, // parse_counts_from_headers
+                config.fasta_anno_comment_delim,
+                config.fasta_header_delimiter,
+                {}, // config.anno_labels
+                [&](std::string sequence, auto labels, uint64_t) {
+                    assert(labels.size() == 1 && "each sequence has a single label (header)");
+                    if (sequence.size() >= k) {
+                        if (headers[i].empty() || headers[i].back() != labels[0]) {
```
adamant-pwn: Should we verify somewhere that, except consecutive sequences, all headers are distinct?

metagraph/integration_tests/test_query.py
```diff
@@ -1327,5 +1327,324 @@ def test_batch_query_with_tiny_batch(self):
...
 
 
+@parameterized_class(('graph_repr', 'anno_repr'),
+    input_values=product(
+        [repr for repr in GRAPH_TYPES if not (repr == 'bitmap' and PROTEIN_MODE)],
+        [anno for anno in ANNO_TYPES if anno.endswith('_coord')],
+    ),
+    class_name_func=get_test_class_name
+)
+class TestAccessions(TestingBase):
```
adamant-pwn: Should we add a test for corrupted / mismatched `.seqs` file?

adamant-pwn (commented):
Overall looks good. Some comments so far.

metagraph/src/annotation/coord_to_accession.hpp
karasikov: Yeah, it's a bit convoluted. Not sure, having a struct will make it much clearer. I'll add a clarifying comment.

metagraph/src/cli/load/load_annotated_graph.cpp
karasikov: yeah, it's to make the error message correct in case of failed loading

metagraph/src/graph/annotated_dbg.cpp
```diff
@@ -501,6 +568,32 @@ AnnotatedDBG::get_kmer_coordinates(const std::vector<node_index> &nodes,
...
 
+    if (coord_to_accession_) {
+        if (tuple_matrix->get_binary_matrix().num_columns() != coord_to_accession_->num_columns()) {
+            logger->error("Incompatible number of columns in coord-to-accession mapping and annotation matrix: {} != {}",
+                          coord_to_accession_->num_columns(), tuple_matrix->get_binary_matrix().num_columns());
+            exit(1);
+        }
+        auto conv_results = coord_to_accession_->rows_tuples_to_label_tuples(rows_tuples);
+        if (num_top_labels < conv_results.size() || min_count > 1) {
+            std::sort(conv_results.begin(), conv_results.end(),
```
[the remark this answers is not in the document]
karasikov: Let's keep the behavior the same (unsorted if no filtering by top-labels, otherwise sorted). I changed it to std::nth_element + std::sort to avoid sorting the discarded stuff

metagraph/src/cli/annotate.cpp
karasikov: They don't have to be distinct (even though the query results might be hard to interpret everything will work even if all headers are the same). It's not super critical, not sure we need to bother about this

metagraph/integration_tests/test_query.py
karasikov: added a test

metagraph/src/annotation/coord_to_accession.hpp
karasikov: I'm fine to rename everything to headers

karasikov: Refactored. Also, now it's going to use the mapping automatically if it exists, so no need to pass special flags while querying.

adamant-pwn (approved):
Looks good to me, except one misaligned (?) indentation (see comment).

PR #563, opened 2025-11-06
Server changes: load all graphs, don't compute num_relations in stats

karasikov:
Improvements in server

* Computing num_relations takes up to 1 min for very large graphs that are loaded with mmap and not cached.
Now we don't report it, hence `/stats` always works instantly.

* Always keep all graphs loaded, so that they don't need to be reloaded for every query.

* Don't store copied annotation labels, which may create a large overhead for indexes with many labels

* Fixed catching exceptions thrown in worker threads, so now it reliably works with incorrect requests in multi-graph regime.

* Logging and other minor improvements

### DIFF
--- a/metagraph/src/cli/server.cpp
+++ b/metagraph/src/cli/server.cpp
@@ -345,8 +345,9 @@ int run_server(Config *config) {
         }
         for (auto &[graph_anno, anno_dbg] : graphs_cache) {
             Config config_copy = *config;
-            config_copy.infbase = graph_anno.first;
-            config_copy.infbase_annotators = { graph_anno.second };
+            const auto &[graph_fname, anno_fname] = graph_anno;
+            config_copy.infbase = graph_fname;
+            config_copy.infbase_annotators = { anno_fname };
             graphs_pool.enqueue([config_copy{std::move(config_copy)},anno_dbg{std::ref(anno_dbg)}]() {
                 initialize_annotated_dbg(config_copy).swap(anno_dbg);
             });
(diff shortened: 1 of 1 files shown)


metagraph/src/cli/server.cpp
```diff
@@ -275,6 +275,12 @@ int run_server(Config *config) {
 
     tsl::hopscotch_map<std::string, std::vector<std::pair<std::string, std::string>>> indexes;
 
+    ThreadPool graphs_pool(get_num_threads());
+    size_t num_server_threads = std::max(1u, get_num_threads());
+    set_num_threads(0);
+
+    std::unordered_map<std::pair<std::string, std::string>, std::unique_ptr<AnnotatedDBG>> graphs_cache;
```
adamant-pwn: Do we have overrides for default hasher of pairs? I don't think STL has one?

Also, I would prefer a small struct with named fields over `std::pair`.

metagraph/src/cli/server.cpp
```diff
@@ -330,24 +336,29 @@ int run_server(Config *config) {
...
-        for (const auto &[graph_fname, anno_fname] : graphs) {
-            auto &out = name_labels[name];
-            const auto &labels = read_labels(anno_fname);
-            out.insert(out.end(), labels.begin(), labels.end());
+        logger->info("[Server] Loading graphs...");
+        for (const auto &[name, graphs] : indexes) {
+            for (const auto &[graph_fname, anno_fname] : graphs) {
+                graphs_cache[{ graph_fname, anno_fname }] = nullptr;
+            }
+        }
```
adamant-pwn:
```suggestion
```
`std::unordered_map` will provide `std::nullptr` by default if the key is not stored, no? Or do we want tree structure to be static for parallel access?

metagraph/src/cli/server.cpp
karasikov: The purpose of two maps is to allow having the same (graph, anno) for multiple different names without duplicating it in memory. It's a weird case, but it happens in tests

adamant-pwn: Ah, alright, that's annoying to deal with. And making a custom hasher for in-place struct is a pain too. Let's just give names to `.first` and `.second` wherever we use them then?

PR #568, opened 2025-11-19
Aggregate k-mer counts with --count-kmers in transform_anno

adamant-pwn:
With a new test to verify correct filtering and aggregation.

### DIFF
--- a/metagraph/src/cli/config/config.cpp
+++ b/metagraph/src/cli/config/config.cpp
@@ -1252,7 +1252,7 @@ if (advanced) {
             fprintf(stderr, "\t                          \t\tignored in --count-kmers mode, use --min-count instead\n");
             fprintf(stderr, "\t   --max-value [INT] \t\tignore pre-aggregation counts larger than this [inf]\n");
             fprintf(stderr, "\t   --max-count [INT] \t\texclude k-mers with aggregated count larger than this [inf]\n");
-            fprintf(stderr, "\t                       \t\tin --count-kmers mode: max_count+1 must fit in count_width bits\n");
+            fprintf(stderr, "\t                       \t\tin --count-kmers mode: max_count+1 must fit in count_width bits (unless inf)\n");
             fprintf(stderr, "\t   --max-fraction [FLOAT] \texclude k-mers appearing in more than this fraction of columns [1.0]\n");
             fprintf(stderr, "\t                          \t\tignored in --count-kmers mode, use --max-count instead\n");
             fprintf(stderr, "\t   --compute-overlap [STR] \tcompute the number of shared bits in columns of this annotation and ANNOTATOR [off]\n");


metagraph/src/cli/config/config.cpp
```diff
@@ -1241,13 +1241,18 @@ if (advanced) {
             // fprintf(stderr, "\t-o --outfile-base [STR] basename of output file []\n");
             fprintf(stderr, "\t   --aggregate-columns \t\taggregate annotation columns into a bitmask (new column) [off]\n");
             fprintf(stderr, "\t                       \t\t\tFormula: min-count <= \\sum_i 1{min-value <= c_i <= max-value} <= max-count\n");
+            fprintf(stderr, "\t                       \t\t\tWith --count-kmers: min-count <= \\sum(filtered_counts) <= max-count\n");
             fprintf(stderr, "\t   --anno-label [STR]\t\tname of the aggregated output column [mask]\n");
-            fprintf(stderr, "\t   --min-value [INT] \t\tmin value for filtering [1]\n");
-            fprintf(stderr, "\t   --min-count [INT] \t\texclude k-mers appearing in fewer than this number of columns [1]\n");
+            fprintf(stderr, "\t   --count-kmers \t\taggregate k-mer counts (requires input annotations with counts) [off]\n");
```
karasikov: Previously, it was also reading counts if they were available.
So, the difference is in whether to _write_ the counts, right?

metagraph/src/cli/transform_annotation.cpp
```diff
@@ -427,23 +427,33 @@ int transform_annotation(Config *config) {
...
-
-        // TODO: set width to log2(max_cols + 1) but make sure atomic
-        //       increments don't overflow
-        sdsl::int_vector<> sum(0, 0, sdsl::bits::hi(num_columns) + 1);
+        
+        uint64_t min_threshold = config->min_count, max_threshold = config->max_count;
+        if (sum_counts) {
+            if (config->min_fraction > 0.0 || config->max_fraction < 1.0) {
+                logger->warn("--min-fraction and --max-fraction are ignored when --count-kmers is used. "
+                            "Use --min-count and --max-count to filter by aggregated k-mer counts.");
```
karasikov: Better throw an error and exit. There is no need to ever save runs with badly passed parameters. Better force the user to correct the parameters. Otherwise they might have wrong expectations and miss or neglect the warning at all.
```suggestion
                logger->error("--min-fraction and --max-fraction are ignored when --count-kmers is used. "
                            "Use --min-count and --max-count to filter by aggregated k-mer counts.");
                exit(1);
```

metagraph/src/cli/transform_annotation.cpp
```diff
@@ -427,23 +427,33 @@ int transform_annotation(Config *config) {
...
+        } else {
+            min_threshold = std::max<uint64_t>(std::ceil(num_columns * config->min_fraction),
+                                               min_threshold);
+            max_threshold = std::min<uint64_t>(std::floor(num_columns * config->max_fraction),
+                                               max_threshold);
+        }
+
+        uint64_t max_sum = max_threshold + 1;
+        if (sum_counts) {
+            max_sum = std::min<uint64_t>(max_sum, (1ULL << config->count_width) - 1);
```
karasikov:
```suggestion
            max_sum = std::min<uint64_t>(max_sum, sdsl::bits::lo_set[config->count_width]);
```

metagraph/src/cli/config/config.cpp
karasikov: I mean, explicitly writing that it "requires input annotations with counts" makes it confusing. Because it makes me think it uses the counts only in that case.

metagraph/src/cli/transform_annotation.cpp
```diff
@@ -427,23 +427,33 @@ int transform_annotation(Config *config) {
...
+                            "Use --min-count and --max-count to filter by aggregated k-mer counts.");
+            }
+        } else {
+            min_threshold = std::max<uint64_t>(std::ceil(num_columns * config->min_fraction),
+                                               min_threshold);
+            max_threshold = std::min<uint64_t>(std::floor(num_columns * config->max_fraction),
+                                               max_threshold);
+        }
+
+        uint64_t max_sum = max_threshold + 1;
```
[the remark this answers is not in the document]
karasikov: I think this should have been caught by integration tests

metagraph/src/cli/transform_annotation.cpp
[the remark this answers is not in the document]
adamant-pwn: Should we throw error if `max_threshold` is `< std::numeric_limits<unsigned int>::max()` and `max_threshold + 1` doesn't fit in `count_width` bits then?

metagraph/src/cli/transform_annotation.cpp
```diff
@@ -427,23 +427,33 @@ int transform_annotation(Config *config) {
...
+        if (sum_counts) {
+            max_sum = std::min<uint64_t>(max_sum, (1ULL << config->count_width) - 1);
+        }
+        sdsl::int_vector<> sum(0, 0, sdsl::bits::hi(max_sum) + 1);
```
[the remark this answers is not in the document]
adamant-pwn: Actually, this can't overflow. `sdsl::bits::hi` is `<= 63` and third argument of `int_vector<>` can be up to `64`.

metagraph/src/cli/transform_annotation.cpp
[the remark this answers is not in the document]
karasikov: No, I think that's not good. max_threshold can be infinity and everything should work even if count_width is relatively small.

But it's okay to throw error if count_width is too small (not enough to fit the result).

metagraph/src/cli/transform_annotation.cpp
```diff
@@ -429,21 +429,43 @@ int transform_annotation(Config *config) {
...
-        //       increments don't overflow
-        sdsl::int_vector<> sum(0, 0, sdsl::bits::hi(num_columns) + 1);
+                                        < std::numeric_limits<decltype(config->max_value)>::max();
+        
+        uint64_t min_threshold = config->min_count;
+        uint64_t max_threshold = config->max_count;
+        const bool fraction_upper_bounded = config->max_fraction < 1.0;
+        const bool count_upper_bounded = config->max_count < std::numeric_limits<decltype(config->max_count)>::max();
+        const bool upper_bounded = fraction_upper_bounded || count_upper_bounded;
+        if (config->min_fraction > 0.0 || fraction_upper_bounded) {
```
karasikov:
```suggestion
        if (config->min_fraction > 0.0 || config->max_fraction < 1.0) {
```

metagraph/integration_tests/test_transform_anno.py
```diff
@@ -146,5 +146,245 @@ def test_aggregate_columns_filtered(self):
...
+        # Use count-width 8 (max 255) - some counts will saturate
+        agg_counts = self._aggregate_and_query(count_width=8, output_name='agg_width8')
+        
+        # Verify each test k-mer with saturation at 255
+        for kmer, (logan_30_counts, logan_30_alt_counts) in self.test_kmers.items():
+            # Compute expected with saturation: cap each count at 255, then sum
+            saturated_30 = [min(c, 255) for c in logan_30_counts]
+            saturated_alt = [min(c, 255) for c in logan_30_alt_counts]
+            expected_raw = sum(saturated_30) + sum(saturated_alt)
+            expected = min(expected_raw, 255)  # Final result also saturates at 255
```
karasikov: are there any counts greater than 255 in the test file, actually?

karasikov: Would be good to add tests for edge cases. Even a single k-mer. Can be generated also in python.
with with different values of count-width, max_count, max_frequency, etc.

metagraph/src/cli/transform_annotation.cpp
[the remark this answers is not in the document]
akahles: I think this would be inconsistent with current behaviour. If I annotate a column without splitting, it will just cap the counts at 2**width. Then, if I split, maybe each split will not exceed the max, but the sum will, then my expectation is that I am not forced to increase the width but that it will cap at the max value.

karasikov (approved):

PR #569, opened 2025-11-21
Server queries in multi-graph regime with full loading to RAM

karasikov:
Support requests with flag `"in_ram": true`, which means the graph would be fully loaded to RAM before querying.


metagraph/src/cli/server.cpp
```diff
@@ -356,8 +359,15 @@ int run_server(Config *config) {
             exit(1);
         }
         logger->info("[Server] All graphs were loaded (with mmap). Ready to serve queries.");
+        utils::set_mmap(false);
```
adamant-pwn: I'm a bit confused on what it tries to do, right after saying "all graphs loaded with mmap".

PR #605, opened 2026-03-17
Parallel BRWT query in one traversal

karasikov:
Significantly speeds up queries in large graphs loaded with memory mapping (see #559).

## Query Performance Benchmarks

The following benchmarks compare several Metagraph query implementations across
multiple datasets. Results show the **mean ± standard deviation** across runs,
with the standard deviation also reported as a **percentage of the mean**.

<details>

<summary>Code for the benchmark</summary>

```bash
wait_for_low_cpu() {
    while true; do
        cpu=$(mpstat 1 1 | awk '/Average/ && $2=="all" {print 100-$12}')
        cpu=${cpu%.*}   # optional: drop decimals

        if (( cpu < 5 )); then
            break
        fi

        echo "CPU load ${cpu}% > 5%, sleeping 30s..."
        sleep 30
    done
}

for M in {metagraph_DNA_master,metagraph_DNA_paral4,metagraph_DNA_newBRWT,metagraph_DNA_new10k,metagraph_DNA_new,}; do
for x in {1..5}; do
    wait_for_low_cpu
    /usr/bin/time -v ~/metagraph/metagraph/build/$M query /data/random_stud/queries/100_studies_5k_short.fq -p 30 -v -i /scratch/nvme0/all_sra/data/superkingdom_bacteria/0052/graph.primary.small.dbg -a /scratch/nvme0/all_sra/data/superkingdom_bacteria/0052/annotation.clean.row_diff_brwt.annodbg -v --batch-size 1000000000  2>&1 1>/dev/null | grep -A 3 "Query annotation with" >>logs.txt
    echo -e "\n" >>logs.txt
done

for x in {1..5}; do
    wait_for_low_cpu
    /usr/bin/time -v ~/metagraph/metagraph/build/$M query --query-mode counts /data/random_stud/queries/100_studies_5k_short.fq -p 30 -v -i /scratch/nvme/all_sra/data/kingdom_fungi_RNA/0012/graph.primary.small.dbg -a /scratch/nvme/all_sra/data/kingdom_fungi_RNA/0012/annotation.clean.row_diff_int_brwt.annodbg -v --batch-size 1000000000  2>&1 1>/dev/null | grep -A 3 "Query annotation with" >>logs.txt
    echo -e "\n" >>logs.txt
done

for x in {1..5}; do
    wait_for_low_cpu
    python3 /data/scratch/drop_caches.py /scratch/nvme0/human/graph_merged_complete_k31.primary.small.indexed.dbg
    python3 /data/scratch/drop_caches.py /scratch/nvme7/human/annotation_10.row_diff_flat.annodbg
    /usr/bin/time -v ~/metagraph/metagraph/build/$M query --mmap /data/random_stud/queries/100_studies_50_short.fq -p 30 -v -i /scratch/nvme0/human/graph_merged_complete_k31.primary.small.indexed.dbg -a /scratch/nvme7/human/annotation_10.row_diff_flat.annodbg --min-kmers-fraction-label 0.0 --query-mode matches  2>&1 1>/dev/null | grep -A 3 "Query annotation with" >>logs.txt
    echo -e "\n" >>logs.txt
done

for x in {1..5}; do
    wait_for_low_cpu
    python3 /data/scratch/drop_caches.py /scratch/nvme4/plants/graph_primary.small.indexed.dbg
    python3 /data/scratch/drop_caches.py /scratch/nvme4/plants/*.row_diff_brwt.annodbg
    /usr/bin/time -v ~/metagraph/metagraph/build/$M query --mmap -i /scratch/nvme4/plants/*indexed.dbg -a /scratch/nvme4/plants/*.row_diff_brwt.annodbg /data/scratch/covid.fa -v -p 10 --min-kmers-fraction-label 0.0 --query-mode matches  2>&1 1>/dev/null | grep -A 3 "Query annotation with" >>logs.txt
    echo -e "\n" >>logs.txt
done

for x in {1..5}; do
    wait_for_low_cpu
    python3 /data/scratch/drop_caches.py /scratch/nvme4/plants/graph_primary.small.indexed.dbg
    python3 /data/scratch/drop_caches.py /scratch/nvme4/plants/*.row_diff_brwt.annodbg
    /usr/bin/time -v ~/metagraph/metagraph/build/$M query --mmap -i /scratch/nvme4/plants/*indexed.dbg -a /scratch/nvme4/plants/*.row_diff_brwt.annodbg /data/scratch/plants_query.fa -v -p 10 --min-kmers-fraction-label 0.0 --query-mode matches  2>&1 1>/dev/null | grep -A 3 "Query annotation with" >>logs.txt
    echo -e "\n" >>logs.txt
done
done
```
</details>

---

### all_sra / superkingdom_bacteria — 100_studies_5k_short.fq, 122,725,698 bp, in RAM, 30 threads

| Method | Runs | Query annotation (s) | Batch query time (s) | Throughput (bp/s) |
|---|---:|---:|---:|---:|
| new | 10 | **6.328 ± 0.136 (2.1%)** | 114.307 ± 6.699 (5.9%) | 1,072,622 ± 59,818 (5.6%) |
| newBRWT | 10 | 7.391 ± 0.133 (1.8%) | **112.606 ± 1.089 (1.0%)** | **1,085,264 ± 11,543 (1.1%)** |

---

### all_sra / kingdom_fungi_RNA + counts — 100_studies_5k_short.fq, 122,725,698 bp, in RAM, 30 threads

| Method | Runs | Query annotation (s) | Batch query time (s) | Throughput (bp/s) |
|---|---:|---:|---:|---:|
| new | 10 | **8.436 ± 0.133 (1.6%)** | **123.267 ± 0.976 (0.8%)** | **990,197 ± 7,797 (0.8%)** |
| newBRWT | 10 | 9.733 ± 0.350 (3.6%) | 124.984 ± 0.847 (0.7%) | 976,594 ± 6,613 (0.7%) |

---

### human — 100_studies_50_short.fq, 122,725,698 bp, mmap, 30 threads

| Method | Runs | Query annotation (s) | Batch query time (s) | Throughput (bp/s) |
|---|---:|---:|---:|---:|
| new | 10 | **43.5 ± 1.4 (3.1%)** | **64.6 ± 1.5 (2.3%)** | **18,977 ± 433 (2.3%)** |
| newBRWT | 10 | 47.0 ± 5.7 (12.2%) | 68.8 ± 5.6 (8.1%) | 17,919 ± 1,345 (7.5%) |

---

Now queries on a huge SRA-Plants graph with 1.4 TB annotation

### plants — covid.fa, 29,903 bp (only 16 k-mer matches), mmap, 10 threads

| Method | Runs | Query annotation (s) | Batch query time (s) | Throughput (bp/s) |
|---|---:|---:|---:|---:|
| new | 10 | 40.6 ± 3.7 (9.2%) | 57.2 ± 3.7 (6.5%) | 524.2 ± 31.7 (6.0%) |
| newBRWT | 10 | **32.9 ± 4.8 (14.6%)** | **49.5 ± 4.9 (9.8%)** | **607.9 ± 49.0 (8.1%)** |

### plants — plants_query.fa, 16,573 bp (16,408 k-mer matches), mmap, 10 threads

| Method | Runs | Query annotation (s) | Batch query time (s) | Throughput (bp/s) |
|---|---:|---:|---:|---:|
| new | 10 | **104.2 ± 0.7 (0.7%)** | **118.6 ± 0.8 (0.6%)** | **139.8 ± 0.9 (0.7%)** |
| newBRWT | 10 | 107.7 ± 4.4 (4.1%) | 122.1 ± 4.5 (3.6%) | 135.9 ± 4.7 (3.4%) |

---

### plants — 100_studies_20_short.fq, 491,594 bp (153,442 k-mer matches), mmap, 10 threads

| Method | Runs | Query annotation (s) | Batch query time (s) | Throughput (bp/s) |
|---|---:|---:|---:|---:|
| new | 1 | 498.9 ± — | 553.7 ± — | 888 ± — |
| newBRWT | 1 | **332.8 ± —** | **386.9 ± —** | **1,270 ± —** |

### plants — 100_studies_50_short.fq, 1,226,462 bp (366,859 k-mer matches), mmap, 10 threads

| Method | Runs | Query annotation (s) | Batch query time (s) | Throughput (bp/s) |
|---|---:|---:|---:|---:|
| new | 1 | 459.7 ± — | 530.0 ± — | 2,313 ± — |
| newBRWT | 1 | **322.8 ± —** | **395.1 ± —** | **3,102 ± —** |

### plants — 100_studies_100_short.fq, 2,454,716 bp (707,334 k-mer matches), mmap, 10 threads

| Method | Runs | Query annotation (s) | Batch query time (s) | Throughput (bp/s) |
|---|---:|---:|---:|---:|
| new | 1 | 446.0 ± — | 551.0 ± — | 4,452 ± — |
| newBRWT | 1 | **318.8 ± —** | **401.9 ± —** | **6,101 ± —** |

### plants — 100_studies_500_short.fq, 12,272,048 bp (3,128,611 k-mer matches), mmap, 10 threads

| Method | Runs | Query annotation (s) | Batch query time (s) | Throughput (bp/s) |
|---|---:|---:|---:|---:|
| new | 1 | 547.7 ± — | 734.3 ± — | 16,673 ± — |
| newBRWT | 1 | **355.0 ± —** | **518.5 ± —** | **23,581 ± —** |

### DIFF
--- a/metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
+++ b/metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
@@ -405,6 +405,8 @@ void BRWT::slice_rows(const std::vector<Row> &row_ids, std::vector<size_t> rows,
 
         for (size_t j = 0; j < child_nodes_.size(); ++j) {
             call_stack.back() = j;
+            // Each lambda captures `rows` and `call_stack` by value, giving each
+            // child its own copy to modify independently.
             thread_pool.force_enqueue_front([=,&thread_pool]() {
                 this->child_nodes_[j]->slice_rows(*child_row_ids_ptr, std::move(rows), root, call_stack,
                                                   max_columns_cutoff, thread_pool, call_slice);
--- a/metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
+++ b/metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
@@ -96,17 +96,12 @@ BRWT::slice_rows_parallel(const std::vector<Row> &row_ids, size_t num_threads) c
 #ifndef NDEBUG
 template <typename T>
 bool check_equal(const std::vector<T> &first, const std::vector<T> &second) {
-    if (first.size() != second.size())
-        return false;
-    for (size_t i = 0; i < first.size(); ++i) {
-        auto a = first[i];
-        auto b = second[i];
+    auto same_rows = [](T a, T b) {
         std::sort(a.begin(), a.end());
         std::sort(b.begin(), b.end());
-        if (a != b)
-            return false;
-    }
-    return true;
+        return a == b;
+    };
+    return std::equal(first.begin(), first.end(), second.begin(), second.end(), same_rows);
 }
 #endif // NDEBUG
 
@@ -199,7 +194,7 @@ std::vector<ColRanks> BRWT::get_column_ranks(const std::vector<Row>& row_ids,
     return result;
 }
 
-// Returns a pair (`nonzero_indices`, `child_row_ids`): indices into `rows` for
+// Returns a pair (`nonzero_indices`, `child_row_ids`): indices into `row_ids` for
 // the rows with the set bit in `nonzero_rows_`, and their recalculated
 // (via rank) row indices for the respective rows in the children.
 using NonzeroAndChildRows = std::pair<std::vector<size_t>, std::vector<BRWT::Row>>;
@@ -375,11 +370,11 @@ void BRWT::slice_rows(const std::vector<Row> &row_ids, Vector<T> *slice,
     slice->insert(slice->end(), row_ids.size() - last_nz, delim);
 }
 
-template <typename T>
+template <typename T, class CallSlice>
 void BRWT::slice_rows(const std::vector<Row> &row_ids, std::vector<size_t> rows,
                       const BRWT *root, std::vector<size_t> call_stack,
                       size_t max_columns_cutoff, ThreadPool &thread_pool,
-                      std::function<void(std::vector<size_t>&&, Vector<T>&&)> call_slice) const {
+                      CallSlice call_slice) const {
     if (row_ids.empty())
         return;
 
@@ -408,8 +403,8 @@ void BRWT::slice_rows(const std::vector<Row> &row_ids, std::vector<size_t> rows,
             // Each lambda captures `rows` and `call_stack` by value, giving each
             // child its own copy to modify independently.
             thread_pool.force_enqueue_front([=,&thread_pool]() {
-                this->child_nodes_[j]->slice_rows(*child_row_ids_ptr, std::move(rows), root, call_stack,
-                                                  max_columns_cutoff, thread_pool, call_slice);
+                this->child_nodes_[j]->slice_rows<T>(*child_row_ids_ptr, std::move(rows), root, call_stack,
+                                                     max_columns_cutoff, thread_pool, call_slice);
             });
         }
     } else {
--- a/metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
+++ b/metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
@@ -98,7 +98,20 @@ BRWT::get_rows(const std::vector<Row> &row_ids, size_t num_threads) const {
     if (get_one_pass_brwt())
         return slice_rows_parallel<SetBitPositions>(row_ids, num_threads);
 
-    return BinaryMatrix::get_rows(row_ids, num_threads);
+    auto result = BinaryMatrix::get_rows(row_ids, num_threads);
+
+#ifndef NDEBUG
+    auto check = slice_rows_parallel<SetBitPositions>(row_ids, num_threads);
+    for (size_t i = 0; i < row_ids.size(); ++i) {
+        auto a = result[i];
+        auto b = check[i];
+        std::sort(a.begin(), a.end());
+        std::sort(b.begin(), b.end());
+        assert(a == b);
+    }
+#endif
+
+    return result;
 }
 
 std::vector<BRWT::SetBitPositions>
@@ -115,6 +128,18 @@ BRWT::get_rows(const std::vector<Row> &row_ids) const {
         rows.emplace_back(row_begin, row_end);
     });
     assert(rows.size() == row_ids.size());
+
+#ifndef NDEBUG
+    auto check = slice_rows_parallel<SetBitPositions>(row_ids, 0);
+    for (size_t i = 0; i < row_ids.size(); ++i) {
+        auto a = rows[i];
+        auto b = check[i];
+        std::sort(a.begin(), a.end());
+        std::sort(b.begin(), b.end());
+        assert(a == b);
+    }
+#endif
+
     return rows;
 }
 
@@ -153,7 +178,7 @@ BRWT::get_column_ranks(const std::vector<Row> &row_ids, size_t num_threads) cons
     if (get_one_pass_brwt())
         return slice_rows_parallel<Vector<std::pair<Column, uint64_t>>>(row_ids, num_threads);
 
-    return get_row_data_parallel<Vector<std::pair<Column, uint64_t>>>(row_ids, num_threads,
+    auto result = get_row_data_parallel<Vector<std::pair<Column, uint64_t>>>(row_ids, num_threads,
                 [this](const std::vector<Row> &row_ids) {
         Vector<std::pair<Column, uint64_t>> slice;
         // expect at least 3 relations per row
@@ -169,6 +194,19 @@ BRWT::get_column_ranks(const std::vector<Row> &row_ids, size_t num_threads) cons
         assert(rows.size() == row_ids.size());
         return rows;
     });
+
+#ifndef NDEBUG
+    auto check = slice_rows_parallel<Vector<std::pair<Column, uint64_t>>>(row_ids, num_threads);
+    for (size_t i = 0; i < row_ids.size(); ++i) {
+        auto a = result[i];
+        auto b = check[i];
+        std::sort(a.begin(), a.end());
+        std::sort(b.begin(), b.end());
+        assert(a == b);
+    }
+#endif
+
+    return result;
 }
 
 std::pair<std::vector<size_t>, std::vector<BRWT::Row>>
--- a/metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
+++ b/metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
@@ -16,6 +16,11 @@ namespace matrix {
 
 const size_t kNumRowsInBlock = 250'000;
 
+static bool ONE_PASS_BRWT = false;
+
+void set_one_pass_brwt(bool value) { ONE_PASS_BRWT = value; }
+bool get_one_pass_brwt() { return ONE_PASS_BRWT; }
+
 
 bool BRWT::get(Row row, Column column) const {
     assert(row < num_rows());
@@ -90,12 +95,27 @@ BRWT::slice_rows_parallel(const std::vector<Row> &row_ids, size_t num_threads) c
 
 std::vector<BRWT::SetBitPositions>
 BRWT::get_rows(const std::vector<Row> &row_ids, size_t num_threads) const {
-    return slice_rows_parallel<SetBitPositions>(row_ids, num_threads);
+    if (get_one_pass_brwt())
+        return slice_rows_parallel<SetBitPositions>(row_ids, num_threads);
+
+    return BinaryMatrix::get_rows(row_ids, num_threads);
 }
 
 std::vector<BRWT::SetBitPositions>
 BRWT::get_rows(const std::vector<Row> &row_ids) const {
-    return get_rows(row_ids, 0);
+    Vector<Column> slice;
+    // expect at least 3 relations per row
+    slice.reserve(row_ids.size() * 4);
+
+    slice_rows(row_ids, &slice);
+
+    std::vector<SetBitPositions> rows;
+    rows.reserve(row_ids.size());
+    call_sliced_rows(slice, [&](auto row_begin, auto row_end) {
+        rows.emplace_back(row_begin, row_end);
+    });
+    assert(rows.size() == row_ids.size());
+    return rows;
 }
 
 void BRWT::call_rows(const std::function<void(const SetBitPositions &)> &callback,
@@ -130,7 +150,25 @@ void BRWT::call_rows(const std::function<void(const SetBitPositions &)> &callbac
 
 std::vector<Vector<std::pair<BRWT::Column, uint64_t>>>
 BRWT::get_column_ranks(const std::vector<Row> &row_ids, size_t num_threads) const {
-    return slice_rows_parallel<Vector<std::pair<Column, uint64_t>>>(row_ids, num_threads);
+    if (get_one_pass_brwt())
+        return slice_rows_parallel<Vector<std::pair<Column, uint64_t>>>(row_ids, num_threads);
+
+    return get_row_data_parallel<Vector<std::pair<Column, uint64_t>>>(row_ids, num_threads,
+                [this](const std::vector<Row> &row_ids) {
+        Vector<std::pair<Column, uint64_t>> slice;
+        // expect at least 3 relations per row
+        slice.reserve(row_ids.size() * 4);
+
+        slice_rows(row_ids, &slice);
+
+        std::vector<Vector<std::pair<Column, uint64_t>>> rows;
+        rows.reserve(row_ids.size());
+        call_sliced_rows(slice, [&](auto row_begin, auto row_end) {
+            rows.emplace_back(row_begin, row_end);
+        });
+        assert(rows.size() == row_ids.size());
+        return rows;
+    });
 }
 
 std::pair<std::vector<size_t>, std::vector<BRWT::Row>>
--- a/metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
+++ b/metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
@@ -93,6 +93,7 @@ BRWT::slice_rows_parallel(const std::vector<Row> &row_ids, size_t num_threads) c
     return rows;
 }
 
+#ifndef NDEBUG
 template <typename T>
 bool check_equal(const std::vector<T> &first, const std::vector<T> &second) {
     if (first.size() != second.size())
@@ -107,6 +108,7 @@ bool check_equal(const std::vector<T> &first, const std::vector<T> &second) {
     }
     return true;
 }
+#endif // NDEBUG
 
 std::vector<BRWT::SetBitPositions>
 BRWT::get_rows(const std::vector<Row> &row_ids, size_t num_threads) const {
@@ -167,20 +169,21 @@ void BRWT::call_rows(const std::function<void(const SetBitPositions &)> &callbac
     }
 }
 
-std::vector<Vector<std::pair<BRWT::Column, uint64_t>>>
-BRWT::get_column_ranks(const std::vector<Row> &row_ids, size_t num_threads) const {
+using ColRanks = Vector<std::pair<BRWT::Column, uint64_t>>;
+std::vector<ColRanks> BRWT::get_column_ranks(const std::vector<Row>& row_ids,
+                                             size_t num_threads) const {
     if (get_one_pass_brwt())
-        return slice_rows_parallel<Vector<std::pair<Column, uint64_t>>>(row_ids, num_threads);
+        return slice_rows_parallel<ColRanks>(row_ids, num_threads);
 
-    auto result = get_row_data_parallel<Vector<std::pair<Column, uint64_t>>>(row_ids, num_threads,
+    auto result = get_row_data_parallel<ColRanks>(row_ids, num_threads,
                 [this](const std::vector<Row> &row_ids) {
-        Vector<std::pair<Column, uint64_t>> slice;
+        ColRanks slice;
         // expect at least 3 relations per row
         slice.reserve(row_ids.size() * 4);
 
         slice_rows(row_ids, &slice);
 
-        std::vector<Vector<std::pair<Column, uint64_t>>> rows;
+        std::vector<ColRanks> rows;
         rows.reserve(row_ids.size());
         call_sliced_rows(slice, [&](auto row_begin, auto row_end) {
             rows.emplace_back(row_begin, row_end);
@@ -191,13 +194,18 @@ BRWT::get_column_ranks(const std::vector<Row> &row_ids, size_t num_threads) cons
 
     // When running the default BRWT query method, check the result against
     // the one-pass BRWT to avoid writing separate tests for the one-pass BRWT.
-    assert(check_equal(result, slice_rows_parallel<Vector<std::pair<Column, uint64_t>>>(row_ids, num_threads)));
+    assert(check_equal(result, slice_rows_parallel<ColRanks>(row_ids, num_threads)));
 
     return result;
 }
 
-std::pair<std::vector<size_t>, std::vector<BRWT::Row>>
-BRWT::get_nonzero_rows(const std::vector<Row> &row_ids, ThreadPool *thread_pool, bool adaptive_chunk_size) const {
+// Returns a pair (`nonzero_indices`, `child_row_ids`): indices into `rows` for
+// the rows with the set bit in `nonzero_rows_`, and their recalculated
+// (via rank) row indices for the respective rows in the children.
+using NonzeroAndChildRows = std::pair<std::vector<size_t>, std::vector<BRWT::Row>>;
+NonzeroAndChildRows BRWT::get_nonzero_rows(const std::vector<Row>& row_ids,
+                                           ThreadPool* thread_pool,
+                                           bool adaptive_chunk_size) const {
     std::vector<size_t> nonzero_indices;
     std::vector<Row> child_row_ids;
     nonzero_indices.reserve(row_ids.size());
@@ -205,14 +213,14 @@ BRWT::get_nonzero_rows(const std::vector<Row> &row_ids, ThreadPool *thread_pool,
 
     constexpr size_t kMaxChunkSize = 10'000;
     if (thread_pool && row_ids.size() > kMaxChunkSize) {
-        using Future = std::shared_future<std::pair<std::vector<size_t>, std::vector<Row>>>;
-        std::vector<std::pair<Future, size_t>> futures;
+        std::vector<std::pair<std::shared_future<NonzeroAndChildRows>, size_t>> futures;
         // allow smaller chunks for the root node for better parallelism
         const size_t chunk_size = adaptive_chunk_size
                                     ? get_chunk_size(row_ids.size(), kMaxChunkSize, thread_pool->num_threads())
                                     : kMaxChunkSize;
         for (size_t offset = 0; offset < row_ids.size(); offset += chunk_size) {
             futures.emplace_back(thread_pool->force_enqueue_front([&,offset]() {
+                // TODO: Use std::span when moving to C++20
                 return get_nonzero_rows({ row_ids.begin() + offset,
                                           std::min(row_ids.begin() + offset + chunk_size, row_ids.end()) });
             }), offset);
@@ -392,14 +400,13 @@ void BRWT::slice_rows(const std::vector<Row> &row_ids, std::vector<size_t> rows,
             rows[i++] = rows[idx];
         }
         rows.resize(i);
-        auto rows_ptr = std::make_shared<const std::vector<size_t>>(std::move(rows));
 
         call_stack.push_back(0);
 
         for (size_t j = 0; j < child_nodes_.size(); ++j) {
             call_stack.back() = j;
             thread_pool.force_enqueue_front([=,&thread_pool]() {
-                this->child_nodes_[j]->slice_rows(*child_row_ids_ptr, *rows_ptr, root, call_stack,
+                this->child_nodes_[j]->slice_rows(*child_row_ids_ptr, std::move(rows), root, call_stack,
                                                   max_columns_cutoff, thread_pool, call_slice);
             });
         }
--- a/metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
+++ b/metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
@@ -403,8 +403,9 @@ void BRWT::slice_rows(const std::vector<Row> &row_ids, std::vector<size_t> rows,
             // Each lambda captures `rows` and `call_stack` by value, giving each
             // child its own copy to modify independently.
             thread_pool.force_enqueue_front([=,&thread_pool]() {
-                this->child_nodes_[j]->slice_rows<T>(*child_row_ids_ptr, std::move(rows), root, call_stack,
-                                                     max_columns_cutoff, thread_pool, call_slice);
+                this->child_nodes_[j]->slice_rows<T>(*child_row_ids_ptr, std::move(rows), root,
+                                                     std::move(call_stack), max_columns_cutoff,
+                                                     thread_pool, std::move(call_slice));
             });
         }
     } else {
--- a/metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
+++ b/metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
@@ -8,7 +8,6 @@
 #include "common/algorithms.hpp"
 #include "common/serialization.hpp"
 #include "common/utils/template_utils.hpp"
-#include "common/unix_tools.hpp"
 
 
 namespace mtg {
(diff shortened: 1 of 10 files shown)


metagraph/src/common/threads/threading.hpp
```diff
@@ -44,12 +46,44 @@ class ThreadPool {
 
     template <class F, typename... Args>
     auto enqueue(F&& f, Args&&... args) {
-        return emplace(false, std::forward<F>(f), std::forward<Args>(args)...);
+        return emplace(false, false, std::forward<F>(f), std::forward<Args>(args)...);
```
adamant-pwn:
```suggestion
        return emplace(/*front*/false, /*force*/false, std::forward<F>(f), std::forward<Args>(args)...);
```
I would mark all those calls like this here and elsewhere, as it's much quicker to read.

Also, are these always known in compile-time? Might also make sense to put one or both of them as template arguments, rather than runtime ones. So, the call would be `empace<false, false>(...)` instead of `emplace(false, false, ...)`. Then, the check could be done with `if constexpr`. I also sometimes use something like `enum dir {front, back}` for this in similar cases, so it could be `emplace<front>(...)` and `emplace<back>(...)` or `emplace<dir>(...)` if enum value is a template argument. This way, the arguments are much easier to read than `true` / `false`.

metagraph/src/common/threads/threading.cpp
```diff
@@ -47,36 +49,45 @@ void ThreadPool::join() {
 
 void ThreadPool::remove_waiting_tasks() {
     std::unique_lock<std::mutex> lock(this->queue_mutex);
-    this->tasks = std::queue<std::function<void()>>();
-    this->empty_condition.notify_all();
+    this->tasks = std::deque<std::function<void()>>();
```
adamant-pwn:
```suggestion
    this->tasks = {};
```
Or would it not compile? I think empty initializer list should work for `std::deque`, maybe it didn't work for `std::queue`.

metagraph/src/common/threads/threading.hpp
```diff
@@ -44,12 +46,44 @@ class ThreadPool {
...
+    auto force_enqueue_front(F&& f, Args&&... args) {
+        return emplace(true, true, std::forward<F>(f), std::forward<Args>(args)...);
+    }
+
+    template <class T>
+    std::shared_future<T>& help_while_waiting(std::shared_future<T> &future) {
+        auto ready = [&]() {
+            return future.wait_for(std::chrono::seconds(0)) == std::future_status::ready;
+        };
+        while (workers.size() && !ready()) {
```
adamant-pwn, line 67:
```suggestion
        while (!workers.empty() && !ready()) {
```
Generally, I think it's semantically better to always use `empty()` or `!empty()` instead of `size()` when used as boolean, in other places too.

metagraph/src/common/threads/threading.hpp
```diff
@@ -58,22 +92,28 @@ class ThreadPool {
...
-    std::queue<std::function<void()>> tasks;
+    size_t num_waiting_;
+    std::deque<std::function<void()>> tasks;
     size_t max_num_tasks_;
 
     std::mutex queue_mutex;
     std::condition_variable empty_condition;
     std::condition_variable full_condition;
+    std::condition_variable all_waiting;
+    std::condition_variable help_condition;
```
adamant-pwn:
```suggestion
    std::condition_variable worker_wakeup; // task available or join
    std::condition_variable can_enqueue;
    std::condition_variable all_idle;
    std::condition_variable waiter_wakeup; // task available or future potentially arrived
```
It seems the standard convention is to name CVs after "what triggers wakeup", rather than "what caused the wait". In any case, it should be consistent. Otherwise, `empty_condition` is named after wait cause, but `all_waiting` is named after wakeup trigger. Should use the same approach for all.

metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
```diff
@@ -111,22 +181,52 @@ BRWT::get_column_ranks(const std::vector<Row> &row_ids, size_t num_threads) cons
...
-BRWT::get_nonzero_rows(const std::vector<Row> &row_ids) const {
+BRWT::get_nonzero_rows(const std::vector<Row> &row_ids, ThreadPool *thread_pool, bool adaptive_chunk_size) const {
     std::vector<size_t> nonzero_indices;
     std::vector<Row> child_row_ids;
     nonzero_indices.reserve(row_ids.size());
     child_row_ids.reserve(row_ids.size());
 
+    constexpr size_t kMaxChunkSize = 10'000;
+    if (thread_pool && row_ids.size() > kMaxChunkSize) {
+        using Future = std::shared_future<std::pair<std::vector<size_t>, std::vector<Row>>>;
```
adamant-pwn:
```suggestion
        using idxs_and_rows = std::pair<std::vector<size_t>, std::vector<Row>>;
        using Future = std::shared_future<idxs_and_rows>;
```
Can also put the alias outside of the function to re-use it for return type? I assume named structure is, again, not feasible here? I don't really like repeating all these deeply nested inline types...

metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
```diff
@@ -111,22 +181,52 @@ BRWT::get_column_ranks(const std::vector<Row> &row_ids, size_t num_threads) cons
...
+        std::vector<std::pair<Future, size_t>> futures;
+        // allow smaller chunks for the root node for better parallelism
+        const size_t chunk_size = adaptive_chunk_size
+                                    ? get_chunk_size(row_ids.size(), kMaxChunkSize, thread_pool->num_threads())
+                                    : kMaxChunkSize;
+        for (size_t offset = 0; offset < row_ids.size(); offset += chunk_size) {
+            futures.emplace_back(thread_pool->force_enqueue_front([&,offset]() {
+                return get_nonzero_rows({ row_ids.begin() + offset,
```
adamant-pwn, line 224:
```suggestion
                // TODO: Use std::span when moving to C++20
                return get_nonzero_rows({ row_ids.begin() + offset,
```
Maybe we should start putting these comments to re-discover this when we actually move. Copy is a bit wasteful, and spans would solve it almost for free.

metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
```diff
@@ -111,22 +181,52 @@ BRWT::get_column_ranks(const std::vector<Row> &row_ids, size_t num_threads) cons
...
+                                          std::min(row_ids.begin() + offset + chunk_size, row_ids.end()) });
+            }), offset);
+        }
+        for (auto &[future, offset] : futures) {
+            const auto &[nz, ids] = thread_pool->help_while_waiting(future).get();
+            for (size_t idx : nz) {
+                nonzero_indices.push_back(offset + idx);
```
adamant-pwn, line 231: Do we lose a lot in parallelism on doing this sequentially? Maybe we could instead allocate full storage of size `row_ids.size()` and make all threads put things in that storage in parallel. Then, when we know actual sizes, also run multi-threaded copy from full-size storage to a correctly compressed one.

metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
```diff
@@ -111,22 +181,52 @@ BRWT::get_column_ranks(const std::vector<Row> &row_ids, size_t num_threads) cons
...
+    constexpr size_t kMaxChunkSize = 10'000;
+    if (thread_pool && row_ids.size() > kMaxChunkSize) {
+        using Future = std::shared_future<std::pair<std::vector<size_t>, std::vector<Row>>>;
+        std::vector<std::pair<Future, size_t>> futures;
+        // allow smaller chunks for the root node for better parallelism
+        const size_t chunk_size = adaptive_chunk_size
+                                    ? get_chunk_size(row_ids.size(), kMaxChunkSize, thread_pool->num_threads())
+                                    : kMaxChunkSize;
+        for (size_t offset = 0; offset < row_ids.size(); offset += chunk_size) {
+            futures.emplace_back(thread_pool->force_enqueue_front([&,offset]() {
```
adamant-pwn, line 222: Did you run experiments with enqueueing front vs back? Any data on whether enqueueing to back here would ever provide better results?

metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
```diff
@@ -262,11 +363,71 @@ void BRWT::slice_rows(const std::vector<Row> &row_ids, Vector<T> *slice) const {
...
+            return;
+        auto child_row_ids_ptr = std::make_shared<const std::vector<Row>>(std::move(child_row_ids));
+
+        // keep only the indices of non-empty rows
+        size_t i = 0;
+        for (size_t idx : nonzero_indices) {
+            rows[i++] = rows[idx];
+        }
+        rows.resize(i);
+        auto rows_ptr = std::make_shared<const std::vector<size_t>>(std::move(rows));
```
adamant-pwn: I don't think this helps for `rows`. We could as well just replace `*rows_ptr` to `std::move(rows)` below, because it's passed by value anyway, and then modified in each child.

metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
```diff
@@ -262,11 +363,71 @@ void BRWT::slice_rows(const std::vector<Row> &row_ids, Vector<T> *slice) const {
...
+        // construct indexing for children and the inverse mapping
+        auto [nonzero_indices, child_row_ids] = get_nonzero_rows(row_ids, &thread_pool, adaptive_chunk_size);
+        if (nonzero_indices.empty())
+            return;
+        auto child_row_ids_ptr = std::make_shared<const std::vector<Row>>(std::move(child_row_ids));
+
+        // keep only the indices of non-empty rows
+        size_t i = 0;
+        for (size_t idx : nonzero_indices) {
+            rows[i++] = rows[idx];
```
adamant-pwn, line 400: Probably overhead here is negligible, but I imagine this could be parallelized too?

metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
```diff
@@ -53,6 +58,68 @@ void call_sliced_rows(Slice &slice, Callback call_row) {
...
+const size_t kMinColumnsForParallel = 10;
+
+// Maximum number of columns per chunk when deciding the cutoff for parallel subtree traversal.
+const size_t kMaxColumnsChunkSize = 1000;
+
+template <typename RowT>
+std::vector<RowT>
+BRWT::slice_rows_parallel(const std::vector<Row> &row_ids, size_t num_threads) const {
+    using T = typename RowT::value_type;
+    ThreadPool thread_pool(num_threads > 1 ? num_threads : 0);
```
adamant-pwn, line 76: I wonder if this check should be done inside `thread_pool`, actually? So, it would just treat `num_thread=1` as `0` always. Do we have any place where we want `thread_pool` to actually honestly have a pool of just 1 worker?

Might make sense to do separately from this PR...

metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
```diff
@@ -53,6 +58,68 @@ void call_sliced_rows(Slice &slice, Callback call_row) {
...
+    std::mutex mu;
+    std::vector<RowT> rows(row_ids.size());
+    slice_rows<T>(row_ids, utils::arange<size_t>(0, row_ids.size()), this, {},
+        std::max(kMinColumnsForParallel, get_chunk_size(num_columns(), kMaxColumnsChunkSize, num_threads)),
+        thread_pool,
+        [&](std::vector<size_t> &&sliced_rows, Vector<T> &&slice) {
+            auto rows_it = sliced_rows.begin();
+            std::lock_guard<std::mutex> lock(mu);
```
adamant-pwn, line 84: Did I already ask about possible overhead in doing this sequentially? We assume it's fine because it is processed while other workers still do something, so it pipelines well?

metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
```diff
@@ -53,6 +58,68 @@ void call_sliced_rows(Slice &slice, Callback call_row) {
...
+            });
+            assert(rows_it == sliced_rows.end());
+        }
+    );
+    thread_pool.join();
+    return rows;
+}
+
+template <typename T>
+bool check_equal(const std::vector<T> &first, const std::vector<T> &second) {
```
adamant-pwn, line 98: Maybe simpler to do with `std::equal` and custom comparator... But also, are you sure this should be here? I feel like it belongs to tests, and adding it to the main flow is too hacky?

metagraph/src/cli/server_utils.hpp
```diff
@@ -16,11 +16,7 @@ void process_request(std::shared_ptr<HttpServer::Response> &response,
 
 // An exception that may be thrown inside the callback `process` to indicate that the response
 // is already sent by the caller and no additional action in `process_request` is required.
-class CustomResponse : public std::exception {
-    const char* what() const noexcept override {
-        return "Response already sent by callback";
-    }
-};
+class CurrentlyInitializingError : public std::exception {};
```
adamant-pwn:
```suggestion
class CurrentlyInitializingError : public std::runtime_error {
public:
    CurrentlyInitializingError()
        : std::runtime_error("Server is currently initializing") {}
};

```

metagraph/src/cli/server_utils.cpp
```diff
@@ -107,25 +91,39 @@ void process_request(std::shared_ptr<HttpServer::Response> &response,
...
+        header = SimpleWeb::CaseInsensitiveMultimap({ { "Content-Type", "application/json" } });
+        if (is_compression_requested(request)) {
+            ret = compress_string(ret);
+            header.insert(std::make_pair("Content-Encoding", "deflate"));
+            header.insert(std::make_pair("Content-Length", std::to_string(ret.size())));
+        }
+    } catch (const CurrentlyInitializingError& e) {
+        logger->info("[Server] Got a request during initialization. Asked to come back later");
+        status = SimpleWeb::StatusCode::server_error_service_unavailable;
+        ret = "Server is currently initializing, please come back later.";
```
adamant-pwn:
```suggestion
        ret = json_str_with_error_msg(e.what());
```
If we also put message into `CurrentlyInitializingError`, to be consistent with the other path? Also, probably best to keep consistent JSON in all error responses, rather than put plain text here?

metagraph/src/cli/server_utils.hpp
```diff
@@ -16,11 +16,7 @@ void process_request(std::shared_ptr<HttpServer::Response> &response,
 
 // An exception that may be thrown inside the callback `process` to indicate that the response
 // is already sent by the caller and no additional action in `process_request` is required.
```
adamant-pwn:
```suggestion
```
Seems to be stale comment?

metagraph/src/cli/server_utils.cpp
```diff
@@ -107,25 +91,39 @@ void process_request(std::shared_ptr<HttpServer::Response> &response,
...
         logger->warn("[Server] Error on request {}", request_id);
-        response->write(SimpleWeb::StatusCode::server_error_internal_server_error,
-                        json_str_with_error_msg("Internal server error"));
+        status = SimpleWeb::StatusCode::server_error_internal_server_error;
+        ret = json_str_with_error_msg("Internal server error");
     }
-    logger->info("[Server] Request {} finished in {} sec", request_id, timer.elapsed());
+    double processing_time = timer.elapsed();
+    response->write(status, ret, header);
+    double transmission_time = timer.elapsed() - processing_time;
```
adamant-pwn: It appears that `response->write` may be buffered, so this may measure buffering time, instead of actual transmission time.

metagraph/src/cli/server_utils.cpp
```diff
@@ -107,25 +91,39 @@ void process_request(std::shared_ptr<HttpServer::Response> &response,
...
+        ret = Json::writeString(Json::StreamWriterBuilder(), process(content));
+        header = SimpleWeb::CaseInsensitiveMultimap({ { "Content-Type", "application/json" } });
+        if (is_compression_requested(request)) {
+            ret = compress_string(ret);
+            header.insert(std::make_pair("Content-Encoding", "deflate"));
+            header.insert(std::make_pair("Content-Length", std::to_string(ret.size())));
+        }
+    } catch (const CurrentlyInitializingError& e) {
+        logger->info("[Server] Got a request during initialization. Asked to come back later");
+        status = SimpleWeb::StatusCode::server_error_service_unavailable;
```
adamant-pwn, line 108: nit: It seems that 503 best practices also include sending Retry-After header to indicate appropriate time to retry.

metagraph/tests/test_utils.cpp
```diff
@@ -500,6 +502,542 @@ TEST(ThreadPool, DummyFuture) {
...
+}
+
+// ---------------------------------------------------------------------------
+//  ThreadPool: help_while_waiting — latency benchmark
+// ---------------------------------------------------------------------------
+
+// Measures the latency of help_while_waiting when the future is completed by
+// a worker (not stolen). The worker loop's help_condition.notify_one() wakes
+// HWW promptly; without it, HWW would rely on the 100μs poll timeout.
+TEST(ThreadPool, HelpWhileWaitingLatencyBenchmark) {
```
adamant-pwn, line 1103: This doesn't have any assertion, and appears to be the only test to use `[BENCH]`. Should we assert its execution time, or something like that? Or move it away from unit tests?

metagraph/tests/test_utils.cpp
```diff
@@ -500,6 +502,542 @@ TEST(ThreadPool, DummyFuture) {
 }
 
 
+// ---------------------------------------------------------------------------
+//  ThreadPool: force_enqueue_front ordering
```
adamant-pwn, line 525: Some further nice-to-have scenarios suggested for consideration from LLM:

- Tests for multiple callers calling `help_while_waiting` simultaneously
- ThreadPool is destroyed with pending tasks, and without calling join()
- A task throws an exception

Up to you if we'd like to add them here.

metagraph/tests/test_utils.cpp
```diff
@@ -500,6 +502,542 @@ TEST(ThreadPool, DummyFuture) {
...
+
+    // The worker picks up this slow task immediately — queue becomes empty.
+    // The task takes much longer than the 100μs wait_for timeout.
+    auto future = pool.enqueue([]() {
+        std::this_thread::sleep_for(std::chrono::milliseconds(50));
+        return 42;
+    });
+
+    // Small delay so the worker dequeues the task before we call HWW.
+    std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
adamant-pwn, line 874: This might be tight / race if e.g. CI runner is busy. LLM suggests doing this instead:
```cpp
    std::promise<void> started;
    auto future = pool.enqueue([&started]() {
        started.set_value();  // task is dequeued and running
        std::this_thread::sleep_for(std::chrono::milliseconds(50));
        return 42;
    });

    started.get_future().wait();
```
Similar things may apply to other tests.

adamant-pwn (changes requested):
Thanks for the pull request, impressive improvements on plants!

Does it improve other mmap-ed graphs to the same degree, like Logan sets? Also, did you make sure for run conditions to be comparable (e.g., doing cache evictions in both / running both on warm pages)? Annotation size seems to be worse in RAM graphs, any ideas why?

I also added some individual comments in specific files, hope they'll be of help 🙂

metagraph/src/common/threads/threading.hpp
karasikov: I'll add comments. Enum and other stuff seem like overkill. It's just 3 calls of a simple helper function.

metagraph/src/common/threads/threading.hpp
karasikov: For me, `workers.size()` reads as "if there are workers", and `!workers.empty()` as "if workers is not empty, and hence, there is one or more workers". I think the first one is easier.

metagraph/src/common/threads/threading.hpp
karasikov: good idea

metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
karasikov: No no, it's just a single pass over the vector -- super quick.
The parallelization above reduces the time from 30 sec to 1-5 sec.
In contrast, this pass takes milliseconds. No need to over-optimize.

metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
karasikov: Here it really shouldn't matter. The main point is to do the queries in parallel (will happen either way). The last part (copy the results and add offsets) is super fast.

metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
karasikov: true

metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
karasikov: yeah, it's milliseconds, no need to over-optimize

metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
karasikov: I think it's good to have this flexibility in the ThreadPool -- allow it to have a single worker (e.g., for debugging or lazy routines), and 0 to do work in the same thread, which can improve code efficiency without writing too much extra code.

Here, it really doesn't matter. We could just do `ThreadPool thread_pool(num_threads);` -- really minor thing.

metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
karasikov: Yeah, relatively well. Of course, it would be perfect if we could get rid of this lock... This is the best I came up with so far

metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
karasikov: I think it's fine. It's neat and simplifies testing a lot -- spares us from writing dozens of new tests.

metagraph/src/cli/server_utils.hpp
karasikov: I think it's not really an error

metagraph/src/cli/server_utils.cpp
karasikov: Do we want to change the API and switch to json here?

metagraph/src/cli/server_utils.cpp
karasikov: Really? I thought it has to send immediately. Otherwise it might exceed the timeout and so on.

metagraph/tests/test_utils.cpp
karasikov: It's a small thing. I think it's fine to have it here. If someone wants to play with the latencies and change condition variables later, they can quickly check on this test. If we move this to some separate place, no one will ever look there.

metagraph/tests/test_utils.cpp
karasikov: Testing with exceptions sounds good

metagraph/tests/test_utils.cpp
karasikov: Let's come back to this if the tests ever fail.

I checked that these tests are meaningful, and they did catch bugs. So, I'd like to have them as is.

karasikov: ok, actually a good point. It makes sense to use it. I applied it to some cases

metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
```diff
@@ -53,6 +58,70 @@ void call_sliced_rows(Slice &slice, Callback call_row) {
...
+        return false;
+    for (size_t i = 0; i < first.size(); ++i) {
+        auto a = first[i];
+        auto b = second[i];
+        std::sort(a.begin(), a.end());
+        std::sort(b.begin(), b.end());
+        if (a != b)
+            return false;
+    }
+    return true;
```
adamant-pwn:
```suggestion
    return std::equal(first.begin(), first.end(), second.begin(), second.end(), [](T a, T b) {
        std::sort(a.begin(), a.end());
        std::sort(b.begin(), b.end());
        return a == b;
    });
```
This should simplify it somewhat. I still feel it's misplaced in brwt.cpp though... Should we put it in `common/algorithms.hpp`? Or is this function too specific for that?

metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
adamant-pwn: So, where does it actually matter whether we enqueue in the front or in the back? I mean, there is a reason we introduce `enqueue_front`, and not just keep using regular enqueue everywhere, right?

metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
```diff
@@ -100,33 +169,72 @@ void BRWT::call_rows(const std::function<void(const SetBitPositions &)> &callbac
...
+    assert(check_equal(result, slice_rows_parallel<ColRanks>(row_ids, num_threads)));
+
+    return result;
 }
 
-std::pair<std::vector<size_t>, std::vector<BRWT::Row>>
-BRWT::get_nonzero_rows(const std::vector<Row> &row_ids) const {
+// Returns a pair (`nonzero_indices`, `child_row_ids`): indices into `rows` for
+// the rows with the set bit in `nonzero_rows_`, and their recalculated
+// (via rank) row indices for the respective rows in the children.
```
adamant-pwn:
```suggestion
// Returns a pair (`nonzero_indices`, `child_row_ids`): indices into `row_ids` for
// the rows with the set bit in `nonzero_rows_`, and their recalculated
// (via rank) row indices for the respective rows in the children.
```
Argument is called `row_ids`, not `rows`?

metagraph/src/cli/server_utils.hpp
adamant-pwn: I mean... Why did you put "Error" in its name then? :wink:  

I think, deriving from `std::runtime_error` is semantically appropriate, and practical (can put the string in constructor instead of overriding `what()`). Returning 503, and using `std::exception` is also something that inherently signifies an error, rather than a normal flow anyway.

metagraph/src/cli/server_utils.cpp
adamant-pwn: Yeah, I think it should be fine. Worst thing that could happen is probably somebody displaying `{"error":"Server is currently initializing..."}` as a string? I think it's lesser evil than e.g. someone seeing that all other responses are JSON, and then breaking due to plain text 503.

metagraph/src/cli/server_utils.cpp
adamant-pwn: Yes, `response->write(...)` only fills internal `asio::streambuf`, and this is what you're measuring. It only actually sends the response in custom deleter of `shared_ptr<Response>`. Which, I assume, also gets called somewhere close to exiting this function, but not insite `response->write(...)`.

metagraph/tests/test_utils.cpp
adamant-pwn: Yeah, I think also it's not that the test would fail, but rather just may not actually test what we expect it to test. So, could be in the future that they become flaky and fail/pass randomly instead of always failing, and it could be annoying to debug.

adamant-pwn (changes requested):
Thanks! A few more comments.

metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
karasikov: I think it's fine. It's only in .cpp and under pragmas, so clearly a thing for debugging.

metagraph/tests/test_utils.cpp
karasikov: I think it's a good test. We also don't want to see it performing horribly slowly. If it ever fails, it's worth looking into why it's so slow; one needs to investigate. Though I really don't think it's ever going to happen.

metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
karasikov: enqueue_front would lead to DFS, and enqueue_back (default) leads to BFS. Also, some tasks spawned from other tasks have to be processed asap, and the queue may already be full -- for that purpose, we have force front enqueue.

metagraph/src/cli/server_utils.cpp
karasikov: I see. So it seems like this thing is pretty useless, and it's not what we want to measure. I just removed it.
It will show processing time + total time.

metagraph/src/cli/server_utils.hpp
karasikov: okay, I made it std::runtime_error

metagraph/src/annotation/binary_matrix/multi_brwt/brwt.cpp
```diff
@@ -408,8 +403,8 @@ void BRWT::slice_rows(const std::vector<Row> &row_ids, std::vector<size_t> rows,
             // Each lambda captures `rows` and `call_stack` by value, giving each
             // child its own copy to modify independently.
             thread_pool.force_enqueue_front([=,&thread_pool]() {
-                this->child_nodes_[j]->slice_rows(*child_row_ids_ptr, std::move(rows), root, call_stack,
-                                                  max_columns_cutoff, thread_pool, call_slice);
+                this->child_nodes_[j]->slice_rows<T>(*child_row_ids_ptr, std::move(rows), root, call_stack,
```
adamant-pwn: nit: `std::move(call_stack)`? It's tiny, so it shouldn't matter, but still a good practice.

adamant-pwn (approved):
Looks good to me! Just one small comment about passing `call_stack`.

PR #611, opened 2026-04-20
Specify the cause of error when can't load a file (e.g., permission issues)

karasikov:
Make loader errors explain *why* a file can't be loaded (missing? unreadable?
  wrong format? corrupt?) and always include the offending path. The multi-graph
  server previously logged only the graph path on annotation-load failures,
  leaving operators to guess which CSV entry was bad.

  ## Motivating example

  `metagraph server_query <graphs.csv>` where one CSV entry points at a missing
  annotation.

  **Before** (graph path logged, annotation path nowhere to be seen):
  ```
  [error] Caught exception when loading columns from /.../missing.column.annodbg: can't open file: iostream error
  [error] Cannot load annotations for graph graph.dbg, file corrupted
  ```

  **After** (annotation path on a single `[error]` line, classified cause first):
  ```
  [error] Cannot load columns from '/.../missing.column.annodbg': No such file or directory (caught: can't open file: iostream error)
  ```

  ## Error-message changes

  | Scenario | Before | After |
  |---|---|---|
  | Missing graph | `Cannot load graph from file 'x.dbg'` | `Cannot load graph from 'x.dbg': No such file or directory` |
  | Unreadable file | same as missing | `... : Permission denied (missing read permission on this file)` |
  | Wrong extension | `needs a valid file extension` | `unrecognized file extension (expected one of .dbg/.orhashdbg/...)` |
  | Corrupt / truncated | same as missing | `... : could not read contents (possibly corrupted or incompatible)` |
  | Missing row-diff side-car | `Can't read anchor file: x.anchors` | `Cannot open anchor file 'x.anchors': No such file or directory` |

  Central helper: `utils::file_read_failure_detail(path)` classifies the likely
  cause and every loader site now appends `: {detail}`.

  ## Side improvements

  - `RowCompressed::merge_load` and `StaticBinRelAnnotator::load` now emit distinct
    "label encoder" vs "matrix" error messages so the user knows which part of
    the file failed.
  - `ColumnCompressed::merge_load` logs the offending label + actual/expected row
    count on column-size mismatch.
  - Graph extension list is driven by a single registry (`kGraphExtToType` in
    `load_graph.cpp`) used both by `parse_graph_type` and the "unrecognized
    extension" message.
  - Inner `anchor_.load()` / `fork_succ_.load()` return values are now checked —
    previously a truncated side-car silently loaded as an empty bitmap.
  - `integration_tests/main.py` enumerates failed/errored tests at the end of a
    parallel run (mirrors gtest's `[  FAILED  ]` block).

  ## Tests

  70+ new parameterized integration tests:
  - `test_query.TestLoadErrorMessages` — missing, unreadable, corrupt payloads,
    missing row-diff side-cars; every graph and annotation extension.
  - `test_api.TestAPIServerLoadErrors` — single-graph and multi-graph server
    load failures.

### DIFF
--- a/metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
+++ b/metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
@@ -476,11 +476,11 @@ bool ColumnCompressed<Label>::load(const std::string &filename) {
         }
 
     } catch (const std::exception &e) {
-        logger->error("Caught exception when loading columns from {}: {} [{}]", f, e.what(),
-                      utils::file_read_failure_detail(f));
+        logger->error("Cannot load columns from '{}': {} (caught: {})", f,
+                      utils::file_read_failure_detail(f), e.what());
         return false;
     } catch (...) {
-        logger->error("Unknown exception when loading columns from {} [{}]", f,
+        logger->error("Cannot load columns from '{}': {} (caught unknown exception)", f,
                       utils::file_read_failure_detail(f));
         return false;
     }
@@ -576,7 +576,7 @@ bool ColumnCompressed<Label>::merge_load(const std::vector<std::string> &filenam
         try {
             offsets[i] = read_num_labels(fname);
         } catch (...) {
-            logger->error("Can't load label encoder from {} [{}]", fname,
+            logger->error("Cannot load label encoder from '{}': {}", fname,
                           utils::file_read_failure_detail(fname));
             error_occurred = true;
         }
@@ -623,11 +623,11 @@ bool ColumnCompressed<Label>::merge_load(const std::vector<std::string> &filenam
                          std::move(new_column));
             }
         } catch (const std::exception &e) {
-            logger->error("Caught exception when loading columns from {}: {} [{}]",
-                          filename, e.what(), utils::file_read_failure_detail(filename));
+            logger->error("Cannot load columns from '{}': {} (caught: {})", filename,
+                          utils::file_read_failure_detail(filename), e.what());
             error_occurred = true;
         } catch (...) {
-            logger->error("Unknown exception when loading columns from {} [{}]",
+            logger->error("Cannot load columns from '{}': {} (caught unknown exception)",
                           filename, utils::file_read_failure_detail(filename));
             error_occurred = true;
         }
--- a/metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
+++ b/metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
@@ -524,6 +524,8 @@ bool ColumnCompressed<Label>::merge_load(const std::vector<std::string> &filenam
                     decompress_bitmap(col) |= *column;
                 }
             } else {
+                logger->error("Cannot load column '{}': got {} rows, expected {}",
+                              label, column->size(), num_rows_);
                 no_errors = false;
             }
         },
--- a/metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
+++ b/metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
@@ -476,10 +476,12 @@ bool ColumnCompressed<Label>::load(const std::string &filename) {
         }
 
     } catch (const std::exception &e) {
-        logger->error("Caught exception when loading columns from {}: {}", f, e.what());
+        logger->error("Caught exception when loading columns from {}: {} [{}]", f, e.what(),
+                      utils::file_read_failure_detail(f));
         return false;
     } catch (...) {
-        logger->error("Unknown exception when loading columns from {}", f);
+        logger->error("Unknown exception when loading columns from {} [{}]", f,
+                      utils::file_read_failure_detail(f));
         return false;
     }
 
@@ -531,9 +533,13 @@ bool ColumnCompressed<Label>::merge_load(const std::vector<std::string> &filenam
     if (merge_successful && no_errors) {
         logger->trace("Annotation loading finished ({} columns)", bitmatrix_.size());
         return true;
-    } else {
-        return false;
     }
+    if (merge_successful && !no_errors) {
+        logger->error("Cannot load annotations from {}: columns from different inputs "
+                      "have inconsistent row counts (reference is {} rows)",
+                      fmt::join(filenames, ", "), num_rows_);
+    }
+    return false;
 }
 
 template <typename Label>
@@ -568,7 +574,8 @@ bool ColumnCompressed<Label>::merge_load(const std::vector<std::string> &filenam
         try {
             offsets[i] = read_num_labels(fname);
         } catch (...) {
-            logger->error("Can't load label encoder from {}", fname);
+            logger->error("Can't load label encoder from {} [{}]", fname,
+                          utils::file_read_failure_detail(fname));
             error_occurred = true;
         }
     }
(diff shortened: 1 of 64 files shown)


metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
adamant-pwn, line 529: Let's add more details about what exactly failed? Filename, label, what we got instead of expected?

metagraph/src/annotation/representation/row_compressed/annotate_row_compressed.cpp
```diff
@@ -102,15 +105,24 @@ void RowCompressed<Label>::serialize(const std::string &filename) const {
...
+                      utils::file_read_failure_detail(fname));
         return false;
+    }
 
     try {
         bool is_successfully_loaded = label_encoder_.load(instream)
                                             && matrix_->load(instream);
-        if (filenames.size() == 1 || !is_successfully_loaded)
-            return is_successfully_loaded;
+        if (!is_successfully_loaded) {
```
adamant-pwn: Should we also separate encoder and matrix messages, like in annotation_matrix.cpp?

metagraph/src/cli/load/load_graph.cpp
```diff
@@ -67,8 +67,9 @@ std::shared_ptr<DeBruijnGraph> load_critical_dbg(const std::string &filename) {
         case Config::GraphType::SSHASH:
             return load_critical_graph_from_file<graph::DBGSSHash>(filename);
         case Config::GraphType::INVALID:
-            logger->error("Cannot load graph from file '{}', needs a valid file extension",
-                          filename);
+            logger->error("Cannot load graph from '{}': unrecognized file extension "
+                          "(expected .dbg, .orhashdbg, .hashstrdbg, .hashfastdbg, .bitmapdbg, "
```
adamant-pwn: I don't like to hardcode it like this... Maybe we could make `static std::map<std::string, Config::GraphType>` constructed from all `{*::kExtension, Config::GraphType::*}` pairs? Then, `parse_graph_type` could go over all keys and try to match them, instead of hardcoding, and here we could also just print the keys of that map.

metagraph/src/cli/transform_annotation.cpp
```diff
@@ -108,7 +108,9 @@ matrix::LinkageMatrix cluster_columns(const std::vector<std::string> &files,
     subsampling_pool.join();
 
     if (!success) {
-        logger->error("Could not load annotations");
+        if (anno_type != Config::ColumnCompressed) {
+            logger->error("Could not load annotations");
```
adamant-pwn: Why is this here, while all other inline messages in this file are removed?

metagraph/src/common/utils/file_utils.cpp
```diff
@@ -55,6 +57,28 @@ std::unique_ptr<std::ifstream> open_ifstream(const std::string &filename, bool m
...
+    }
+    if (!fs::exists(st))
+        return "no such file or directory";
+#if defined(__unix__) || defined(__unix) || defined(unix) || (defined(__APPLE__) && defined(__MACH__))
+    if (::access(path.c_str(), R_OK) != 0) {
+        const char *err = std::strerror(errno);
+        return std::string(err ? err : "access failed")
+                + " (missing read permission on this file)";
+    }
+#endif
```
adamant-pwn, line 83: Huh, these scenarios do not write in `ec`?

adamant-pwn (commented):
Looks good overall! Some (mostly) minor comments.

metagraph/src/annotation/representation/row_compressed/annotate_row_compressed.cpp
karasikov: done

metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
karasikov: done

metagraph/src/cli/transform_annotation.cpp
karasikov: fixed

metagraph/src/common/utils/file_utils.cpp
karasikov: added a comment with explanation

metagraph/src/cli/load/load_graph.cpp
karasikov: Great idea!

metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
```diff
@@ -524,6 +524,8 @@ bool ColumnCompressed<Label>::merge_load(const std::vector<std::string> &filenam
                     decompress_bitmap(col) |= *column;
                 }
             } else {
+                logger->error("Cannot load column '{}': got {} rows, expected {}",
+                              label, column->size(), num_rows_);
                 no_errors = false;
```
adamant-pwn, line 529: What do you think of making callback return success as true / false instead of being void and writing this into `no_errors`? Then it'd stop execution on first failure, and report it to `merge_successful`. It feels a bit counter-intuitive that this could fail, while having `merge_successful = true`...

metagraph/src/cli/load/load_graph.cpp
```diff
@@ -63,14 +57,15 @@ std::shared_ptr<DeBruijnGraph> load_critical_dbg(const std::string &filename) {
...
             return load_critical_graph_from_file<DBGHashFast>(filename);
         case Config::GraphType::BITMAP:
-            return load_critical_graph_from_file<graph::DBGBitmap>(filename);
+            return load_critical_graph_from_file<DBGBitmap>(filename);
         case Config::GraphType::SSHASH:
-            return load_critical_graph_from_file<graph::DBGSSHash>(filename);
-        case Config::GraphType::INVALID:
+            return load_critical_graph_from_file<DBGSSHash>(filename);
+        case Config::GraphType::INVALID: {
+            auto exts = utils::get_firsts<std::vector<std::string>>(kGraphExtToType);
```
adamant-pwn: Hmm, this still writes every graph type down a second time... But we don't seem to use `parse_graph_type` anywhere else, it's just a bridge between extensions and loaders. How about we map extensions to loaders directly:

```cpp
static const std::array kGraphLoaders{
    std::pair{DBGSuccinct::kExtension,    &load_critical_graph_from_file<DBGSuccinct>},
    std::pair{DBGHashOrdered::kExtension, &load_critical_graph_from_file<DBGHashOrdered>},
    std::pair{DBGHashString::kExtension,  &load_critical_graph_from_file<DBGHashString>},
    std::pair{DBGHashFast::kExtension,    &load_critical_graph_from_file<DBGHashFast>},
    std::pair{DBGBitmap::kExtension,      &load_critical_graph_from_file<DBGBitmap>},
    std::pair{DBGSSHash::kExtension,      &load_critical_graph_from_file<DBGSSHash>},
};

...


std::shared_ptr<DeBruijnGraph> load_critical_dbg(const std::string &filename) {
    for (const auto &[ext, loader] : kGraphLoaders) {
        if (utils::ends_with(filename, ext))
            return loader(filename);
    }

    
    auto exts = utils::get_firsts<std::vector<std::string>>(kGraphLoaders);
    logger->error("Cannot load graph from '{}': unrecognized file extension "
                  "(expected one of {})", filename, fmt::join(exts, "/"));
    std::exit(1);
}
```
And then drop `parse_graph_type`?

adamant-pwn (approved):
Looks good overall. Two more ~minor suggestions.

metagraph/src/annotation/representation/column_compressed/annotate_column_compressed.cpp
karasikov: Yeah, I see what you mean. However, I think it won't make it much nicer, because the main issue right now is that `merge_load` doesn't really merge -- it just calls columns from different annotators, and they may be incompatible. I think we'd need a bigger refactoring to make it absolutely perfect.

adamant-pwn: Alright then :eyes:

PR #612, opened 2026-04-20
Added support for conversion row_diff_brwt -> row_diff_disk

karasikov:


metagraph/src/annotation/annotation_converters.cpp
adamant-pwn:
```suggestion
                         fname, true /* append_file */);
```
Why are RowFlat and RowDisk inconsistent in positioning of `fname` argument, though?.. And only one of them has `append_file`? Would it be feasible to make the interface uniform?

Then, the code would be the same and we could keep it a more generic template, writing `RowDiffAnnotator::serialize(...)` for template argument `RowDiffAnnotator`.

metagraph/src/cli/config/config.cpp
```diff
@@ -626,11 +626,23 @@ Config::Config(int argc, char *argv[]) {
...
         } else if (!to_row_diff && infbase.size()) {
-            std::cerr << "Graph is only required for transform to row_diff types" << std::endl;
+            std::cerr << "Graph is only required for transform to row_diff types"
+                      << std::endl;
+            print_usage_and_exit = true;
+        } else if (input_has_embedded_anchor && infbase.size()) {
+            std::cerr << "Graph is not needed when the input annotation has "
+                         "embedded anchors (row_diff_brwt, row_diff_flat)"
+                      << std::endl;
             print_usage_and_exit = true;
```
adamant-pwn: Should this really be an error, rather than e.g. just ignoring the graph?

metagraph/src/cli/config/config.cpp
```diff
@@ -626,11 +626,23 @@ Config::Config(int argc, char *argv[]) {
                                     || anno_type == RowDiffDiskCoord
                                     || anno_type == RowDiffBRWTCoord
                                     || anno_type == RowDiffCoord;
-        if (to_row_diff && !infbase.size()) {
+        // For row_diff_brwt and row_diff_flat inputs, the anchor and fork-succ
+        // bitmaps are embedded in the .annodbg, so no graph is needed when
+        // re-formatting to another row_diff representation.
+        const bool input_has_embedded_anchor = fnames.size()
+                && (utils::ends_with(fnames.front(), ".row_diff_brwt.annodbg")
```
adamant-pwn: Is there `kExtension` or similar we could use here instead of inline constants? Or, better yet, maybe we could peek into annotation header to check this regardless of file extension?

metagraph/src/cli/config/config.cpp
```diff
@@ -626,11 +626,23 @@ Config::Config(int argc, char *argv[]) {
...
+                        || utils::ends_with(fnames.front(), ".row_diff_flat.annodbg"));
```
adamant-pwn: Should we check that `fnames.size() == 1` here already, rather than in transform_annotation?

metagraph/integration_tests/test_transform_anno.py
```diff
@@ -397,5 +397,101 @@ def test_aggregate_counts_combinations(self, count_width, min_value, max_value,
                 f"K-mer {kmer}: test={test_name}, expected {expected}, got {actual}")
 
 
+class TestRowDiffAnnoConverters(TestingBase):
```
adamant-pwn, line 400: AI suggestions for more scenarios:


- All-zero rows untested. RowDisk's delta encoding should handle empty SetBitPositions, but the new tests don't exercise it. Adding a label that doesn't appear in any sequence would cover it.
- No round-trip test. E.g. column → row_diff → brwt → disk → flat and check semantic equivalence. Useful for catching subtle ordering or off-by-one issues. Could be a follow-up.
- No memory-bound test. The PR claims streaming/O(1), and the code looks like it. A test that converts a multi-GB matrix and watches RSS would prove it. Realistically out-of-scope for unit/integration tests.

Up to you if worth checking.

metagraph/src/annotation/annotation_converters.cpp
```diff
@@ -382,21 +382,26 @@ void convert_to_row_diff<RowDiffRowFlatAnnotator>(
...
+template <class Anno>
+static void write_row_diff_header(const std::string &fname, const Anno &anno) {
     std::ofstream out = utils::open_new_ofstream(fname);
     if (!out.good())
         throw std::ofstream::failure("Can't write to " + fname);
-
     anno.get_label_encoder().serialize(out);
-
-    // serialize RowDiff<RowFlat<>>
     out.write("v2.0", 4);
```
adamant-pwn, line 395: We have this `v2.0` inline everywhere... Should we update to e.g. `kLatest` constant?

metagraph/src/annotation/annotation_converters.cpp
```diff
@@ -406,6 +411,40 @@ void convert_to_row_diff<RowDiffRowFlatAnnotator>(const RowDiffBRWTAnnotator &an
...
+// cheap enough that streaming row-by-row keeps up with disk I/O.
+template <>
+void convert_to_row_diff<RowDiffDiskAnnotator>(const RowDiffRowFlatAnnotator &anno,
+                                               const std::string &outfbase) {
+    const auto &fname = utils::make_suffix(outfbase, RowDiffDiskAnnotator::kExtension);
+    write_row_diff_header(fname, anno);
+
+    const auto &flat = anno.get_matrix().diffs();
+    RowDisk::serialize(fname,
+        [&](auto callback) {
```
adamant-pwn: I feel like it'd be better to implement `call_rows` member function for flat annotator as well, so that all three converters just have the same base code and the only change is which type gets substituted. We can maybe add `enable_if` or similar to restrict the function to only RowDiff types?

adamant-pwn (changes requested):
Thanks for the PR! I've added a few comments.

metagraph/src/annotation/annotation_converters.cpp
karasikov: I prefer to write this explicitly

metagraph/integration_tests/test_transform_anno.py
karasikov: added tests

adamant-pwn (approved):
Looks good to me!

PR #613, opened 2026-04-20
run unit tests in parallel

karasikov:


metagraph/tests/test_fasta_io.cpp
```diff
@@ -236,10 +237,10 @@ TEST(FastaFile, twice_iterator_read_empty_with_move) {
 
         num_records = 0;
         total_size = 0;
-        std::for_each(std::move(begin), parser.end(), [&](const auto &record) {
+        for (auto it = std::move(begin); it != parser.end(); ++it) {
             num_records++;
-            total_size += record.seq.l;
-        });
+            total_size += it->seq.l;
+        }
```
adamant-pwn, line 243:
```suggestion
        for (const auto &record: parser) {
            num_records++;
            total_size += record.seq.l;
        }
```

This should work as intended for any structure that has well-defined begin() and end() already, no?

Would also need to remove `auto begin = parser.begin()` that isn't used anywhere else, and also change in other tests in this file.

metagraph/src/common/utils/file_utils.cpp
```diff
@@ -25,6 +25,11 @@ namespace fs = std::filesystem;
 std::filesystem::path SWAP_PATH;
 std::vector<std::string> TMP_DIRS;
 std::mutex TMP_DIRS_MUTEX;
+// The pid of the process that first registered a temp dir. Used to skip
+// cleanup in forked children — e.g. gtest's threadsafe death-test style forks
+// the test process, and the child's atexit would otherwise wipe the parent's
+// temp dirs out from under it.
+pid_t TMP_DIRS_OWNER_PID = 0;
```
adamant-pwn: Should this be `std::atomic<pid_t>`?

.github/workflows/linux.yml
```diff
@@ -168,7 +168,7 @@ jobs:
       - &run-unit-tests
         name: run unit tests
         run: |
-          cd metagraph/build && chmod +x unit_tests && ./unit_tests
+          cd metagraph/build && chmod +x unit_tests && ../scripts/unit_tests_parallel.sh
```
adamant-pwn:
```suggestion
          cd metagraph/build && chmod +x unit_tests && ./unit_tests_parallel
```

.github/workflows/macos.yml
```diff
@@ -159,7 +159,7 @@ jobs:
 
       - &run-unit-tests
         name: run unit tests
-        run: cd metagraph/build && chmod +x unit_tests && ./unit_tests
+        run: cd metagraph/build && chmod +x unit_tests && ../scripts/unit_tests_parallel.sh
```
adamant-pwn:
```suggestion
        run: cd metagraph/build && chmod +x unit_tests && ./unit_tests_parallel
```

metagraph/tests/test_helpers.hpp
```diff
@@ -1,13 +1,83 @@
...
+            if (getpid() != owner_pid)
+                return;
+            std::error_code ec;
+            fs::remove_all(test_dump_dir(), ec);
+        });
+        // Clean up on Ctrl+C / kill too — default signal actions terminate
+        // the process without running atexit, which would leak the tempdir.
+        // Only install where the disposition is still default, so we don't
+        // clobber handlers gtest or the death-test machinery installed.
+        auto handler = +[](int sig) {
```
adamant-pwn: Are you sure we need signal handling in this function? It seems to be already handled in unit_tests_parallel.sh that removes test root normally, so this is redundant?

metagraph/scripts/unit_tests_parallel.sh
```diff
@@ -0,0 +1,262 @@
...
+
+GREEN=$'\033[0;32m'
+RED=$'\033[0;31m'
+RST=$'\033[0m'
+
+print_help() {
+    sed -n '2,/^$/ { s/^# \{0,1\}//; p; }' "${BASH_SOURCE[0]}"
+}
+
+WORKERS=$(nproc 2>/dev/null || sysctl -n hw.ncpu 2>/dev/null || echo 4)
```
adamant-pwn: It looks like shards aren't balanced very well now. Comparing macos-latest, x86_64 runs on master and here, we have 2684s on master and 1681s, 925s, 1506s here. It's a nice improvement, but looks like could be even better with a more uniform load. I would try putting WORKERS to 2x of this. It would create some contention, but could hopefully help overall if it allows for a more uniform runtime distribution between shards.

metagraph/scripts/unit_tests_parallel.sh
```diff
@@ -0,0 +1,262 @@
...
+    LOGS[$i]="$LOG"
+    # TEST_DUMP_DIR pins test_dump_dir() inside $ROOT so transient dump
+    # files get swept along with the rest of the shard workdir, including
+    # on Ctrl+C when C++ atexit handlers wouldn't have run.
+    # `exec` replaces the subshell with unit_tests so $! points at the real
+    # binary — otherwise `kill $!` just kills the subshell and orphans the
+    # child, which then reparents to init and keeps running.
+    (
+        cd "$ROOT/shard$i" \
+            && GTEST_TOTAL_SHARDS="$WORKERS" GTEST_SHARD_INDEX="$i" \
```
adamant-pwn, line 165: What if user specified their own GTEST_TOTAL_SHARDS or GTEST_SHARD_INDEX via env? Could be good idea to also use it as default for WORKERS / --shard.

metagraph/scripts/unit_tests_parallel.sh
```diff
@@ -0,0 +1,262 @@
...
+WORKERS=$(nproc 2>/dev/null || sysctl -n hw.ncpu 2>/dev/null || echo 4)
+UNIT_TESTS="${UNIT_TESTS:-}"
+
+# Parse our flags; everything else is forwarded to unit_tests.
+while [[ $# -gt 0 ]]; do
+    case "$1" in
+        -j|--jobs)   WORKERS="$2"; shift 2 ;;
+        -j*)         WORKERS="${1#-j}"; shift ;;
+        --jobs=*)    WORKERS="${1#--jobs=}"; shift ;;
+        -h|--help)   print_help; exit 0 ;;
```
adamant-pwn, line 68: Maybe add `--shard` and `--keep-tmp` (skips rm of `$ROOT`) or `--workdir DIR` for debugging.

adamant-pwn (commented):
Nice idea! Too bad that runners only have 3x CPU in macos-latest on x86_64, so speedup there is a bit limited, but this could be pretty huge on local runs.

Theoretically we can further push it by splitting unit tests (e.g. by filtering) between runners, but we'll probably still be bottlenecked by small number of MacOS runners.

metagraph/tests/test_fasta_io.cpp
karasikov: It's an unusual type of iterator; it doesn't work like that here.

metagraph/tests/test_helpers.hpp
karasikov: One can still run tests with the binary `unit_tests`

metagraph/scripts/unit_tests_parallel.sh
karasikov: I looked into that a little. It looked like a rabbit hole. With this setup, it seems best to keep things simple.

metagraph/scripts/unit_tests_parallel.sh
karasikov: So far, I haven't needed that. Do you think that could be useful?

metagraph/src/common/utils/file_utils.cpp
karasikov: It's not meant to work with multithreading. There is no thread safety with other variables as well.

karasikov: Ah, no. My bad

metagraph/scripts/unit_tests_parallel.sh
adamant-pwn: I think I might want to inspect workdir for failing tests sometimes. But it's also not a big deal to modify locally... We can skip if it feels like too much work for little benefit at the moment.

metagraph/scripts/unit_tests_parallel.sh
adamant-pwn: Huh? What kind of a rabbit hole is it to multiply number of processes by a small constant and see if it improves balancing? I don't suggest anything more complicated, like runtime-aware split or anything :thinking:

.github/workflows/linux.yml
adamant-pwn: Whoops, looks like this also needs to add unit_tests_parallel to uploaded artifacts...

metagraph/tests/test_fasta_io.cpp
karasikov: No, actually it does work, but the whole point of the test is to actually check that std::move works for the iterator.

adamant-pwn: Ahhh, I see, okay then.

.github/workflows/linux.yml
adamant-pwn, line 131:
```suggestion
            metagraph/build/unit_tests_parallel
            metagraph/build/unit_tests
```
And same for macos should help.

.github/workflows/macos.yml
adamant-pwn, line 129:
```suggestion
            metagraph/build/unit_tests_parallel
            metagraph/build/unit_tests
```

metagraph/scripts/unit_tests_parallel.sh
karasikov: I think it only has a static assignment, or you suggest running n_cores * 2 tasks?

adamant-pwn: I suggest running `nproc * 2` processes with static scheduling, e.g. 6 shards on 3 cores. With 6 buckets instead of 3, the variance of the slowest bucket goes down (each heavy test gets averaged against fewer peers, so the max-shard runtime gets closer to the mean). The OS time-slices between the extra processes; the hope is the variance reduction outweighs the context-switching cost.

karasikov: yeah, that we can do

metagraph/src/common/utils/file_utils.cpp
karasikov: Actually, yeah, I was correct. In create_temp_dir, it's already in a critical section, and on cleanup, nothing is synced; we assume that all worker threads are done at that point and forgo thread-safety there.

.github/workflows/linux.yml
```diff
@@ -168,7 +169,7 @@ jobs:
       - &run-unit-tests
         name: run unit tests
         run: |
-          cd metagraph/build && chmod +x unit_tests && ./unit_tests
+          cd metagraph/build && chmod +x unit_tests && ./unit_tests_parallel
```
adamant-pwn:
```suggestion
          cd metagraph/build && chmod +x unit_tests_parallel && ./unit_tests_parallel
```

.github/workflows/macos.yml
```diff
@@ -159,7 +160,7 @@ jobs:
 
       - &run-unit-tests
         name: run unit tests
-        run: cd metagraph/build && chmod +x unit_tests && ./unit_tests
+        run: cd metagraph/build && chmod +x unit_tests && ./unit_tests_parallel
```
adamant-pwn:
```suggestion
        run: cd metagraph/build && chmod +x unit_tests_parallel && ./unit_tests_parallel
```

.github/workflows/linux.yml
```diff
@@ -168,7 +169,7 @@ jobs:
       - &run-unit-tests
         name: run unit tests
         run: |
-          cd metagraph/build && chmod +x unit_tests && ./unit_tests
+          cd metagraph/build && chmod +x unit_tests_parallel && ./unit_tests_parallel
```
adamant-pwn:
```suggestion
          cd metagraph/build && chmod +x unit_tests{,_parallel} && ./unit_tests_parallel
```

.github/workflows/macos.yml
```diff
@@ -159,7 +160,7 @@ jobs:
 
       - &run-unit-tests
         name: run unit tests
-        run: cd metagraph/build && chmod +x unit_tests && ./unit_tests
+        run: cd metagraph/build && chmod +x unit_tests_parallel && ./unit_tests_parallel
```
adamant-pwn:
```suggestion
        run: cd metagraph/build && chmod +x unit_tests{,_parallel} && ./unit_tests_parallel
```

metagraph/scripts/unit_tests_parallel.sh
karasikov: On mex, I got a tiny improvement: 90 sec -> 83 sec.
Let's see what happens with the CI.

metagraph/scripts/unit_tests_parallel.sh
karasikov: Hm. Not sure I got this point. The number of shards is controlled by parameter `-j`. And runnin a specific shard shouldn't be needed. One can filter tests by `--gtest_filter`

adamant-pwn: I was thinking whether we should react in any way if `GTEST_TOTAL_SHARDS` or `GTEST_SHARD_INDEX` are already set externally when this runs (e.g. using it as default value for `WORKERS`, rather than overriding). But maybe flags are sufficient for this.

metagraph/scripts/unit_tests_parallel.sh
karasikov: In the CI it's also ~3min faster per run. Okay, let's keep it at 2 * n_proc

metagraph/scripts/unit_tests_parallel.sh
karasikov: Yeah, no need. Everything can be done with `-j` and `--ftest_filter`

metagraph/scripts/unit_tests_parallel.sh
```diff
@@ -0,0 +1,292 @@
...
+# Aggregate gtest-style summary across all shards.
+TOTAL_TESTS=0
+TOTAL_PASSED=0
+TOTAL_FAILED=0
+FAILED_LIST=()
+CRASHED_SHARDS=()
+for i in "${!PIDS[@]}"; do
+    PLAIN=$(sed -E $'s/\x1b\\[[0-9;]*m//g' "${LOGS[$i]}")
+
+    # "[==========] N tests from M test suites ran. (X ms total)"
```
adamant-pwn, line 230:
```sh
Shard 0: PASSED  603 tests from 233 test suites ran. (1351993 ms total)
Shard 1: PASSED  603 tests from 231 test suites ran. (1040328 ms total)
Shard 2: PASSED  603 tests from 235 test suites ran. (1012202 ms total)
Shard 3: PASSED  602 tests from 235 test suites ran. (882694 ms total)
Shard 4: PASSED  602 tests from 233 test suites ran. (254515 ms total)
Shard 5: PASSED  602 tests from 233 test suites ran. (1130800 ms total)
```
Eh... Best to switch to seconds 😅

metagraph/src/common/logger.cpp
```diff
@@ -41,7 +44,16 @@ class split_sink : public spdlog::sinks::sink {
...
     spdlog::set_automatic_registration(false);
     auto sink = std::make_shared<split_sink>();
-    return std::make_shared<spdlog::logger>("", sink);
+    auto logger = std::make_shared<spdlog::logger>("", sink);
+    // Honor SPDLOG_LEVEL at construction time so any log calls made during
+    // static/dynamic init of other TUs (e.g. tests creating temp dirs) are
+    // filtered at the intended level, not the default of `info`.
+    if (const char* env = std::getenv("SPDLOG_LEVEL"); env && *env) {
+        auto lvl = spdlog::level::from_str(env);
+        if (lvl != spdlog::level::off || std::string(env) == "off")
```
adamant-pwn: Should we add some warning when `SPDLOG_LEVEL` fails to parse?

metagraph/scripts/unit_tests_parallel.sh
adamant-pwn: Nice, macos_latest on x86_64 seems to have reduced by 5 minutes!

It's still slightly imbalanced, though... For the future, it could be worthwhile to split `BOSSConstruct.ConstructionEQAppendingCanonical` into more tests, for example converting it into `TEST_P` over `(k, container, weighted)` triples.

adamant-pwn (approved):
Seems good overall. Maybe some further cleanup of comments could also help, they feel a bit overly verbose.

PR #618, opened 2026-04-25
Support loading RowDisk annotations without mmap

karasikov:
## Summary
The disk-resident row-diff annotators (`RowDiff<RowDisk>`, `IntRowDiff<IntRowDisk>`, `TupleRowDiff<CoordRowDisk>`) previously forced mmap regardless of `--mmap`. Now they respect the flag, with one exception: `boundary_` is always mmap'd (too large to materialize into RAM) and gets `madvise(MADV_RANDOM)` since access is `select1`-driven.

| Component | `--mmap` on | `--mmap` off |
|---|---|---|
| `set_bits_` (`int_vector_buffer<>`) | disk-resident | disk-resident |
| `boundary_` | mmap | **mmap** |
| `anchor_`, `fork_succ_` | mmap | RAM |

## Mechanism
- `utils::named_ifstream` mirrors `sdsl::mmap_ifstream` (carries the filename); `utils::get_filename(istream&)` recovers it via `dynamic_cast` so loaders work with either stream type.
- `RowDisk`/`IntRowDisk`/`CoordRowDisk` `load()` open a fresh `mmap_ifstream` for `boundary_`.
- `annotation_matrix.cpp` drops the `is_same_v<T, RowDiffDisk*Annotator>` mmap override.

## Test plan
- [x] `RowDiffDiskAnnotatorTest.LoadConsistentAcrossMmapSettings` — round-trip with mmap on/off
- [x] `AnnotatedDBGWithNTest/5.*` (all `RowDiff<RowDisk>` typed tests)
- [x] CI green

### DIFF
--- a/metagraph/src/annotation/binary_matrix/row_disk/row_disk.cpp
+++ b/metagraph/src/annotation/binary_matrix/row_disk/row_disk.cpp
@@ -50,14 +50,13 @@ BinaryMatrix::SetBitPositions RowDisk::View::get_row(Row row) const {
 }
 
 bool RowDisk::load(std::istream &f) {
-    auto _f = dynamic_cast<sdsl::mmap_ifstream *>(&f);
-    assert(_f);
     try {
         num_columns_ = load_number(f);
 
         auto boundary_start = load_number(f);
 
-        buffer_params_.filename = _f->get_filename();
+        // int_vector_buffer<> opens the row data by filename + offset.
+        buffer_params_.filename = utils::get_filename(f);
         buffer_params_.offset = f.tellg();
 
         assert(boundary_start >= buffer_params_.offset);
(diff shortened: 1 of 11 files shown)


metagraph/src/annotation/binary_matrix/row_disk/row_disk.cpp
```diff
@@ -50,22 +50,27 @@ BinaryMatrix::SetBitPositions RowDisk::View::get_row(Row row) const {
...
+        // boundary_ is too large to load into RAM, always mmap it.
+        sdsl::mmap_ifstream boundary_in(buffer_params_.filename);
+        boundary_in.seekg(boundary_start, ios_base::beg);
+        boundary_.load(boundary_in);
+        // boundary_ is accessed via select1, hint random access.
+        if (madvise(boundary_in.get_mmap_context()->data(),
+                    boundary_in.get_mmap_context()->file_size_bytes(),
+                    MADV_RANDOM)) {
+            logger->warn("madvise(MADV_RANDOM) on RowDisk boundary failed");
+        }
```
adamant-pwn: Should we factor out this block somewhere, rather than copy-paste it 3 times?

adamant-pwn (approved):
Looks good to me! Maybe factor out boundary_ mmap if there is a more appropriate area though...

